PR #157, opened 2015-01-07
Add test for final N being prime

AlexMooney:
This test would catch the case where the sieve algorithm only goes up to N-1, which is a fairly natural fence post error.  You can see an example of this problem happen [here](http://exercism.io/submissions/d27f2945c27a182eb60f5e4a) and corrected version [here](http://exercism.io/submissions/c0c63408acab08a59fa030ab).  I think adding a third test to this very simple problem is worth helping programmers to carefully consider how they're iterating to stop at the exact number desired.


AlexMooney:
Poignantly enough, the `example.py` has calculates that `sieve(7)` should be `[2,3,5]` and not `[2,3,5,7]`.  Nonetheless, if you ask for primes up to the number 7, it should return that 7 is prime.  I'll fix the `example.py` and resubmit.

PR #159, opened 2015-01-07
Sieve: Updated Example

Dog:
I feel this is a better solution to the problem.


Dog:
This solution has problems with Python 3 support.

While we can convert generators to lists, it becomes much less ideal because we'd have to convert a generator after every filter.

Closing till I think about this further.

PR #169, opened 2015-02-28
Forcing Collision Detection in robot-name Exercise

tstirrat15:
The `test_rest_name` test seems to be written to catch whether the user is using collision detection, but since most people use RNG in their solutions, there's a minute probability of recognizing whether someone has implemented collision detection as currently written.

I rewrote the test to reset to a predefined seed before initialization and before resetting the name, which, if the same logic is used for generating the name in each case, will always result in identical names between the initialization and reset, which should force the need for collision detection.

I guess it'll still fail if there's a big enough difference between the initialization and reset logic (i.e. if they generate numbers first, then letters on the first run, and then do the reverse on the second), but it's a whole lot better than it was.

One last thing: the README for this exercise still treats collision detection as being optional, despite there being a test written to that effect. Should the README be modified as well?

### DIFF
--- a/robot-name/example.py
+++ b/robot-name/example.py
@@ -6,6 +6,7 @@ class Robot(object):
 
     def __init__(self):
         self._name = None
+        self._past_names = set()
 
     def prefix(self):
         return ''.join([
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thanks!

The example solution is failing the new test. Would you mind updating the example to pass? Also--once the tests pass, would you mind squashing the commits?

The README is shared between all the languages, some of which don't test for it. I'll try to clarify it, though.

tstirrat15:
I modified the example code and the commits are squashed. Is there anything else I'd need to do?

kytrinyx:
This looks great, thanks.

PR #182, opened 2015-06-19
Move bob slightly later in the track

kytrinyx:
This puts bob after hamming, leap, and a few others that are fairly simple algorithmically, but before 'meetup', which seems like a bigger problem.


Dog:
Thanks =)

PR #183, opened 2015-06-20
WORKFLOW: Improved Documentation On Tests

Dog:
* Recommended that users install pytest and pytest-cache
* Gave examples of how to use pytest
* Documented how to run tests without pytest

### DIFF
--- a/docs/WORKFLOW.md
+++ b/docs/WORKFLOW.md
@@ -1,22 +1,59 @@
-## Running Tests for an Exercise
+##  Tests
 
-Execute tests for an exercise with:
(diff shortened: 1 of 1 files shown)


kytrinyx:
Much improved; nice!

Dog:
I'm looking into pytest's looponfailing mode. I think it would allow people to have the tests run automatically as they worked on their files - essentially live reload for tests. However, I think the track would then need a .ini file in each exercise.

http://pytest.org/latest/xdist.html#running-tests-in-looponfailing-mode

kytrinyx:
The .ini file would be delivered by default if it were there, I think. If you give me an example of an .ini file (or make a branch where one of the exercises has an .ini file) I could give this a whirl.

PR #188, opened 2015-07-09
Fix atbash-cipher test cases

andrelaszlo:
The current test cases are a bit strange. Some of them expect the cipher to
move spaces around, for example.

The tests now expect encode/decode to:

* Convert to lower case
* Encrypt alphabetical characters
* Preserve all other characters

### DIFF
--- a/atbash-cipher/atbash_cipher_test.py
+++ b/atbash-cipher/atbash_cipher_test.py
@@ -15,30 +15,30 @@ class AtbashCipherTest(unittest.TestCase):
         self.assertEqual("lnt", encode("OMG"))
 
     def test_encode_O_M_G(self):
-        self.assertEqual("lnt", encode("O M G"))
+        self.assertEqual("l n t", encode("O M G"))
 
     def test_encode_long_word(self):
-        self.assertEqual("nrmwy oldrm tob", encode("mindblowingly"))
+        self.assertEqual("nrmwyoldrmtob", encode("mindblowingly"))
 
     def test_encode_numbers(self):
-        self.assertEqual("gvhgr mt123 gvhgr mt",
+        self.assertEqual("gvhgrmt, 1 2 3, gvhgrmt.",
                          encode("Testing, 1 2 3, testing."))
 
     def test_encode_sentence(self):
-        self.assertEqual("gifgs rhurx grlm",
+        self.assertEqual("gifgs rh urxgrlm.",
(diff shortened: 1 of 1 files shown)


andrelaszlo:
Erm, I always wondered why that file was called README...

kytrinyx:
:) Thanks (even though it turned out to not be relevant)

andrelaszlo:
Thanks for a cool and fun site. I'm learning a lot! (For example to read instructions more carefully...) :)

kytrinyx:
I have a friend who used to tell his kid: "The first rule of taking an exam... is to read the %^#$ing question"

andrelaszlo:
I'm trying to get over this reflex that if something is confusing, its probably broken.

It saves a lot of time if you are working on something that is actually mostly broken, but in general it makes you seem arrogant I guess.

Sounds like solid parenting to me. :+1:

kytrinyx:
I think that confusing, in the case of docs, _is_ actually a form of broken, but I don't always know how to fix it. It helps to get it pointed out... gently :)

Dog:
Hey @andrelaszlo, I just wanted to explain Travis a little bit in case you would like to contribute in the future. There is a lot to absorb with Git and various tools used in development, but I think its useful to know and may apply to other GitHub projects you want to contribute to.

One of the things I noticed from your commit was you only updated the test files. When a pull request is submitted for our project, we start a build with Travis.

https://travis-ci.org/

Travis runs all the test cases against our example solutions. Travis knows what to do due to the `.travis.yml` file in the root of our repository.

Here is the Travis result for your pull request. I got the link by clicking the icon next to your pull request.

https://travis-ci.org/exercism/xpython/builds/70264747

When you go there, you will see three build jobs. Each of these build jobs is for a different version of Python. This is because we want to support three versions of Python in the track (2.7, 3.3, and 3.4). If you click on one of those jobs, you can see the output for the build. If you look around line 450, you'll see that the tests failed on the atbash exercise.

The reason the tests failed is because you didn't update the example solution for the exercise. If you go to the repository and click on a problem then you'll see `example.py`. When the build runs, we have Travis test against these example solutions. If you had updated the example solution to meet your test cases, the Travis build would have passed.

If you want to learn more about Travis you can find the documentation here:
http://docs.travis-ci.com/

Hope this helps, and isn't overwhelming. Let me know if you have any other questions. :dog:

PR #190, opened 2015-07-24
Updated meetup tests to prevent creating a `datetime.date` with `mont…

Peque:
A common mistake I have seen is to calculate the last day of a month as:

    >>> date(year, month + 1, 1) - timedelta(days=1)

Which will raise a `ValueError` when `month` is `12` (December). This test
will prevent students from making this mistake again.


kytrinyx:
Thanks!

PR #191, opened 2015-07-30
New (faster and more efficient) implementation of the "sieve" exercise

Peque:
It runs orders of magnitude faster than the previous implementation, specially with large numbers (i.e.: with a value of `100000`, the old implementation took over 14 seconds, while this new one only takes 0.33 seconds).


kytrinyx:
This is great, thanks!

PR #193, opened 2015-08-11
word-count: add test_tabs()

ZacharyRSmith:
I saw multiple solutions that passed the current tests but would fail this one


kytrinyx:
Nice! Thanks so much :)

PR #194, opened 2015-08-11
Allergies: sort expected and actual vals in a particular test, and rename list -> lst

ZacharyRSmith:
I changed ```test_allergic_to_everything()``` because I thought it was broken.

I also re-named a ```list``` property to the more Pythonic ```lst```, since Python has ```list()```

### DIFF
--- a/allergies/allergies_test.py
+++ b/allergies/allergies_test.py
@@ -21,19 +21,19 @@ class AllergiesTests(unittest.TestCase):
         self.assertFalse(allergies.is_allergic_to('strawberries'))
 
     def test_no_allergies_at_all(self):
-        self.assertEqual([], Allergies(0).list)
+        self.assertEqual([], Allergies(0).lst)
 
     def test_allergic_to_just_peanuts(self):
-        self.assertEqual(['peanuts'], Allergies(2).list)
+        self.assertEqual(['peanuts'], Allergies(2).lst)
 
     def test_allergic_to_everything(self):
         self.assertEqual(
-            ('eggs peanuts shellfish strawberries tomatoes '
-             'chocolate pollen cats').split(),
-            Allergies(255).list)
+            sorted(('eggs peanuts shellfish strawberries tomatoes '
+                    'chocolate pollen cats').split()),
+            sorted(Allergies(255).lst))
 
     def test_ignore_non_allergen_score_parts(self):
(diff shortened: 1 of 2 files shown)


kytrinyx:
Good call, thanks!

Dog:
First I'd like to ask how lst is more pythonic. I think lst is a better name because it doesn't collide with the built in function, but I'd like the justification.

Secondly, I left list as list to not break old solutions. With that change that means all solutions before August 11th were broken.

kytrinyx:
Ugh, I'm sorry I didn't leave this for you. I don't have any python experience and am not a good judge of this.

In the Ruby and Go tracks we've been versioning the test suites, since they've inevitably evolved.

I don't actually mind having broken test suites in general, as long as it doesn't start conversations that go "you're doing it wrong".

kytrinyx:
One thing: We could revert this, and then post a comment to all the solutions since Aug 11th (automatically, by either me or rikki- the robot, apologizing for the change and letting them know that we've changed the test suite, if they want to delete the existing test suite and fetch it again to submit a new iteration. That would also avoid the "you're doing it wrong" thing, since the conversation would already have something about broken test suites in them.

ZacharyRSmith:
I've read that "lst" is pythonic because it avoids using the same name as the built-in function "list()". That's all I've got. I'm in no way an expert. So, do as you wish. ( :

Dog:
@kytrinyx 
I don't think its a big deal. I think we should keep it as is. I was more bringing it up because I noticed the issue, so I wanted to point it out. I don't really have a gauge regarding how important it is we don't break previous solutions.

@ZacharyRSmith 
That is the same reason I'd do that kind of variable name change. I just wanted to ask to make sure I understood what the justification was. In this particular instance, its not a big issue. Since its an attribute on our class rather than a function, we aren't overriding the built in function. It'd be up the the implementer if `obj.list` or `obj.lst` is more appropriate. 

If I were implementing this from scratch, I'd get rid of the attribute and instead implement the `__iter__` magic method. This would instead allow people to call `list(obj)` for the list representation, and means you could do something like

```python
for allergy in obj:
    do_something()
```

ZacharyRSmith:
@Dog 
`assertItemsEqual()` and `assertCountEqual()` are awesome functions. Thx for introducing me to them! And thanks for the thorough response.

PR #195, opened 2015-08-11
allergies: Comment-out bitwise-specific test

ZacharyRSmith:
I saw too many non-bitwise solutions where they passed this test with a hack.

### DIFF
--- a/allergies/allergies_test.py
+++ b/allergies/allergies_test.py
@@ -32,8 +32,10 @@ class AllergiesTests(unittest.TestCase):
              'chocolate pollen cats').split(),
             Allergies(255).list)
 
-    def test_ignore_non_allergen_score_parts(self):
-        self.assertEqual(['eggs'], Allergies(257).list)
+    # EXTRA CREDIT:
+    # This test should pass only with a specific type of solution:
+#    def test_ignore_non_allergen_score_parts(self):
+#        self.assertEqual(['eggs'], Allergies(257).list)
 
 
 if __name__ == '__main__':


kytrinyx:
I like the extra credit approach. Would you mind rebasing onto master?

ZacharyRSmith:
#197 First time rebasing. Please let me know if I should have done anything differently, thanks! I'm wondering if I should have tried to alter this pull request?...not even sure if that's feasible.

kytrinyx:
Oh, yepp. Sorry! I could have totally spelled out the steps better. If you want to update a pull request you can do a `git push -f origin yourbranch`. It's totally no biggie, I'll go look at the other one :)

ZacharyRSmith:
No prob. I prob could've/should've figured out how to update a pull request with a little more google fu. Thanks! ( :

Dog:
I'm also not convinced that commenting out a test because people are cheating is the right solution.

kytrinyx:
Yeah, that's a good point. Do you have any other suggestions? Or is it something to bring up during discussions instead of worrying about a technical solution?

Dog:
I think its more of a discussion thing. If its really intended to be extra credit, I think that is fine and making it skipped is the best solution. If we want to remove the requirement, then we should remove the test.

kytrinyx:
Let's remove the test and make this part of the discussion. I think there's more to be learned by talking about it than by making it optional.

Dog:
I think if we remove the test, it won't be a part of the discussion. I think we will have people submit "naive solutions" because they meet requirements. If its kept optional, it allows users who would be frustrated with the bitwise portion to continue.

kytrinyx:
Ah, right, if leaving it in and making it optional is more likely to make it part of the discussion, then I think that is the right way to go.

PR #199, opened 2015-09-01
Peque

Peque:
Updated "simple-cipher" exercise to satisfy all the rules defined in the README file.

### DIFF
--- a/simple-cipher/example.py
+++ b/simple-cipher/example.py
@@ -5,13 +5,13 @@ import random
 
 class Cipher:
 
-    def __init__(self, k=None):
-        if k:
-            self.key = _normalize(k)
-        else:
+    def __init__(self, key=None):
+        if not key:
             random.seed(time())
-            self.key = ''.join(random.choice(ascii_lowercase)
-                               for i in range(100))
+            key = ''.join(random.choice(ascii_lowercase) for i in range(100))
+        elif not key.isalpha() or not key.islower():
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thanks! Would you mind squashing these commits, please?

Peque:
If you would insist, I will, yes. :blush:

However, as stated in the [official Git documentation](https://git-scm.com/book/ch5-2.html):

> [...] try to make each commit a logically separate changeset [...] try to make your changes digestible [...] split your work into at least one commit per issue, with a useful message per commit [...] The project snapshot at the tip of the branch is identical whether you do one commit or five, as long as all the changes are added at some point, so try to make things easier on your fellow developers when they have to review your changes. This approach also makes it easier to pull out or revert one of the changesets if you need to later.

Each of those commits in the pull request do a different thing, which is well explained in the commit message. That way they are easier to understand, review and maybe revert in the future.

So, why would you prefer me to squash them into one? :sweat_smile:

kytrinyx:
"a logically separate changeset" can mean different things at different levels of abstraction. Here I was looking at the entire PR as a global improvement to "simple-cipher".

If you do not wish to squash them, that's fine. In that case, please update the second and forth commit messages so that they mention which exercise they are changing.

Peque:
@kytrinyx Sorry for that. Commit messages updated.

kytrinyx:
Thanks!

PR #201, opened 2015-09-03
rectangles: converted exercice from Rust track

tuxication:


tuxication:
I'm not sure what the problem with the checks.. I can modify the code with some help:)

kytrinyx:
It passes in python 2.x but fails in python 3.x -- unfortunately I don't know enough python to guess at what bits of syntax are the culprits here.

Dog:
This shows as passing for me. I'll try to merge it later today and will manually double check it.

Dog:
Thanks. Sorry it took so long to merge.

PR #202, opened 2015-09-08
Peque

Peque:
Updates to "scrabble-score" exercise:

- Fix the case-insensitive test. The word now really has a mix of upper and lower case letters.
- Create a new, more general, "invalid_word_scores_zero" test. Score zero not only on empty word, but on invalid word as well (including empty).


kytrinyx:
Nice, thanks!

PR #203, opened 2015-09-08
sublist: New code example

Peque:
This new implementation is much faster and cleaner. It does not
need to handle exceptions and it does not need to allocate nor zip
sublists at any point.

Also, a test was added to prevent the use of sets, which leads to incorrect results (although all the other tests pass).


kytrinyx:
Thanks, @Peque!

PR #205, opened 2015-09-12
house: more test

MartinDelille:
I also add a double \n at the end of the rhyme

### DIFF
--- a/house/example.py
+++ b/house/example.py
@@ -12,7 +12,7 @@ parts = [('lay in', 'the house that Jack built'),
          ('', 'the horse and the hound and the horn')]
 
 
-def _verse(n):
+def verse(n):
     v = ['This is {}'.format(parts[n][1])]
     v.extend(['that {0} {1}'.format(parts[i][0], parts[i][1])
               for i in range(n-1, -1, -1)])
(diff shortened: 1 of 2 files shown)


kytrinyx:
This is failing CI, would you mind taking a look?

I've been removing the double "\n" at the end of long texts, on the logic that when I write in text files I don't keep a blank line at the end.

MartinDelille:
ok no problem. Since I had a "verse" function import the test is failing. I'm looking at the .`travis.yml`  and the `check_exercises.py` scripts but I can figure where are the files that are test by the spec?

kytrinyx:
The test runs against the reference solution (example.py)

MartinDelille:
Travis is ok and I removed the double \n. I had to rename `_verse()` to `verse()` in _example.py_ to fix the error. Do you want me to squash the commit?

kytrinyx:
Yeah, if you wouldn't mind squashing the commit, that would be great, thanks!

MartinDelille:
hi @kytrinyx ! What about my PR? Tell me if you think there is a problem about it!

kytrinyx:
Hi @MartinDelille GitHub doesn't notify me when a PR is updated, only when there's a new comment. I never saw this. I'm so sorry!

MartinDelille:
Ok no problem! :+1:

PR #207, opened 2015-09-18
Expect case normalized words

wobh:
Ruby, Elixir, Clojure, and others expect words to be normalized in
some way so that "Go", "go", "GO" are not considered different words.

### DIFF
--- a/word-count/example.py
+++ b/word-count/example.py
@@ -5,4 +5,4 @@ def word_count(text):
     """Return a Counter object that maps from the words contained in
     the phrase to their respective counts
     """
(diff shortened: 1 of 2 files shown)


wobh:
Other language tracks I've seen actually expect downcased words. Here I've tried to make a test that would expect the right count, but be indifferent to the method of normalizing.

kytrinyx:
> Here I've tried to make a test that would expect the right count, but be indifferent to the method of normalizing.

I like that it leaves the choice up to the implementer as to how to do normalization.

PR #213, opened 2015-11-07
Added support for Python 3.5

behrtam:
**Python 3.5.0** was released a month ago and all our tests run successfully with it. I added 3.5 to the travis-ci config and to the readme. Additionally I removed 3.3 which we could keep as the build doesn't take to long, but I don't think we really get something of it.

```
python3.5 test/check-exercises.py
# accumulate
......
Ran 6 tests in 0.000s


[...]


# zebra_puzzle
.
Ran 1 test in 0.000s

OK

SUCCESS!
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 
 python:
   - 2.7
+  - 3.3
   - 3.4
   - 3.5
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: python
+
 python:
-  - "2.7"
-  - "3.3"
-  - "3.4"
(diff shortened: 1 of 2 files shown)


Dog:
I think we should keep the test for 3.3 running for now. It at least lets us know we are still compatible with it. I think it would be better to keep it on until we have a problem with it. As you mentioned, the build doesn't take too long.

behrtam:
Python3.3 only as target in the travis config or also mentioned in the readme?

Dog:
I'd like it if the build supported 3.3+. Thanks for adding 3.5.

Dog:
Thanks =)

PR #221, opened 2015-11-08
robot-simulator: Implemented new exercise

behrtam:
I implemented the **x-common/robot-simulator** exercise ([README.md](https://github.com/exercism/x-common/blob/master/robot-simulator.md). I placed it between `series` and `sieve` in the exercise order to provide some fun between the two math problems ;)


Dog:
I want to look at this a bit closer before I merge it. I don't see anything wrong with it after a quick glance though.

behrtam:
Please take your time. I'm a bit sorry for procrastinating (should really do something different) on this project.
How do you handle the order of the exercises? Random, FIFO or by complexity while avoiding that to similar are close?

kytrinyx:
We kind of order by complexity, but it's pretty subjective and we have no problem reordering later.

We have two discussions that deal with this, where we're not quite decided what the solution is:

- https://github.com/exercism/x-common/issues/102
- https://github.com/exercism/x-common/issues/63

Dog:
Here is my solution under 2.7:

```python

NORTH, EAST, SOUTH, WEST = range(4)


class Map:
    @property
    def coordinates(self):
        return (self.x, self.y)

    def __init__(self, x=0, y=0):
        self.x = x
        self.y = y

    def north(self):
        self.y += 1

    def south(self):
        self.y -= 1

    def east(self):
        self.x += 1

    def west(self):
        self.x -= 1


class Compass:
    compass = [NORTH, EAST, SOUTH, WEST]

    def __init__(self, bearing=NORTH):
        self.bearing = bearing

    def left(self):
        self.bearing = self.compass[self.bearing - 1]

    def right(self):
        self.bearing = self.compass[(self.bearing + 1) % 4]


class Robot:

    @property
    def bearing(self):
        return self.turn.bearing

    @property
    def coordinates(self):
        return self.move.coordinates

    def __init__(self, x=0, y=0, bearing=NORTH):
        self.move = Map(x, y)
        self.turn = Compass(bearing)

    def forward(self):
        move = self.move
        bearing = self.bearing

        if bearing == NORTH:
            move.north()
        elif bearing == SOUTH:
            move.south()
        elif bearing == EAST:
            move.east()
        elif bearing == WEST:
            move.west()

    def left(self):
        self.turn.left()

    def right(self):
        self.turn.right()

    def simulate(self, commands):
        lookup = {
            'A': self.forward,
            'R': self.right,
            'L': self.left}

        for cmd in commands:
            if cmd in lookup:
                lookup[cmd]()
```

And my modified test cases:

```python
import unittest

from robots import Robot, NORTH, SOUTH, EAST, WEST


class RobotTests(unittest.TestCase):

    def test_setup(self):
        robot = Robot(0, 0, NORTH)
        self.assertEqual((0, 0), robot.coordinates)
        self.assertEqual(NORTH, robot.bearing)

    def test_turn_right(self):
        robot = Robot()
        for direction in [EAST, SOUTH, WEST, NORTH, EAST]:
            robot.right()
            self.assertEqual(robot.bearing, direction)

    def test_turn_left(self):
        robot = Robot()
        for direction in [WEST, SOUTH, EAST, NORTH, WEST]:
            robot.left()
            self.assertEqual(robot.bearing, direction)

    def test_forward_positive_north(self):
        robot = Robot(0, 0, NORTH)
        robot.forward()
        self.assertEqual((0, 1), robot.coordinates)
        self.assertEqual(NORTH, robot.bearing)

    def test_forward_positive_east(self):
        robot = Robot(0, 0, EAST)
        robot.forward()
        self.assertEqual((1, 0), robot.coordinates)
        self.assertEqual(EAST, robot.bearing)

    def test_forward_negative_south(self):
        robot = Robot(0, 0, SOUTH)
        robot.forward()
        self.assertEqual((0, -1), robot.coordinates)
        self.assertEqual(SOUTH, robot.bearing)

    def test_forward_positive_west(self):
        robot = Robot(0, 0, WEST)
        robot.forward()
        self.assertEqual((-1, 0), robot.coordinates)
        self.assertEqual(WEST, robot.bearing)

    def test_simulate_1(self):
        robot = Robot(0, 0, NORTH)
        robot.simulate("LAAARALA")
        self.assertEqual((-4, 1), robot.coordinates)
        self.assertEqual(WEST, robot.bearing)

    def test_simulate_2(self):
        robot = Robot(2, -7, EAST)
        robot.simulate("RRAAAAALA")
        self.assertEqual((-3, -8), robot.coordinates)
        self.assertEqual(SOUTH, robot.bearing)

    def test_simulate_3(self):
        robot = Robot(8, 4, SOUTH)
        robot.simulate("LAAARRRALLLL")
      Avoid using lists as default arguments  self.assertEqual((11, 5), robot.coordinates)
        self.assertEqual(NORTH, robot.bearing)

if __name__ == '__main__':
    unittest.main()
```

What I noted:

* Avoid using default mutable arguments: https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/

* You don’t need to explicitly extend object

* I wouldn't capitalize kwargs

* Since you are dealing with only a 2 coordinate system, I'd represent x and y in a tuple rather than a list.

* In the tests, I'd avoid smurf naming (prog1, prog2, prog3 -> 1, 2, 3)

behrtam:
#### Comments on your notes
**no default mutable arguments**
No argument about that.

**explicitly extend object**
If you don't do it you get *old style classes* in Python 2.X, which have been removed in Python 3. In Python2.X you only get *new style classes* if you extend `object`.
I'm in favor of extending `object` so we minimize differences between Python 2-3. One difference for example between new and old style is the result of `type` which may be important information if tests fail.

**avoid smurf naming**
If I do understand the term correctly, renaming it from `test_simulate_prog3` to `test_simulate_3` only makes the common prefix shorter but it still is *smurf naming*. But I do not really have a good naming idea for the 3 test runs.

**capitalize kwargs**
Neither your nor mine implementation uses `kwargs` so can you help me with that comment.

#### Comments on your implementation
I like your `class Compass` because of left/right which makes turning between directions shorter and clearer.
Should we drop the `enum` for Python 3? Don't really see a benefit. My only reason was to try it and maybe that it adds context if it's called `Bearing.NORTH` instead of just `NORTH`.

Using a `list` for the 2D coords really was a bad choice but isn't the `class Map` too much?
```python
class Robot(object):

    @property
    def coordinates(self):
        return (self.x, self.y)

    def __init__(self, x=0, y=0, bearing=NORTH):
        self.x = x
        self.y = y
        self.compass = Compass(bearing)

    def forward(self):
        bearing = self.bearing

        if bearing == NORTH:
            self.y += 1
        elif bearing == SOUTH:
            self.y -= 1
        elif bearing == EAST:
            self.x += 1
        elif bearing == WEST:
            self.x -= 1
```

#### Notes
Maybe we should change the `test_setup` or add a test to check the defaults?
```python
def test_setup(self):
        robot = Robot()  # Robot(0, 0, NORTH)
        self.assertEqual((0, 0), robot.coordinates)
        self.assertEqual(NORTH, robot.bearing)
```

Dog:
Smurf naming - I may not have referred to this right. I was trying to point out prog didn't seem to have any real significance to the tests. Test has meaning for test discovery, simulate is the function being testing, and 1 is its first test. test_forward_positive_west has test for discovery, forward for the function being tested, positive_west for expected result. prog was the only portion that didn't seem to have any real significance under that naming convention.

explicitly extend object - thats fair

kwargs is short for key word arguments, and its why its common to use as a variable name for **kwargs in a method. I was trying to say I would change Bearing to bearing in your arguments.


My implementation

I was experimenting with the idea of making the robot modular. I figured robots tend to be several systems brought together. A robot would probably consist of a movement system, an AI system, a weapons system, etc. If you wrote an interface with common functionality expected from a system, it would be easier to reimplement individual components as needed. A change from a 2 coordinate system to a 3 coordinate system may mean using a different interface, but the AI system may not care if the movement system is implemented via a cartesian system or a polar system if all it needs to do is advance forward. If you were to define interfaces, then the AI system could expect and use a function like forward() without worrying about whats implemented under the hood in the movement system.

If I were to take this a step farther then I would probably take Map and Compass and make them implementations of a movement system interface. Map is simple, but its separation was because I felt the movement was a separate problem from the robot's bearing.

In terms of Enums, I think for this specific problem it may be better to make them define NORTH, SOUTH, EAST, WEST as constants in their code. This way they can define how they want to denote them, and the test cases should still act as expected. I know that in Python 3 enums can give you extra power out of the box, but I think this problem would probably use a simple implementation of it.

behrtam:
> I was trying to say I would change Bearing to bearing in your arguments.

```python
def advance(self):
    if self.bearing == Bearing.NORTH:
        self.y += 1
    elif self.bearing == Bearing.SOUTH:
        self.y -= 1
...
```

Now I know what you are talking about. That are not capitalized kwargs that is just the way you use Enums. <enum class>.<enum member> While Enums are a perfect choice for `Bearings` I understand that in the case of *exercism* a bit more flexibility might be a good idea.

Dog:
I now see the kwargs issue was a typo on my end. While working locally at one point I had this:

Bearing=Bearing.NORTH

So I was suggesting instead:

bearing=Bearing.NORTH

I can see that isn't in the copy in the repo now, so its irrelevant.

Issue #214, opened 2015-11-07
Make all exercises PEP8 compliant except line length


behrtam:
I would like to make all exercises PEP8 compliant except line length (E501) like I already did with the `secret-handshake` 14b726c. Now I am not sure what's the best way to do this. With one pull request for all or a single pull request per exercise that is not compliant.

Exercises with 5 problems on average: [~~'wordy'~~, ~~'word-count'~~, ~~'poker'~~, ~~'simple-cipher'~~, 'nucleotide-count', ~~'nth-prime'~~, ~~'ocr-numbers'~~, ~~'house'~~, 'sublist', 'pythagorean-triplet', ~~'roman-numerals'~~, ~~'meetup'~~, 'strain', ~~'saddle-points'~~, 'secret-handshake', ~~'minesweeper'~~, 'accumulate', 'triangle', ~~'bob'~~, 'pascals-triangle', ~~'phone-number'~~]

Dog:
I would prefer it was submitted per exercise, with the commit message prepending the exercise.

Something like:

wordy: PEP8 Compliance Updated

behrtam:
The end is in sight.🏁

Dog:
Has this issue been completed?

behrtam:
Just created the last 4 pull requests.

PR #230, opened 2015-11-10
phone-number: PEP8 compliance updated

behrtam:
The exercise `phone-number` had one minor inconsistency with PEP8 which I fixed (#214).
```
tmo$ flake8 phone-number/
phone-number/example.py:31:17: E126 continuation line over-indented for hanging indent
```


Dog:
Thanks

PR #233, opened 2015-11-10
triangle: PEP8 compliance updated

behrtam:
The exercise `triangle ` had one inconsistency with PEP8 which I fixed (#214). It could also be called *unreachable code*.
```
tmo$ flake8 triangle/
triangle/example.py:16:16: F821 undefined name 's'
```


Dog:
Thanks

PR #238, opened 2015-11-10
accumulate: PEP8 compliance updated

behrtam:
The exercise `accumulate` had some inconsistency with PEP8 which I fixed (#214).
```
tmo$ flake8 accumulate/ --ignore=E501
./accumulate/accumulate_test.py:8:56: E226 missing whitespace around arithmetic operator
./accumulate/accumulate_test.py:12:37: E226 missing whitespace around arithmetic operator
./accumulate/accumulate_test.py:22:26: E226 missing whitespace around arithmetic operator
./accumulate/accumulate_test.py:22:31: E226 missing whitespace around arithmetic operator
./accumulate/accumulate_test.py:33:59: E226 missing whitespace around arithmetic operator
./accumulate/example.py:4:19: E231 missing whitespace after ','
```

PR #237, opened 2015-11-10
nucleotide-count: PEP8 compliance updated

behrtam:
The exercise `nucleotide-count` had one inconsistency with PEP8 which I fixed (#214).
```
tmo$ flake8 nucleotide-count/
./nucleotide-count/nucleotide_count_test.py:29:1: W293 blank line contains whitespace
```

PR #236, opened 2015-11-10
secret-handshake: PEP8 compliance updated

behrtam:
The exercise `secret-handshake` had some inconsistency with PEP8 which I fixed (#214).
```
tmo$ flake8 secret-handshake/
./secret-handshake/example.py:18:27: E226 missing whitespace around arithmetic operator
./secret-handshake/example.py:33:16: E226 missing whitespace around arithmetic operator
./secret-handshake/example.py:37:15: E226 missing whitespace around arithmetic operator
./secret-handshake/example.py:37:30: E226 missing whitespace around arithmetic operator
./secret-handshake/example.py:47:23: E226 missing whitespace around arithmetic operator
./secret-handshake/example.py:47:31: E226 missing whitespace around arithmetic operator
./secret-handshake/example.py:47:35: E226 missing whitespace around arithmetic operator
./secret-handshake/example.py:47:38: E226 missing whitespace around arithmetic operator
```

PR #235, opened 2015-11-10
pascals-triangle: PEP8 compliance updated

behrtam:
The exercise `pascals-triangle ` had some inconsistency with PEP8 which I fixed (#214).
```
tmo$ flake8 pascals-triangle/
./pascals-triangle/example.py:2:38: E226 missing whitespace around arithmetic operator
./pascals-triangle/example.py:4:1: E302 expected 2 blank lines, found 1
./pascals-triangle/example.py:5:28: E226 missing whitespace around arithmetic operator
./pascals-triangle/example.py:8:1: E302 expected 2 blank lines, found 1
./pascals-triangle/example.py:10:21: E231 missing whitespace after ','
./pascals-triangle/example.py:10:25: E226 missing whitespace around arithmetic operator
./pascals-triangle/example.py:11:27: E226 missing whitespace around arithmetic operator
./pascals-triangle/example.py:11:32: E226 missing whitespace around arithmetic operator
./pascals-triangle/example.py:11:34: E226 missing whitespace around arithmetic operator
./pascals-triangle/example.py:11:37: E226 missing whitespace around arithmetic operator
```

PR #239, opened 2015-11-10
Add flake8 to travis-ci build process

behrtam:
#215, #216, #217, #218, #219, #220, #222, #223, #224, #225, #226, #227, #228, #229, #230, #231, #232, #233, #235, #236, #237, #238

After all the pull requests to fix the **PEP8** compliance I would like to never have to do this again. So I added `flake8` to the travis-ci build process. I think especially in a project that is not run by one person alone a strict linting is needed.

I deliberately excluded `E501 line too long`. There would be the option `--max-line-length` to raise the PEP8 limit of 79 characters but there are some test cases that have very long string data.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,14 @@ python:
   - 3.4
   - 3.5
 
+install:
+  - pip install flake8
+
+before_script:
+  # ignore:
+  # * E501 - line length limit
+  - flake8 . --ignore=E501
+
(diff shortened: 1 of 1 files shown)


Dog:
Thanks

kytrinyx:
Yeah, this is a good idea. Thanks for taking the time to go through all of that!

PR #240, opened 2015-11-11
pig-latin: Implemented exercise from x-common

behrtam:
I implemented the **x-common/pig-latin** exercise ([README.md](https://github.com/exercism/x-common/blob/master/pig-latin.md). I placed it after `prime-factors` to alternate text and number exercises. Now after the commit I see it's followed by a *cipher* and *scrabble*, so we should search for a better spot in the track. Maybe between `nth-prime`and `saddle-points` might be a good spot.

### DIFF
--- a/config.json
+++ b/config.json
@@ -28,6 +28,7 @@
     "luhn",
     "etl",
     "prime-factors",
+    "pig-latin",
     "simple-cipher",
     "scrabble-score",
     "crypto-square",
--- a/pig-latin/example.py
+++ b/pig-latin/example.py
@@ -1,0 +1,23 @@
+import re
+
+re_cons = re.compile('^([^aeiou]?qu|[^aeiou]+)([a-z]*)')
+re_vowel = re.compile('^([aeiou]|y[^aeiou]|xr)[a-z]*')
+
+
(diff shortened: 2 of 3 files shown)


Dog:
Why did you decide to use regular expressions on this problem rather than string slicing?

behrtam:
Both rules (starts with vowel and split consonant) have small exceptions which makes it complicated to write a clean non regex version. Personally I try to avoid regex where ever I can. Maybe I missed something and there is another way ... 

```python
VOWELS = ('a', 'e', 'i', 'o', 'u')

def split_initial_consonant_sound(word):
    if word.startswith('qu'):
        return 'qu', word[2:]
    if word.startswith('squ'):
        return 'squ', word[3:]
    head = ''.join(c for c in takewhile(lambda n: n not in VOWELS, word))
    return head, word[len(head):]

def starts_with_vowel_sound(word):
    if word[0] == 'y' and not word[1] in VOWELS:
        return True
    return word.startswith('xr') or any(word.startswith(x) for x in VOWELS)
```

Dog:
Makes sense. I've gone ahead and merged this problem. Thanks.

PR #242, opened 2015-11-11
circular-buffer: Implemented new exercise

behrtam:
I implemented the **x-common/circular-buffer** exercise. [README.md](https://github.com/exercism/x-common/blob/master/circular-buffer.md)

### DIFF
--- a/circular-buffer/circular_buffer_test.py
+++ b/circular-buffer/circular_buffer_test.py
@@ -1,0 +1,109 @@
+import unittest
+
+from circular_buffer import CircularBuffer
+from circular_buffer import BufferFullException, BufferEmptyException
+
+
+class CircularBufferTest(unittest.TestCase):
+
+    def test_read_empty_buffer(self):
+        buf = CircularBuffer(1)
+        with self.assertRaises(BufferEmptyException):
+            buf.read()
+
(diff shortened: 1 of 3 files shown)


behrtam:
```
tmo$ python3.5 ./test/check-exercises.py circular-buffer
..........
Ran 10 tests in 0.002s

OK

tmo$ cp circular-buffer/example.py circular-buffer/circular_buffer.py
tmo$ python3.5 circular-buffer/circular_buffer_test.py
EEEEEE.EEE
ERROR: test_alternate_read_and_overwrite (__main__.CircularBufferTest)
  File "circular-buffer/circular_buffer_test.py", line 88, in test_alternate_read_and_overwrite
    buf.write(c)
  File "/Users/tmo/github/xpython/circular-buffer/circular_buffer.py", line 22, in write
    self.buffer[self.write_point] = data
TypeError: an integer is required
```

I perfectly understand the `TypeError`. What I don't understand is why there is no error when I run the `check-exercises.py` script with python3.5. At least *travis-ci* caught it.

patricksjackson:
I'd suggest adding a test to check for failure if the input is greater than a single byte, but it looks like other tracks may not have that test either.

Overall I think it looks fine.

PR #246, opened 2015-11-15
Skip tests except the initial in every test suite

TFarla:
Added the skip annotations to make xpython more consistent with other xlanguages.


Dog:
This is a conscious decision. We have the desired workflow documented.

https://github.com/exercism/xpython/issues/189

PR #247, opened 2015-11-15
nucleotide-count: now both count() and nucleotide_counts() reject invalid input.

orozcoadrian:
added validation to nucleotide_counts(). now both count() and nucleotide_counts() reject invalid input. added a new test and corresponding implementation in the example.py

### DIFF
--- a/nucleotide-count/example.py
+++ b/nucleotide-count/example.py
@@ -7,6 +7,8 @@ def count(strand, abbreviation):
 
 
 def nucleotide_counts(strand):
+    for x in strand:
+        _validate(x)
     return {
         abbr: strand.count(abbr)
         for abbr in 'ATGC'
--- a/nucleotide-count/nucleotide_count_test.py
+++ b/nucleotide-count/nucleotide_count_test.py
@@ -37,5 +37,9 @@ class DNATest(unittest.TestCase):
         expected = {'A': 20, 'T': 21, 'G': 17, 'C': 12}
         self.assertEqual(expected, nucleotide_counts(dna))
 
+    def test_validates_nucleotide_counts(self):
+        with self.assertRaises(ValueError):
+            nucleotide_counts('X')
+
 if __name__ == '__main__':
     unittest.main()


kytrinyx:
Thank you, @orozcoadrian. I'll let @Dog and/or @sjakobi review the changes.

Two tiny things

1. Would you mind squashing the commits into a single commit? http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request
2. Take a look at this short blog post about commit messages: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

orozcoadrian:
squashed into a single commit and re-formatted the commit message. thanks.

nucleotide-count/example.py
```diff
@@ -7,6 +7,7 @@ def count(strand, abbreviation):
 
 
 def nucleotide_counts(strand):
+    [_validate(x) for x in strand]
```
behrtam: List comprehensions are meant to be used with functions having return values. Just use a `for` loop.

orozcoadrian:
updated the last commit per the review comment. the ci build failure seems unrelated to my changes. let me know if that's something i should look into...

behrtam:
It's not really related to this pr, but why is there a `'U'` in the `NUCLEOTIDES` constant? Doesn't that make the validation incorrect? Based on the [README](https://github.com/exercism/x-common/blob/master/nucleotide-count.md) only 'A', 'C', 'G', and 'T' are nucleotides.

orozcoadrian:
yea i saw that. according to wikipedia, it's "uracil" and it's only found in RNA, not in DNA. i thought about refactoring a bit to minimize the duplication between 'ATGC' and the NUCLEOTIDES variable but decided to go simple for my first pr. i can remove the 'U' and reuse the variable on line 14 if you want.

kytrinyx:
The build is failing on Python 3.4, but I haven't looked into the details of why.

Dog:
I wouldn't validate the strand inside of this function. While its worth considering validation, I don't think it should be necessary for this problem. In my work, I'd usually be validating the strand before calling functions like this.

orozcoadrian:
ok. i can agree with not incorporating this change. i'll keep looking for any contributions i can help with.

behrtam:
You could start with refactoring the `example.py` by removing the `'U'` and reusing the constant on line 14.

PR #251, opened 2015-11-20
word-count: Change separator from whitespace to non alphanumeric

behrtam:
Change the separator as discussed in #177 to non alphanumeric as it is also done in the haskell track.


Dog:
Thanks

PR #255, opened 2015-11-22
atbash-cipher: Improve assertion message

behrtam:
By using a more specific assertion method, we can improve the detail of the assertion error message. `assertMultiLineEqual` is used by default when Unicode strings are compared with assertEqual(). So this only effects Python 2.7 while Python 3 stays the same.

Now you can see which character is missing or wrong, ...
```python
FAIL: test_encode_long_word (__main__.AtbashCipherTest)
  File "/atbash_cipher_test.py", line 21, in test_encode_long_word
    self.assertMultiLineEqual("nrmwyoldrmtob", encode("mindblowingly"))
AssertionError: 'nrmwyoldrmtob' != 'nrmwy oldrm tob'
- nrmwyoldrmtob
+ nrmwy oldrm tob
?      +     +

```
... instead of a simple it doesn't match.
```python
FAIL: test_encode_long_word (__main__.AtbashCipherTest)
  File "/atbash_cipher_test.py", line 21, in test_encode_long_word
    self.assertEqual("nrmwyoldrmtob", encode("mindblowingly"))
AssertionError: 'nrmwyoldrmtob' != 'nrmwy oldrm tob'

```

### DIFF
--- a/atbash-cipher/atbash_cipher_test.py
+++ b/atbash-cipher/atbash_cipher_test.py
@@ -6,39 +6,41 @@ from atbash_cipher import decode, encode
 class AtbashCipherTest(unittest.TestCase):
 
     def test_encode_no(self):
-        self.assertEqual("ml", encode("no"))
+        self.assertMultiLineEqual("ml", encode("no"))
 
     def test_encode_yes(self):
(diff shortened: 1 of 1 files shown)


Dog:
Thanks

PR #257, opened 2015-11-22
pangram: Implemented new exercise

behrtam:
Example solution for new proposed exercise https://github.com/exercism/x-common/pull/133

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,9 +6,10 @@
   "problems": [
     "hello-world",
     "leap",
-    "word-count",
+    "pangram",
     "rna-transcription",
     "hamming",
+    "word-count",
     "gigasecond",
     "bob",
     "meetup",
--- a/pangram/example.py
+++ b/pangram/example.py
@@ -1,0 +1,5 @@
+from string import ascii_lowercase
+
+
+def is_pangram(sentence):
+    return all(char in sentence.lower() for char in ascii_lowercase)
--- a/pangram/pangram_test.py
+++ b/pangram/pangram_test.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+
+import unittest
+
+from pangram import is_pangram
+
+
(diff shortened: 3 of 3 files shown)


Dog:
Under what case is your set going to have a length over 26? (assuming ascii_letters)

Isn't this exercise region specific due to the use of isalpha? I don't know much about region differences, so I'm going to use German characters in my example only because I'm more familiar with the language.

My understanding of isalpha is it uses the constant string.letters to determine how it should return. If we switched to a "German" region, I'm worried ä, ö, ü, ß will also be included in the set because string.letters is region dependent. Since the size of string.letters changed, this means we could potentially substitute the letter z for the letter ä and still pass.

This raises a few questions: 

* Should a German pangram include  ä, ö, ü, ß?
* Should we consider limiting to ascii_letters?

behrtam:
Good point! While the behaviour of the string method `isalpha()` doesn't change if you switch region or localisation, it still is a bad choice here. `'abcdefghijklmnopqrstuvwΣΩδ'` as input would return `true` while it clearly is missing `'xyz'`. We could use `? in ascii_lowercase` instead of `isalpha()`.
```python
>>> from string import ascii_lowercase
>>> pan = lambda s: len(set(char for char in s.lower() if char in ascii_lowercase)) == 26
>>> pan('abcdefghijklmnopqrstuvwxyz')
True
>>> pan('abcdefghijklmnopqrstuvwöüä')
False
>>>
```

behrtam:
We could even get ride of the `len/set` combination. This is fare more readable.
```python
>>> pan = lambda s: all(char in s for char in ascii_lowercase)
>>> pan('abcdefghijklmnopqrstuvwΣΩδxyz')
True
>>> pan('abcdefghijklmnopqrstuvwΣΩδ')
False
>>>
```

Dog:
I really like the solution with all

PR #258, opened 2015-11-24
run-length-encoding: Implemented new exercise

behrtam:
Example solution for new proposed exercise exercism/x-common#134.


Dog:
I think this problem looks fine to push if the problem is added.

kytrinyx:
The problem has been added now.

PR #259, opened 2015-11-25
Clean up docs for x-api v3

kytrinyx:
I'm getting rid of the help.exercism.io site, and integrating the documentation into the language-specific pages: https://github.com/exercism/exercism.io/pull/2628

There might be some tweaks we'll want to do once it's in production and we see where we need more granularity etc.

This is a direct port from the docs repository, split up into separate files.


behrtam:
I'm not really sure what to do with this and how it is connected to #275 and #278.

kytrinyx:
Oh, this is all old and can be closed. I lost track of it.

PR #263, opened 2015-11-29
rectangles: Change empty string test cases into empty list

monzee:
The test cases with strings as input make no sense in this exercise as there is no happy path at all for just strings. All inputs are expected to be lists of strings. If one practices TDD, these cases would make one start off the wrong foot (especially since these are at the top) writing handlers for strings.

Also included in this PR are minor changes in the example solution. I didn't really mean to include this but I don't know how to split it off.

### DIFF
--- a/rectangles/example.py
+++ b/rectangles/example.py
@@ -12,26 +12,19 @@ class corners():
 
 
 # return corner on the same line
-def same_line(index, list):
-    for c in list:
-        if c.i == index:
-            return c
+def same_line(index, items):
+    return next((c for c in items if c.i == index), None)
 
 
 # return corner on the same column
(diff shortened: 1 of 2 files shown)


behrtam:
```
$ git reset HEAD^ --hard
$ git push -f
```

That should delete the last commit. You should probably save it first, so you can submit it in a second pr.

monzee:
Alright, I'm withdrawing this and will try again tomorrow.

kytrinyx:
@monzee for fututre reference, if you do a `git push -f` it updates the existing pull request which lets you change the existing PR from the ground up rather than closing and reopening a new one.

PR #270, opened 2015-12-01
nucleotide-count: refactor by removing duplication.

orozcoadrian:
refactored by removing duplication of symbol 'NUCLEOTIDES'.
also removed 'U' because we're not considering it a valid nucleotide
for the purpose of this exercise.


behrtam:
👍

PR #273, opened 2015-12-01
rail-fence-cipher: Implement new exercise

behrtam:
Example solution for new proposed exercise exercism/x-common#144.


kytrinyx:
Merged the metadata. I like this idea!

PR #276, opened 2015-12-07
Updated INSTALLATION.md with Anaconda/Miniconda

mandel01:
Anaconda offers a useful environment especially when integrating additional scientific and data analytic packages (e.g., numpy, scipy, pandas). As an example, this is the installation that is preferred by the Software Carpentry workshops: http://software-carpentry.org/workshops/setup.html.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,5 +1,7 @@
 If Python isn't already available on your system follow the instructions at [the Hitchhiker's Guide to Python](http://docs.python-guide.org/en/latest/#getting-started) to install Python on your computer.
 
+If you intend to use Python for scientific computing or data analytics, consider instead installing the [Anaconda or Miniconda versions](https://www.continuum.io/downloads), in which additional relevant packages have been bundled in the installation.
+
 Exercism currently supports Python3.4, Python3.3 and Python2.7.
 
 


behrtam:
Good idea. I don't think that the exercises really need `numpy` or `pandas` but anaconda also makes the start with Python on windows a lot easier.

Maybe we should update the Python versions (https://github.com/exercism/xpython/pull/213) in the `INSTALLATION.md` file in a separat pull request.

mandel01:
Great idea, especially with Anaconda having Python 3.5 as their default Python 3 installation.

Dog:
I'm not sure how I feel about mentioning Anaconda. Its not really related to exercism.

mandel01:
I see your point here, especially as there are different directions (vendors, etc.) that people could use.

However, the current instructions point to a page for experts, not for programming beginners. It advocates installing Pip, Setuptools, VirtualEnv, homebrew (on Mac), etc. It lays out commands, but does not provide any support if the commands don't work, which is realistic if a user has already tinkered with installations previously but not done things properly (or used Enthought or another method). Anaconda gets around all of this. I don't know that it is the right solution here. Software Carpentry thinks so and I've seen it work well for their users across platforms, but perhaps there is a different solution for Exercism? I would like to see the current text change and would be happy to craft some better guidance if there is consensus on what direction to go here.

One option might be to rewrite what I wrote as mentioning [additional implementations of Python](https://www.python.org/download/alternatives/), though I still think we should suggest Anaconda (or some similarly cross-platform, user-friendly version for beginners).

mandel01:
I am going to close this PR. I still think that having an easy onboarding process to Python is a good idea, but perhaps Anaconda is not the way to do that. I will think more about the general issue of providing an easy instruction set to get started and submit a new PR when I think I have a better idea there.

kytrinyx:
I agree--it would be great to make sure the Python track onboarding experience is good for beginners, especially since Python is one of the languages that beginners tend to come here for.

PR #277, opened 2015-12-07
Normalize YAML in Travis CI config

kytrinyx:
The YAML was not quite valid. Travis can read it
and wasn't complaining, but this ensures that the YAML
is clean.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,18 @@
+---
 sudo: false
-
 language: python
-
 python:
-  - 2.7
-  - 3.3
-  - 3.4
-  - 3.5
-
+- 2.7
+- 3.3
+- 3.4
+- 3.5
 install:
-  - pip install flake8
-
+- pip install flake8
 before_script:
-  # ignore:
-  # * E501 - line length limit
-  - flake8 . --ignore=E501
-
+# ignore:
+# * E501 - line length limit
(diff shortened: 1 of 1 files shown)


behrtam:
I'm not a yaml expert, but if I understand http://yaml.org/spec/1.2/spec.html the current yaml file is as valid as your new one.
Strings do not require enclosure in quotations. If we would wan't to use quotations, than we should be strict and also use them in lines 2, 3, 5, 6, 7, 8, 10 and 14. Optional blank lines for readability and indentation for key-property structures are also both part of the yaml spec.

kytrinyx:
Ah, I thought that the missing `---` at the top was invalid. Closing this.

behrtam:
If you have multiple documents per YAML file, you need to separate them by three dashes. If you only have one document, you can remove them.
The three dashes are also used to separate directives (e.g. `%YAML 1.2`) from document content.

Example from the spec:
```yaml
# Ranking of 1998 home runs
---
- Mark McGwire
- Sammy Sosa
- Ken Griffey

# Team ranking
---
- Chicago Cubs
- St Louis Cardinals
```

kytrinyx:
Cool, I didn't realize this was a multiple-document thing. I was running all the .travis.yml files through an external tool, and it added the `---` so I figured it was necessary.

Issue #280, opened 2015-12-10
Need LEARNING.md


outkaj:
"Learning Python" section [here] (http://exercism.io/languages/python) currently reads "We don't have any good resources for learning this from scratch". This should be fixed by adding a LEARNING.md file to docs/ with relevant resources. 

This and #279 are good candidates for [24pullrequests] (http://24pullrequests.com/).

Issue #279, opened 2015-12-10
Need TESTS.md


outkaj:
"Running the Tests" section [here] (http://exercism.io/languages/python) currently reads "We're missing the documentation about how to run the tests". This should be fixed by adding a TESTS.md file to docs/ with appropriate directions.

PR #285, opened 2016-01-24
Created LEARNING.md, TESTS.me and added some resources

oalbe:
Hey there,

Issue #280 and #279 stated the need for a `LEARNING.md` and a `TESTS.md` doc files. I created them and did my best with English, but I can't really guarantee zero grammar errors since I'm not a native.

I hope this will turn out to be useful.

### DIFF
--- a/largest-series-product/largest_series_product_test.py
+++ b/largest-series-product/largest_series_product_test.py
@@ -34,12 +34,25 @@ class SeriesTest(unittest.TestCase):
         series = "52677741234314237566414902593461595376319419139427"
         self.assertEqual(28350, largest_product(series, 6))
 
-    def test_identity(self):
+    def test_string_with_all_zeroes(self):
+        self.assertEqual(0, largest_product("0000", 2))
+
+    def test_string_where_all_spans_contain_zero(self):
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks great. Would you mind checking what happened with the travis build? It seems unlikely that a documentation fix caused it to fail, but I'm not familiar with python. I kicked off the build again in case it was transient.

oalbe:
In fact from the logs it doesn't look like the documentation files are the problem. I'm not extremely familiar with Python either, but to me, it looks like the command
```
 flake8 . --ignore=E501
```
Returned a bunch of errors from five files. If you have the access to the Travis configuration for building, would you mind trying this command instead:
```
 flake8 . --ignore=E501,E731
```
This will also ignore the error that is being thrown 6 times from that files, though it would leave the warning alone. If it works, it should build without problems this time, and we can wait for someone who knows more about those things (like the track maintainer?) to definitely figure out and fix what happened.

I'm sorry but I wouldn't know how to help apart from this :/

kytrinyx:
The build should be fixed if you want to rebase it onto master.

PR #293, opened 2016-01-26
Created LEARNING.md and TESTS.md.

oalbe:
[Here](https://github.com/exercism/xpython/pull/285) for reference, because I messed up and had to reset the fork.

@kytrinyx I'm sorry, this is the only solution that worked... I'll pay more attention next time.

### DIFF
--- a/docs/LEARNING.md
+++ b/docs/LEARNING.md
@@ -1,0 +1,18 @@
+# Learning Python From Scratch
+Python is, as Wikipedia goes, a powerful *general-purpose high-level programming language*. It basically means that it can be used to write a wide variety of different kinds of software, from videogames to HTTP servers to command-line tools.
+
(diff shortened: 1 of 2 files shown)


behrtam:
> pip is already installed if you're using Python 2 >=2.7.9 or Python 3 >=3.4 ... https://pip.pypa.io/en/stable/installing/

So it's not totally true to say it only comes with Python3. I wouldn't mention it, because almost all should have it bundled with their python.

> ... install `pytest` though `pip` ...

small typo -> through or with

> This because in the latter case pytest will just return a positive result, only specifying that it ran zero tests.

*I would rewrite that one a bit. Maybe ...* "If you test your implementation (`bob.py`) instead of the test suite (`bob_test.py`), it will return a positive result because it contains no test cases. It looks like this:" *... but writing english is not really a skill of mine.*

oalbe:
@behrtam I'll fix those things asap, thank for the corrections. Though it would still be nice to have a native speaker proofread it anyway.

oalbe:
@behrtam Let me know how it looks to you now.

behrtam:
Could you please squash the two commits to a single one and while you do this you can also update your commit msg (imperative mood and no period [#more](http://chris.beams.io/posts/git-commit/#end)).

oalbe:
@behrtam Better? Thank for the link, I had no idea about those punctuation and mood rules.

Issue #294, opened 2016-01-27
check tests against the config.json


behrtam:
As discovered by @oalbe while working on the test runner (`check-exercises.py` #286) it would be good if the actual executed tests are checked against the listed exercises in the `config.json`. Maybe we could also stop running tests for deprecated exercises.

kytrinyx:
Yeah, that would be great. Configlet checks that the configured problems all exist, but this sounds like it would solve a different problem.

oalbe:
Ehm... I think I did this. I mean, I modified the `check-exercises.py` script so that it now only runs tests if they are in the `'problems'` key of the `config.json` file and prints something in other cases.

What I was wondering is: is it okay to submit a PR even when one could use some peer-review first? (like I do in this case)

P.S.: I've seen the in the `config.json` file there is an `'ignore'` key to tell it to ignore non-exercise folders. I think it's not needed anymore since now the exercises are in their own folder.

kytrinyx:
>  is it okay to submit a PR even when one could use some peer-review first?

Yepp, that's the perfect use case for a PR. Sometimes we open them when we just want to have a discussion about something and it's easier to show what we mean than to describe it.

> P.S.: I've seen that in the config.json file there is an 'ignore' key to tell it to ignore non-exercise folders. I think it's not needed anymore since now the exercises are in their own folder.

Yeah, once all of the tracks have moved everything to `exercises/` then I'll remove that and update configlet to stop checking.

oalbe:
Alright, so I will do the PR. let's see how it goes.

PR #296, opened 2016-01-27
Fixed the check-exercises.py script to run tests against the config.json file

oalbe:
As for issue #294.

This time I'm directly pushing changes with the "single dot" paths. I tested it with two dots and everything seemed to be fine.

Parts that don't really convince me:
- Lines 73-76. I did the checks in a ways that the script returns an hint in case you try checking deprecated exercises or exercises that don't even exists (which means typos too).

Before I put my hands on this script, there weren't really many suggestions returned by the scripts, so I was thinking maybe this is a code of conduct you decided to adopt? I don't know, in any case let me know thanks.

There's also a chance that I completely misunderstood the assignment for this issue in which case I apologise.

### DIFF
--- a/test/check-exercises.py
+++ b/test/check-exercises.py
@@ -7,6 +7,7 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import json
 
 
 def check_assignment(name, test_file, example_name):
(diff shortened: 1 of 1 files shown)


behrtam:
Okay, so [configlet](https://github.com/exercism/configlet) checks, that every exercise has an example solution and an entry in the `config.json`. That makes it save to work with the problems from the json-file in line 81.

Line 70 & 82: String formatting should be preferred over raw concatenation. `'./exercises/{}/*_test.py'.format(exercise_name)` Maybe its also a good idea to standardize the two names `exercise_path` and `exercise`.

Your solution doesn't handle the reason we are doing all this, which is the case that the test file does not exist or is in a wrong location. To achieve this we could work with `os.path.exists` and remove the usage of the glob module entirely.

For example if we rename a test file it looks like this:
```
$ mv exercises/luhn/luhn_test.py exercises/luhn/luhn_test_na.py
$ ./test/check-exercises.py luhn
  File "./test/check-exercises.py", line 97, in <module>
    main()
  File "./test/check-exercises.py", line 71, in main
    )[0]
IndexError: list index out of range
$ ./test/check-exercises.py
# grains
.......
Ran 7 tests in 0.000s

OK

  File "./test/check-exercises.py", line 97, in <module>
    main()
  File "./test/check-exercises.py", line 82, in main
    test_file = glob.glob('./exercises/' + exercise + '/*_test.py')[0]
IndexError: list index out of range
```

oalbe:
@behrtam I'm sorry, I don't get it. In the issue you stated that the `check-exercise.py` should check that the exercise is in the `config.json` and is not deprecated before doing the test. But from what you are saying here it looks like the _configlet_ script already ensures that every exercise is listed in the `config.json` file and has an example implementation to test against.

So, what is exactly the problem? Assuring that all the test files exist (_and_ have a legal name) before running the test? Should the `check-exercise.py` go scavenge all the possible paths in case it can't find the test file? Something else?

I'm sorry if I got it wrong, I thought it was just a matter of ensuring that the exercise was listed in the `config.json` file and not deprecated before running the tests. I still think I can manage to fix this, but first I need to understand what's the required behaviour you expect from it.

PR #300, opened 2016-02-03
limit line length to 79 chars

rootulp:


kytrinyx:
I'm not familiar with python style. In Go the recommendation is to leave everything on one line even if it exceeds 79 columns. It's not that they like long lines, but they prefer longer lines to broken lines. Is there a way of using shorter variable names without making it more unreadable in order to keep it both on a single line and below 79 columns?

PR #303, opened 2016-02-03
acronym: implement new exercise

### LINKED ISSUE #260: acronym: implement new exercise
The exercise `acronym` ([md](https://github.com/exercism/x-common/blob/master/acronym.md), [yml](https://github.com/exercism/x-common/blob/master/acronym.yml)) is a very easy one `''.join(word[0].upper() for word in words.split())`. Should we implement it or should we add it to the "foregone" section in the `config.json`.

rootulp:
Implementation of the acronym exercise closes #260

### DIFF
--- a/config.json
+++ b/config.json
@@ -22,6 +22,7 @@
     "sieve",
     "atbash-cipher",
     "sum-of-multiples",
+    "acronym",
     "say",
     "largest-series-product",
     "kindergarten-garden",
--- a/exercises/acronym/acronym_test.py
+++ b/exercises/acronym/acronym_test.py
@@ -1,0 +1,28 @@
+import unittest
+
+from acronym import abbreviate
+
+
+class AcronymTest(unittest.TestCase):
+
+    def test_basic(self):
+        self.assertEqual('PNG', abbreviate('Portable Network Graphics'))
+
+    def test_lowercase_words(self):
+        self.assertEqual('ROR', abbreviate('Ruby on Rails'))
+
+    def test_camelcase(self):
+        self.assertEqual('HTML', abbreviate('HyperText Markup Language'))
+
+    def test_punctuation(self):
+        self.assertEqual('FIFO', abbreviate('First In, First Out'))
+
+    def test_all_caps_words(self):
+        self.assertEqual('PHP', abbreviate('PHP: Hypertext Preprocessor'))
+
+    def test_hyphenated(self):
+        self.assertEqual('CMOS', abbreviate('Complementary metal-oxide semiconductor'))
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/exercises/acronym/example.py
+++ b/exercises/acronym/example.py
@@ -1,0 +1,6 @@
+import re
+
+
+def abbreviate(words):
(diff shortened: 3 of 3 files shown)


rootulp:
I believe the Travis CI build is failing because I haven't added `"acronym"` to config.json
I'm wondering what is the ordering of problems in this file / where should `"acronym"` be added?

behrtam:
Finding the right spot in the track is quite hard. As this is a really easy exercise it should definitely be somewhere at the start maybe even between `leap` and `pangram`.

What are your reasons for choosing a **regular expression** instead of maybe **`.split()`**?

```python
''.join(word[0].upper() for word in (re.findall('[A-Z]+[a-z]*|[a-z]+', words)))

''.join(word[0].upper() for word in words.split())
```

rootulp:
```python
''.join(word[0].upper() for word in words.split())
``` 
won't work for the following tests:
```python
    def test_camelcase(self):
        self.assertEqual('HTML', abbreviate('HyperText Markup Language'))

    def test_hyphenated(self):
        self.assertEqual('CMOS',
                         abbreviate('Complementary metal-oxide semiconductor'))
```
so I used a regular expression similar to the [ruby](https://github.com/exercism/xruby/blob/master/exercises/acronym/example.rb) and [javascript](https://github.com/exercism/xjavascript/blob/master/exercises/acronym/example.js) example solutions

behrtam:
I didn't look at the test cases and assumed them to be even simpler, sorry.

While it's somehow possible to solve this without a `regex` your solution is definitvly more readable and easier to understand.

```python
def abbreviate(words):
    words = ''.join(c if c.isalnum() else ' ' for c in words).split()
    
    abb = ''

    for word in words:
        if not (word.isupper() or word.istitle() or word.islower()):
            abb += ''.join(c for c in word if c.isupper())
        else:
            abb += word[0].upper()

    return abb
```

rootulp:
I don't think acronym should be the 3rd task in this track. Introducing `regex` too early might be intimidating for beginners. I recommend moving acronym later in the track but before other exercises that use more complicated regular expressions ([pig-latin](https://github.com/exercism/xpython/blob/master/exercises/pig-latin/example.py) or [phone-number](https://github.com/exercism/xpython/blob/master/exercises/phone-number/example.py)).

behrtam:
Good point, maybe after `sum-of-multiples` is a fitting spot.

Could you please do three last small code adjustments, even if you might no like them as they are longer than 79 characters line limit.

* remove the line break in line 24/25 (acronym_test.py)
* remove the unnecessary brackets in line 6 `(re.findall..)` (example.py)
* remove the line break in line 5/6 (example.py)

If you really don't like the long line in the example implementation, maybe this is an option:
```python
def abbreviate(words):
    regex = '[A-Z]+[a-z]*|[a-z]+'
    return ''.join(word[0].upper() for word in re.findall(regex, words))
``

behrtam:
To cleanup all the work, could you [squash](http://push.cwcon.org/learn/squashing-commits.html) the 5 commits into a single one. Than finally this new exercise can go live.

behrtam:
Thank you for all the work and time.

rootulp:
No problem! Thanks for all the helpful feedback!

Issue #295, opened 2016-01-27
pinning versions in travis-ci config


behrtam:
As discussed in #287 it might be a good idea to pinn the version of `flake8` and its dependencies in the travis-ci config (`.travis.yml`).

**Edit:** Maybe we can use https://requires.io to monitor our dependencies. Never used it.

Issue #287, opened 2016-01-25
Travis CI build is failing across the board... somewhat mysteriously


kytrinyx:
Since it started happening consistently and quite suddenly, I wonder if something changed on Travis CI's end with how they set up the environment.

If someone has a moment to dig into what's going on, that would be great.

kytrinyx:
Ah. See discussion in https://github.com/exercism/xpython/pull/285 -- @oalbe has a suggestion. I'll give it a try.

petertseng:
I dug into it and have [the explanation](https://github.com/exercism/xpython/pull/283#issuecomment-174270903) for what's going on, but I'm not stylistically opinionated enough to decide how to fix it (either do what the violations direct, or ignore them), so I leave that up to y'all.

kytrinyx:
Ah, OK, that explanation makes senes. Yeah, I don't have any strong opinions on this (not being a python dev, and all).

behrtam:
it's a good idea to only make use of `lambda` to use a function without assigning it, like:
```python
sorted(data, key=lambda item: item.score)
```

behrtam:
I have no experience with continuous integration. We could change the `travis-ci` config, to work with a fixed flake8 version `pip install flake8==2.5.1` so we don't break builds without changing anything, only because there is a new `flake8` version. The down side is, that we would need to upgrade the `travis-ci`config every time we want to use a new version of `flake8` (shouldn't happen too often).

kytrinyx:
If it's not something that happens too often, I think it's worth it to just change the travis-ci configuration (`.travis.yml`).

petertseng:
1. Short term: Thanks for dealing with this; other PRs are good to go once they rebase (I know because I rebased mine and it's good)

2. Long term: Pinning versions would indeed help prevent things like this; y'all would have to periodically check for new versions as y'all deem appropriate. Note that you would want to pin at least pep8 in addition to flake8. Probably also pyflakes too. Reasoning being that flake8 hasn't changed since December 30 2015 and yet things still broke in here; it was pep8 that changed.

behrtam:
`flake8` changed their pinning from `"pep8 >= 1.5.7, < 1.6"` in 2.4 to `"pep8 >= 1.5.7, != 1.6.0, != 1.6.1, != 1.6.2"` in 2.5. They basically decided against pinning ([Issue](https://gitlab.com/pycqa/flake8/issues/32)). So I think we should pin down all the versions.

PR #305, opened 2016-02-08
Pin versions of used pip packages

behrtam:
To avoid breaking our builds because of changes in pip packages we use
for testing it makes sense to introduce version pinning to the travis
build. (see #287, #295)


kytrinyx:
This makes good sense.

PR #307, opened 2016-02-12
Correct iPython to IPython

mandel01:


behrtam:
Maybe there is also an **i**Python by Apple. Thanks for the PR.

PR #308, opened 2016-02-12
Change version pinning to only use minor versions

behrtam:
Instead of pinning to an exact release (X.Y.Z) this way we pin to a minor version which will include the bug-fix releases automatically.

As [requires.io](https://requires.io/github/exercism/xpython/requirements/?branch=master) detected the first update, I thought we improve the version pinning a bit. Pinning to minor versions should be enough.


kytrinyx:
This looks good to me. I suppose that technically patch releases could go above 99, but it seems like that is unlikely to happen (and if it does we can fix it manually).

Issue #284, opened 2016-01-24
Luhn exercise: are you sure about the checksums in the tests?


oalbe:
Hey there,

I was solving the exercise `luhn.py` and I'm noticing quite a weird thing.
When I did this exercise for the JavaScript track, some of the checksums were different from what they are in the Python track. Isn't the Luhn method supposed to generate always the same results regardless of the language?

Here are some example:
*luhn_test.py*
```python
def test_checksum1(self):
        self.assertEqual(2, Luhn(4913).checksum())

def test_ckecksum2(self):
        self.assertEqual(1, Luhn(201773).checksum())
```

*luhn.spec.js*
```javascript
it('checksum',function() {
    var luhn = new Luhn(4913);
    expect(luhn.checksum).toEqual(22);
  });

it('checksum again',function() {
    var luhn = new Luhn(201773);
    expect(luhn.checksum).toEqual(21);
  });
```

As you can see, the numbers to check are the same, but the results differ from the two languages.
Is this an error? (in which case I can correct it now and submit a pull request) Or is it the way it is supposed to be? (unlikely)

Thank you.

kytrinyx:
I'm not sure about the checksums in the tests--they should be the same. It would be great to figure out what the correct inputs/outputs are and then extract all the data into a luhn.json file per https://github.com/exercism/todo/issues/110

Then we could normalize all of the tracks that have this exercise.

If you have time to figure out what the right numbers are, would you mind adding them to exercism/todo#110? If you have time to do a pull request, that would also be awesome!

oalbe:
Yep, I can do that, but it will take me a while since I don't have much free time at the moment. I will submit a comment there as soon as I will have the thing sorted out. And I apologize in advance if this will end up protracting for a bit.

kytrinyx:
No worries at all -- that's the nature of this beast. Now that it's documented, someone else might have time to take a look as well.

behrtam:
Not all languages that implement this exercise use those two numbers in their tests, but for example `ruby` and `java` match with `javascript`. Looks like the Python checksums are wrong.
http://www.ee.unb.ca/cgi-bin/tervo/luhn.pl?N=4913
http://www.ee.unb.ca/cgi-bin/tervo/luhn.pl?N=201773

behrtam:
@oalbe Do you want to make the small adjustment to the luhn implementation?

oalbe:
[...]

PR #310, opened 2016-02-16
Fix luhn's checksum wrong test values and its implementation.

oalbe:
I finally managed to do it. Issue #284 pointed out the wrong checksum values in the test cases for the `luhn` exercise. I corrected them and corrected the implementation.

There's something I have to point out regarding the implementation of the exercise (`example.py`) though. I didn't really correct it, I just replaced the old one with my implementation _slightly_ modified. It works and all tests pass, but it would be nice to first give it a look and highlight potential problems.

As for the _slightly_ differences with my implementation all I changed is:
- added the `(object)` part to the first line `class Luhn(object)`. I don't really have a precise idea what it does, but it was there in the old implementation so I added it.
- removed the comments.
- cleaned up the unnecessary white spaces from empty lines (since last time I remember Flake8 complaining).

I hope I did everything the right way.

### DIFF
--- a/exercises/luhn/example.py
+++ b/exercises/luhn/example.py
@@ -10,10 +10,10 @@ class Luhn(object):
                 for i, n in enumerate(old_digits, start=len(old_digits) % 2)]
 
     def checksum(self):
-        return sum(self.addends()) % 10
+        return sum(self.addends())
 
     def is_valid(self):
-        return self.checksum() == 0
+        return self.checksum() % 10 == 0
 
     @staticmethod
     def create(n):
--- a/exercises/luhn/luhn_test.py
+++ b/exercises/luhn/luhn_test.py
@@ -16,10 +16,10 @@ class LuhnTests(unittest.TestCase):
                          Counter(Luhn(8631).addends()))
 
     def test_checksum1(self):
-        self.assertEqual(2, Luhn(4913).checksum())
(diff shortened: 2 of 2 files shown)


behrtam:
Why not just move the `% 10` to the appropriate position? That small change would sufficiently satisfy the changed test cases.

```python
def checksum(self):
    return sum(self.addends())

def is_valid(self):
    return self.checksum() % 10 == 0
```

oalbe:
Because... I didn't think about it. I took back the old implementation and made the fix. I also added the empty line at the end of the file and rebased into a single commit. Should be fine now.

behrtam:
You accidentally changed the import in the test file with the new commit.

```
 -from luhn import Luhn
 +from example import Luhn
```

Please shorten your commit message which also shouldn't end with a point ([more about commit msg](http://chris.beams.io/posts/git-commit/#seven-rules)). Than we are good to go.

oalbe:
Done, I hope in a good way. And rebased everything in a single commit once again. (it's quite a challenge to fit the description for something in just 50 characters)

behrtam:
Thanks for finding and fixing it.

oalbe:
Thanks to you for the help.

kytrinyx:
Sweet, I appreciate the fix!

PR #314, opened 2016-02-27
Fix typo in documentation where pip is pip3

StevenACoffman:


behrtam:
Thanks!

StevenACoffman:
So... I'm actually a total newbie to Python. It looks like if you have pypy3 installed (e.g. `brew install pypy3`), then you have to use `pip_pypy3 install pytest`, if you have Python3, I assume you use `pip3 install pytest`. I only had python 2.7 installed, so I got an error when I followed the directions. Is it fair to assume that anyone with a different version has `pip` aliased to the correct thing for them?

PR #316, opened 2016-03-13
largest-series-product: Do not test slices

petertseng:
The slices functions is an internal implementation details and thus the
test case for the largest-series-product problem should not be concerned
with testing it.

Its presence may cause students to falsely think that their solution has
to use this function, instead of the alternative implementation of only
iterating through the digits once.

The series tests are already well-covered by the series exercise already
existing in this track.

If it is desired to give hints on how to approach this problem (which
was one advantage of having the slices test), then consider including a
hints file and/or directory in the largest-series-product directory.

This PR arises from discussion in
https://github.com/exercism/x-common/issues/192

### DIFF
--- a/exercises/largest-series-product/largest_series_product_test.py
+++ b/exercises/largest-series-product/largest_series_product_test.py
@@ -1,26 +1,17 @@
 """Tests for the largest-series-product exercise
 
 Implementation note:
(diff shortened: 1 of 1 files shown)


behrtam:
That totally makes sense. Thanks!

PR #321, opened 2016-03-26
Add a unit test for python minesweeper

RryLee:
I think should add this test.

    +-----+
        * |
    +-----+

this is invalid


kytrinyx:
Thanks, @RryLee!

@betegelse You implemented this exercise originally; would you have time to review this change?

betegelse:
Sure, give me a couple of days

behrtam:
Do you need any help?

RryLee:
@behrtam I dont't see the offical solution, just completed it and found this flaw

behrtam:
So the actual example implementation already works for this proposed new test case. The existing test case `test_faulty_border` does basically the same check, just with a different incorrect character. That's why I don't see any reason to add this.

Issue #322, opened 2016-04-02
Grade school test is imposing unnecessary restrictions on implementations


de2Zotjes:
The tests for the grade school exercise assert against an attribute of the class to implement. This imposes restrictions on the name and type of the attribute. Elsewhere in the testsuite the data is accessed via the 'grade' method, that is possible for all tests.

kytrinyx:
That's a good observation. We've seen this problem in a number of different language tracks, and have slowly been improving the test suite to allow a wider variety of approaches in the solutions.

de2Zotjes:
Since you agree this is an issue I have taken the liberty to create what I think is an improved set of tests for the grade-school exercise

PR #324, opened 2016-04-02
Update grade_school_test.py

de2Zotjes:
Changed the tests to minimize the restrictions on implementation.

Ref. issue #322


de2Zotjes:
copy pasting never works ....

kytrinyx:
The build is failing; would you take a look, please?

PR #325, opened 2016-04-02
Update grade_school_test.py

de2Zotjes:
Changed the tests to minimize the restrictions on implementation.

Ref Issue #322

### DIFF
--- a/exercises/grade-school/grade_school_test.py
+++ b/exercises/grade-school/grade_school_test.py
@@ -7,36 +7,38 @@ from school import School
 
 class SchoolTest(unittest.TestCase):
     def setUp(self):
+        if not hasattr(self, 'assertCountEqual'):
+            self.assertCountEqual = self.assertItemsEqual
         self.school = School("Haleakala Hippy School")
 
     def test_an_empty_school(self):
         for n in range(1, 9):
-            self.assertItemsEqual(set(), self.school.grade(n))
+            self.assertCountEqual(set(), self.school.grade(n))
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
This is failing on Python 3.x. Would you mind taking a look?

behrtam:
The method `assertItemsEqual ` is named `assertCountEqual` in Python3. [[see docs]](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertItemsEqual)

As the order in witch the names are added does matter in this exercise I would suggest to stay with `assertEqual`.

Could you change your commit message to match the style described in the [readme](https://github.com/exercism/xpython#pull-requests).

de2Zotjes:
So I did something ugly to make this test work for 2.7 and 3.5 :)

As to the ordering being important I do not read that in the exercise readme. There is a sorted output where ordering is (ofcourse) important, but I took that as an indication that at other moments ordering is free/optional.
using the assertEqual operation gives rise to added restrictions on the return type of the data. A list won't equal a tuple, whereas the assertCountEqual considers both list and tuples as sets of data with counts of data.

behrtam:
Had just done the Scala version of this exercise where they test on the initial add-order but I agree the description doesn't mention it.
It's a bit ugly but I don't see a better why of supporting version 2 and 3.

If you could adjust the commit message to the form `exercise-name: Add some things` I can merge this PR.

de2Zotjes:
Last commit message is clean :)

PR #331, opened 2016-04-09
pangram: Add extra invalid pangram test case

helenst:
I reviewed a solution that was passing the tests by coincidence - they were performing an alphabetic string comparison between sets of letters, and the only non empty False test case was passing because the string ('a quick movement...') is alphabetically less than 'abcdef...'. It seemed like another test case would help here!

(See http://exercism.io/submissions/c6790b9a65fe4c08b865e5eb621fe94d)


kytrinyx:
This is a great find, thank you!

PR #333, opened 2016-04-28
palindrome-products: test fix

rootulp:
Remove redundant '3' because {3, 3} is equivalent to {3}

### DIFF
--- a/exercises/palindrome-products/palindrome_products_test.py
+++ b/exercises/palindrome-products/palindrome_products_test.py
@@ -20,7 +20,7 @@ class PalindromesTests(unittest.TestCase):
     def test_largest_palindrome_from_single_digit_factors(self):
         value, factors = largest_palindrome(max_factor=9)
         self.assertEqual(9, value)
-        self.assertIn(set(factors), [{1, 9}, {3, 3}])
+        self.assertIn(set(factors), [{1, 9}, {3}])
 
     def test_largest_palindrome_from_double_digit_factors(self):
         value, factors = largest_palindrome(max_factor=99, min_factor=10)


kytrinyx:
I don't see how `{3}` is equivalent to `{3, 3}`. We need two factors, both the first and second are 3, no?

rootulp:
Curly braces are shorthand for set notation. I also think that the test wants either `{1, 9}` or `{3}` because this test uses `self.assertIn` as opposed to `self.assertEqual`.

As a side note, I would've left `{3, 3}` for clarity in order to symbolize that both factors are indeed `3`. 
But a following test looks like this:
```python
     def test_smallest_palindrome_from_double_digit_factors(self):
         value, factors = smallest_palindrome(max_factor=99, min_factor=10)
         self.assertEqual(121, value)
         self.assertEqual({11}, set(factors))
```
So I fixed the `3` for consistency. Hope that makes sense.

kytrinyx:
Ah, that's what I get for not knowing Python.

In this case, a set might not actually be the right data type. We want the non-orderedness of sets, with the non-uniqueness of lists.

kytrinyx:
Actually, @petertseng, do you happen to have an opinion on this one?

petertseng:
There seems to (coincidentally or not) be a fair bit of discussion around palindrome products in the various tracks these days, given https://github.com/exercism/xruby/issues/305 also exist!

Note that `{101, 101}` on line 43 that also deserves the same treatment (whatever we decide the treatment should be).

I can see why the test uses `set(factors)`: The advantage is it avoids having to write both `[1, 9]` and `[9, 1]` which we would have to use if it were `list(factors)`. The disadvantage comes when we're testing a square number.

We have to be careful too, because I see at the top of the test file that it says "factors is an iterable containing both factors of the palindrome in arbitrary order."

Using `set(factors)` allows a solution that doesn't conform to this interface to pass the test (a solution can now return any nonzero number of threes, not just two). Obviously it probably *wouldn't* because it would take some unusual work to make it do that and simultaneously pass all other tests, but it *could maybe* happen.

So if we are very concerned that a non-conforming implementation might pass, it could mean that we would want to do something other than `set(factors)`. Has anyone (in this track or otherwise), seen a solution that warrants this concern?

petertseng:
Edit: Or just test that `len(factors) == 2` to avoid the weird non-conforming solutions I mentioned above

petertseng:
Edit2: 
Note a caveat with this is that this iterates over the `factors` twice, what if someone's solution returns an iterable that can only be iterated once? Again, I find this unlikely, but not outside the realm of possibility and not in violation of the spec. You could first convert `factors` to a list, check its length, turn the list into a set, and check it against the possible sets, but it seems the tests are getting a bit confusing at this point. So you have to decide what you want based on how likely you think submitted solutions will fall into the weird categories.

rootulp:
From the comment above the test:
> 
 Their return value must be a tuple (value, factors) where value is the
 palindrome itself, and factors is an iterable containing both factors of the
 palindrome in arbitrary order.

It seems like we're adding a lot of complexity to the tests in order to satisfy this interface. What if we changed the interface to expect a tuple of both factors in ascending order? Seems like the tests would be a lot simpler although the solution required would be slightly more constricting.

kytrinyx:
I think it would make sense to change the interface rather than jump through hoops in order to satisfy it.

petertseng:
Am going to unwatch on this thread while y'all deal with that, so that I can focus on other things, but feel free to mention again if something needs my attention.

kotp:
@petertseng When you are ready to jump back in, message me in chat (twitter, the gitter channel, gmail, whatever) and maybe we can knock both of these out together (xruby and xpython).

behrtam:
*Disclaimer: I didn't try to understand the problem or research this test case in other repos.*

@petertseng @kotp Did you figure this out somewhere else? Should this be closed?

kotp:
@behrtam Looks like it is on haitus.  Until someone does spend some time confused, then hopefully that will sway a solution, if any, if ever.

Issue #342, opened 2016-06-26
Sublist tests pass with no implementation


IanWhitney:
http://exercism.io/submissions/bc21579f3bf54f6d99a39093d8396157

A member of my Kata group reports that all tests pass with that implementation. It's, uh, very concise. But I don't think it's accurate.

patricksjackson:
The simple fix for this is an extra test for checking for unique return values, so I threw together the pull request for that.

That implementation above still passes all the other tests, so there may be a better way to do it.

PR #343, opened 2016-06-26
sublist: Adding test for unique return values

patricksjackson:
Adds a simple test for checking if the user-defined return values (SUBLIST, SUPERLIST, EQUAL, and UNEQUAL) have different values.

Simple fix to get around the problem raised in Issue #342


Dog:
Thanks

PR #349, opened 2016-07-21
bob: Adding non-question trailing whitespace test

### LINKED ISSUE #312: bob_test doesn't fully test bob specification
The bob test suite doesn't test the case of ending a non-question statement with whitespace.  You can see that by this code passing all given tests, but failing if a test case of a statement with trailing whitespace is given.

```
from string import letters

def hey(what):

    if not what or what.isspace():
    	return "Fine. Be that way!"

    elif what == what.upper() and not set(letters).isdisjoint(set(what)):
    	return "Whoa, chill out!"

    elif what.endswith(("?"," ")):
    	return "Sure."

    return "Whatever."
```

An example test case is 

```
def test_non_question_ends_with_whitespace(self):
        self.assertEqual(
            'Whatever.', bob.hey('This is a statement with trailing whitespace   ')
        )
```

patricksjackson:
As suggested by @matts-code

Resolves #312

### DIFF
--- a/exercises/bob/bob_test.py
+++ b/exercises/bob/bob_test.py
@@ -114,5 +114,10 @@ class BobTests(unittest.TestCase):
             'Sure.', bob.hey('What if we end with whitespace?   ')
         )
 
+    def test_non_question_ends_with_whitespace(self):
+        self.assertEqual(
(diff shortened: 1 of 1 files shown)


Dog:
Thanks

PR #350, opened 2016-07-30
Clock: remove duplicated tests

rootulp:
These four tests are repeated immediately above


kytrinyx:
Nice find!

CI has an odd failure on this one, saying that an equality test fails. Do you have a moment to look into it? It seems weird that it would be related to this change.

rootulp:
Turns out the second iteration of test_clocks_with_hour_overflow overwrote the first one. There was a discrepancy between testing for equality and inequality. I made the fix!

kytrinyx:
Sweet, thanks so much!

kytrinyx:
@Dog I'm going to go ahead and merge this one.

PR #351, opened 2016-08-03
Word-count test name fix

samdec11:
This test says "preserves_punction" when it's really testing that punctuation is being ignored (i.e. not included in the list of words being counted).


kytrinyx:
Good catch, thank you! :heart:

Issue #344, opened 2016-07-08
series, largest-series-product: Conflict when running all tests


patricksjackson:
`py.test` gives you the option of running all the tests at once, which is part of the reason why it's used instead of the simpler `unittest` (#309).

When I actually try executing it I hit a problem. It appears that the "largest-series-product" problem reused the filename in the series problem (Both tests use `from series import ...`, so both tests expect a `series.py`). Pytest ignores one of them since you can't have two libraries with the same name.

(Pytest _doesn't_ break if you use the example solution since it imports alphabetically and the example copies the needed function from the "series" problem, which might be what was intended when the problem was originally implemented, but that was taken out with #316.)

The fix for this would be trivial (change the import line in "largest-series-product), but I'm not sure if it's worthwhile breaking the ~380 existing solutions just to make the naming consistent.

Thoughts?

kytrinyx:
@Dog Would you mind taking a look at this?

Dog:
I'll be able to take a closer look after Siggraph.

Dog:
I believe it is a real issue, and that it is affecting more tests than just the two mentioned. For example, several exercises like rna-transcription and point-mutations expect a file dna.py. I did have py.test break with those two exercises, despite them passing in their respective folders.

Unfortunately I think the correct way to fix the problem may be to standardize the exercise files to follow the convention of <exercise_name>.py and <exercise_name>_test.py. This would also make it a lot easier to work with the repository as a whole. It was a headache when trying to copy all of the example.py files, and having to handle all of the special cases.

It wouldn't be a difficult change, but tedious. If we were to do that, we'd probably want to change all of them at once. The good news is that we aren't really changing functionality, just file names. This means a simple script could be written to migrate solutions.

For now I'd like to hold off on anyone picking this up for at least a week to get feedback and spend some more time considering the approach.

kytrinyx:
> I think the correct way to fix the problem may be to standardize the exercise files to follow the convention of .py and _test.py

For the API to work (in order to not deliver the example solutions to people) we need the filename to be called something with example in it.

In some of the other tracks we have a makefile that copies the exercises to a temporary directory and renames the example files to the expected file name. I don't know if that would be appropriate here.

Dog:
Its fine that the example solution is example.py. I'm suggesting we change the filename users create for their solution to be `<exercise_name>.py`. Sorry, I think its more clear what I am suggesting now that I have the markup fixed above.

kytrinyx:
[...]

PR #352, opened 2016-08-17
Standardizing exercise names #344

Dog:
This is a first pass


kytrinyx:
This looks good to me.

Dog:
Cool,

I want to do a second pass to make sure #344 really is fixed/imports make sense, and then write some more documentation in Contributions before I merge.

Issue #353, opened 2016-08-18
Update config.json to match new specification


kytrinyx:
For the past three years, the ordering of exercises has been done based on gut feelings and wild guesses. As a result, the progression of the exercises has been somewhat haphazard.

In the past few months maintainers of several tracks have invested a great deal of time in analyzing what concepts various exercises require, and then reordering the tracks as a result of that analysis.

It would be useful to bake this data into the track configuration so that we can adjust it over time as we learn more about each exercise.

To this end, we've decided to add a new key _exercises_ in the config.json file, and deprecate the `problems` key.

See exercism/discussions#60 for details about this decision.

Note that we will **not** be removing the `problems` key at this time, as this would break the website and a number of tools.

The process for deprecating the old `problems` array will be:

* Update all of the track configs to contain the new _exercises_ key, with whatever data we have.
* Simultaneously change the website and tools to support both formats.
* Once all of the tracks have added the _exercises_ key, remove support for the old key in the site and tools.
* Remove the old key from all of the track configs.

In the new format, each exercise is a JSON object with three properties:

* _slug_: the identifier of the exercise
* _difficulty_: a number from 1 to 10 where 1 is the easiest and 10 is the most difficult
* _topics_: an array of strings describing topics relevant to the exercise. We maintain
a list of common topics at https://github.com/exercism/x-common/blob/master/TOPICS.txt. Do not feel like you need to restrict yourself to this list;
it's only there so that we don't end up with 20 variations on the same topic. Each
language is different, and there will likely be topics specific to each language that will
not make it onto the list.

The _difficulty_ rating can be a very rough estimate.

The _topics_ array can be empty if this analysis has not yet been done.

Example:

    "exercises": [
      {
        "slug": "hello-world" ,
        "difficulty": 1,
        "topics": [
            "control-flow (if-statements)",
            "optional values",
            "text formatting"
        ]
      },
      {
        "difficulty": 3,
        "slug": "anagram",
        "topics": [
            "strings",
            "filtering"
        ]
      },
      {
        "difficulty": 10,
        "slug": "forth",
        "topics": [
            "parsing",
            "transforming",
            "stacks"
        ]
      }
    ]

It may be worth making the change in several passes:

1. Add the _exercises_ key with the array of objects, where _difficulty_ is 1 and _topics_ is empty.
2. Update the difficulty settings to reflect a more accurate guess.
3. Add topics (perhaps one-by-one, in separate pull requests, in order to have useful discussions about each exercise).

[...]

PR #354, opened 2016-08-20
Update config.json to match new specification

Roshanjossey:
> This addresses issue #353


kytrinyx:
Excellent thanks!

For reference, this addresses step 1 of #353 where topics and difficulty level are simple placeholders. The next steps are to adjust the difficulty levels and add topics.

PR #357, opened 2016-08-26
add topics to leap exercise in config.json

bordeltabernacle:
This addresses issue #353 

I was tempted to push the difficulty of this up to 2. Only because I seem to remember confronting this early on in my programming experience (with no CS background), and taking some time wrapping my head around the logic of it. The problem seems easy now, though I wonder if it only is because I know the answer, it may be good to have someone who has more recently attempted this to comment on it's difficulty.

I guess, as well, as more problems receive difficulty ratings, it will be easier to compare the levels of difficulty.

### DIFF
--- a/config.json
+++ b/config.json
@@ -76,6 +76,9 @@
       "slug": "leap",
       "difficulty": 1,
       "topics": [
+          "control-flow (if-else statements)",
+          "booleans",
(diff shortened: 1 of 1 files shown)


Dog:
Thanks

PR #341, opened 2016-06-22
binary-search implemented.

michaelkunc:

PR #359, opened 2016-09-02
style: Update README for style requirements

patricksjackson:
The Travis build fails if PRs don't pass the style checker. (See #341). This is just adding that requirement to the README section on style.


kytrinyx:
Thanks, this is a good clarification.

PR #360, opened 2016-09-08
binary-search: clean up

patricksjackson:
Cleans up a few issues remaining from adding the binary-search exercise.

- Removed extra `.cache` files
- Altered filenaming system to match other exercises
- Updated tests to match common json


patricksjackson:
I'll just note that without this patch, anyone doing the exercise needs to name their solution `example.py` unless they modify the test file.

Dog:
Thanks 👍

kytrinyx:
We're missing `binary-search` from the `"exercises"` key in the `config.json`.

Issue #364, opened 2016-09-28
Add Python 3.6 to the travis build


behrtam:
We should validate and finally add Python 3.6 to the travis test setup. https://github.com/exercism/xpython/blob/master/.travis.yml#L9

surajnarwade:
I would like to take this task.

behrtam:
Looks like the nightly build works with our existing code base: https://travis-ci.org/exercism/xpython/jobs/168941309

PR #366, opened 2016-10-03
Added Python 3.6 in travis

surajnarwade:


surajnarwade:
Oh it seems travis-ci do not support python 3.6, but I checked with docs, it tells me to add 'nightly' which points to '3.6-dev', should I add that ?

surajnarwade:
Fixes #364

kytrinyx:
Oh, sweet!

Should we allow the build to fail against `nightly`, @behrtam? Or merge as is, and if it becomes a problem we can adjust it.

patricksjackson:
Python 3.6 is still in beta. I expect that @behrtam wanted to add it once it is actually released rather than pointing at `nightly`, which is Pythons active development branch.

Using that is very different from adding a new version of Python to test against.

behrtam:
I mixed up the release of the beta 1 (the end of the feature development phase) with the actual final version. But adding the nightly to know where we stand does sound like a good idea.

```
matrix:
    allow_failures:
        - python: "nightly"
```

surajnarwade:
should I sent new patch or this is going to be closed :( ?

behrtam:
You can just update this PR.

PR #367, opened 2016-10-04
Isogram: Exercise for isogram with test cases.

sbulage:
Written python code which checks whether given string is isogram or not.

### DIFF
--- a/config.json
+++ b/config.json
@@ -64,9 +64,16 @@
     "poker",
     "zebra-puzzle",
     "rectangles",
-    "binary-search"
+    "binary-search",
+    "isogram"
   ],
   "exercises": [
+    {
+      "slug": "isogram",
+      "difficulty": 1,
+      "topics": [
+      ]
+    },
     {
       "slug": "hello-world",
       "difficulty": 1,
--- a/exercises/isogram/example.py
+++ b/exercises/isogram/example.py
@@ -1,0 +1,3 @@
+def isogram(string):
+    string = string.lower()
+    return len(set(string)) == len(string)
--- a/exercises/isogram/isogram_test.py
+++ b/exercises/isogram/isogram_test.py
@@ -1,0 +1,20 @@
+import unittest
+from example import isogram
+
+
+class TestIsogram(unittest.TestCase):
+
+    def test_isogram(self):
+        tests = ['lumberjacks', 'background', 'downstream']
+        for test in tests:
+            string = isogram(test)
+            self.assertTrue(string, True)
+
+    def test_isogram_false(self):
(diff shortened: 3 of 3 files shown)


sbulage:
@kytrinyx Kindly review my code.

kytrinyx:
@sbulage thanks for the contribution!

The build is failing with the following information in the log:

```
$ ./bin/configlet .
Evaluating .
-> config.json does not include [isogram].

The command "./bin/configlet ." exited with 1.
```

Configlet is an exercism-specific linter that checks some of the basic things about an exercise. The way to fix this is to add `"isogram"` to the "exercises" key in the `config.json` file: https://github.com/exercism/xpython/blob/master/config.json#L69

You can make the "difficulty" default to 1 and the "topics" an empty array. We will be filling in this information later.

The "exercises" array is ordered roughly by difficulty of the exercise, so "isogram" should probably be not too early, not too late. Maybe in the second quarter of exercises somewhere. It doesn't matter too much getting it exactly right, as we can adjust this later when we get some feedback from people who solve the exercise.

sbulage:
@kytrinyx Thanks for your suggestions. I made the required changes.

patricksjackson:
For adding `isogram` into the exercises array, I suggest placing it before pangram as they are very similar problems, and this actually serves as a very good warmup to that problem.

In addition, please add the rest of the common test cases from the x-common repo.  https://github.com/exercism/x-common/blob/master/exercises/isogram/canonical-data.json

The tests should all be broken up so that they test one word per test case. Look at how the other problems do it.

sbulage:
@patricksjackson Sure, I will write the test cases for the rest. :+1:

behrtam:
@sbulage Do you need any help completing this exercise? I'm happy to help.

sbulage:
@behrtam No, Thanks :)
There are some test cases which are returning isogram_test as False but below link it is showing as True as it contains some whitespaces.

Link --> https://github.com/exercism/x-common/blob/master/exercises/isogram/canonical-data.json

behrtam:
**A)**
Looks like white spaces don't count. So you need to either remove them first or find a way to ignore them.

**B)**
```python
from example import isogram
```
Than you need to fix your import. I would suggest to rename your function. 

```python
from isogram import is_isogram
```
You could keep both files (`example.py` and `isogram.py`) or use the test script (`./test/check-exercises.py`) which will create a temporary file from the example one.

**C)**
Finally you need to break up the different test cases and add the missing ones as @patricksjackson already pointed out. Than we are good to go and can merge this new exercise.

behrtam:
Any chance that you might find some time to finish this PR?

PR #374, opened 2016-10-17
Add flatten array exercise

treyhunner:
I'm not entirely sure whether I updated the `config.json` file appropriately.

One issue I have with this exercise currently: "array" is not a word we use in Python.  Could we rename this to "flatten-list" or just "flatten" (which I think would be best)?  Would that be problematic because the exercise is called something different for the other languages?

### DIFF
--- a/config.json
+++ b/config.json
@@ -28,6 +28,7 @@
     "largest-series-product",
     "kindergarten-garden",
     "grade-school",
+    "flatten-array",
     "roman-numerals",
     "space-age",
     "grains",
@@ -218,6 +219,12 @@
     },
     {
       "slug": "grade-school",
+      "difficulty": 1,
+      "topics": [
+      ]
+    },
+    {
+      "slug": "flatten-array",
       "difficulty": 1,
       "topics": [
       ]
--- a/exercises/flatten-array/example.py
+++ b/exercises/flatten-array/example.py
@@ -1,11 +1,17 @@
-from collections.abc import Iterable
+def is_iterable(thing):
+    try:
+        iter(thing)
+    except TypeError:
+        return False
+    else:
+        return True
 
 
 def flatten(iterable):
     """Flatten a list of lists."""
     flattened = []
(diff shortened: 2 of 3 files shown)


behrtam:
The `iter` builtin checks for the `__iter__` method or in the case of strings for the `__getitem__` method. So for this exercise you could rewrite the function like this:
```python
def is_iterable(thing):
    return hasattr(thing, '__iter__')
```
Than it would be possible to drop the `isinstance(item, (str ...` check. That would make the example implementation shorter. Not sure if this is a better way, just something to think about.

I think it would be a good idea to drop *array* from the exercise name. But it would have to be changed in the *x-common* repo first I guess.

The `config.json` looks good to me. The difficulty and topics can be updated later.

treyhunner:
The `iter` function isn't entirely equivalent to an `__iter__` check.

Since `__getitem__` is checked as well, we'd potentially be ignoring some types of iterables:

```pycon
>>> class A:
...     def __getitem__(self, x):
...         return 0
...
>>> a = A()
>>> hasattr(a, '__iter__')
False
>>> iter(a)
<iterator object at 0x7f510adc5668>
```

This also doesn't fix the string issue because strings implement `__iter__` in Python 3 (I'm not really sure why they didn't but lists/tuples did in Python 2).

As for "array", should I make a pull request on the common repo and see what they say about it?

behrtam:
Sorry, only checked Python2 (wired change).

A pull request over there should do it. But I would suggest we add the exercise with the old name and rename it later, because it may take some time.

treyhunner:
I opened an issue to rename it. I agree that it should be merged in the meantime.

behrtam:
Does anyone have different thoughts about the exact placing in the problems order?

treyhunner:
I had no idea where to put it so I placed it somewhere that seemed reasonable. Feel free to move it based on whatever logic is currently used to determine exercise order.

PR #379, opened 2016-10-22
Remove deprecated problems key

behrtam:
The **problems** key is deprecated in favour of **exercises**
so it can be removed.

We have confirmation from https://github.com/exercism/x-api/pull/137
that it is safe to *delete* the problems key.


kytrinyx:
:heart: this feels pretty nice :)

Issue #272, opened 2015-12-01
Skeleton files should be included in all problems to make it easier


gytdau:
I'm on the Python section, and I've found that only a few challenges include a skeleton file. This is helpful, since you can then start coding without the delay of creating your own file. Example: the Bob challenge comes with bob.py:

```
#
# Skeleton file for the Python "Bob" exercise.
#


def hey(what):

    return
```

It'd be great if a skeleton file like this was included in all of the challenges, to make things just a little bit easier for everyone.

behrtam:
I already had a similar feeling #210.

kytrinyx:
We originally put stub files in early exercises because some people are stumped about how to even start solving the problem without a stub file in place.

However, I've found that one important part of TDD for me has been to recognize categories of errors. Recognizing the type of error that comes from a missing file and a missing constant or definition seems like a fairly important step to me.

I like stub files in the early problems to help people get started, but once they are confident, I find that stub files are training wheels that are no longer necessary.

behrtam:
I see it exactly the other way around. While I agree that recognizing categories of errors is an important skill but once you have created 1-3 files and know how to find out the function signatures this task get's boring and is just an unnecessary step.

gytdau:
In any case, I see creating the file as a mild annoyance. It would be nice if the file was created for you, but this is not a huge issue.

behrtam:
I just started to work on the **Scala** track and it's really annoying to create the files. We should create them for all exercises. Mostly without any code, so we don't force specific ways of solving the problems.

kytrinyx:
Do you mean completely empty files? Or with empty function definitions?

behrtam:
The test suits define the *class* and/or *function* names, so I would add them as the developer has no choice about that. It's just tiring work that doesn't add any learning value.

kytrinyx:
That sounds reasonable to me.

lbain:
For what it's worth, I learned about how python file/function names can be imported from Exercism because the default file and functions weren't supplied. Hello World simply uses `import hello_world` and then explicitly calls `hello_world.hello()` from there. But the next exercise (leap year) uses `from year import is_leap_year` and it took me a bit of investigating to work out how it all fits together. Sure, it's probably less valuable after the first two or three, but needing to create new files did teach me something at first.

Dog:
[...]

PR #380, opened 2016-10-22
Add word-search exercise

behrtam:
I don't really like the example solution, but it works. Also I'm not sure if it is the right way to provide the `class Point` this way.

https://github.com/exercism/x-common/blob/master/exercises/word-search/description.md

### DIFF
--- a/config.json
+++ b/config.json
@@ -429,6 +429,14 @@
       "difficulty": 1,
       "topics": [
       ]
+    },
+    {
+      "slug": "word-search",
+      "difficulty": 6,
+      "topics": [
+        "strings"
+        "searching"
+      ]
     }
   ],
   "deprecated": [
--- a/config.json
+++ b/config.json
@@ -434,7 +434,7 @@
       "slug": "word-search",
       "difficulty": 6,
       "topics": [
-        "strings"
+        "strings",
         "searching"
       ]
     }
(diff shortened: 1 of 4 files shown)


patricksjackson:
I like providing the `Point` class. It could've been done with just having them return the coords as a tuple, but this way is more explicit.

LGTM 👍

pheanex:
Hey @behrtam, I've got two questions on this:
* Since in #272 you were quite in favor of very minimal skeleton files, why did you provide such a (huge)  skeleton file here?
* Aren't we limiting the user by forcing the class-structure? (I do not see yet the added value of the "Point" class/general class structure)

behrtam:
Well it's four months away, but I assume I did it that way because the test suit relies on the `class Point`. This is probably a unpleasant result of my example implementation which I normally write first.

Maybe if we would instead test with tuples that represent a point, we would not need a dedicated class for that, which also means we could reduce the skeleton file and the people solving this problem would have more freedom.

pheanex:
Ah ok that makes sense, thank you.
If its okay with you, I'll try to refactor it a bit and send a PR :smile:

pheanex:
Okay i created PR #420, please review.
Please note: I also changed the order of point tuples from coordinate-based (x,y) to a (at least to my mind) more intuitive (row,colum) notation. (If you prefer the coordinate-system, i'll update it).

PR #381, opened 2016-10-25
Add grep exercise

behrtam:
I didn't use regex on purpose, because the description doesn't mention it so I wanted to test how or if it would work without it.

https://github.com/exercism/x-common/blob/master/exercises/grep/description.md


kytrinyx:
Fun! I think this looks good. Want to fix that conflict? (Also, if you're playing Hacktoberfest this month, feel free to merge this when the conflict is fixed, or ping me so I can merge it).

behrtam:
Resolved the conflict.

PR #390, opened 2016-10-28
Add info to the About page.

ramiroluz:


docs/ABOUT.md
```diff
@@ -1,7 +1,47 @@
...
 
 The most common form of Python is compiled in C. This is often invisible to the beginning programmer, but if there are uses for which exceptionally fast implementation is needed then C extensions can be written to optimize Python execution.
 
 [Python is used extensively](https://www.python.org/about/apps/) in scientific computing, finance, games, networking, internet development, and in assembling pipelines of other programs.
 
 Python was started by Guido van Rossum in 1989. Its name is an homage to the comedy troupe Monty Python. Python 2 is used widely but support may end by 2020. Python 3 was introduced in 2008 and is beginning to be adopted more widely. They are similar, but users will encounter [some differences](http://blog.teamtreehouse.com/python-2-vs-python-3). Python development is shepherded by [The Python Software Foundation](https://www.python.org/about/) and there are active community-based user groups worldwide.
+
+### What python is good for?
```
kytrinyx: If we're going to have a separate section for this, then I think we should move the "Python is used extensively" bit down as the intro to this section.

docs/ABOUT.md
```diff
@@ -1,7 +1,47 @@
...
+
+- [Web and Internet Development](https://www.python.org/about/apps/#web-and-internet-development)
+- [Database Access](https://www.python.org/about/apps/#database-access)
+- [Desktop GUIs](https://www.python.org/about/apps/#desktop-guis)
+- [Scientific & Numeric](https://www.python.org/about/apps/#scientific-and-numeric)
+- [Education](https://www.python.org/about/apps/#education)
+- [Network Programming](https://www.python.org/about/apps/#network-programming)
+- [Software & Game Development](https://www.python.org/about/apps/#software-development)
+
+### Who uses python?
```
kytrinyx: I don't think we need this section—we're already linking to the about page, and if people want to know more they'll find it in that about link.

docs/ABOUT.md
```diff
@@ -1,7 +1,47 @@
...
+
+Some reasons to learn Python [Reasons to learn](http://www.skilledup.com/articles/reasons-to-learn-python):
+
+- It is easy to learn. Python was designed to help the developer not the machine.
+- It helps the novice programmer. Concepts used in python can be applied to other programming languages.
+- It is fun. Some projects like [Raspberry Pi](http://www.raspberrypi.org/faqs#introWhatIs) uses Python as the main language and can be used to create robots, remote controlled cars and video game consoles.
+- There are many job opportunities for Python developers in companies like Google, Yahoo!, Disney, Nokia, and IBM. Not to mention a huge amount of startups. 
+- It have a wide range of open source web application frameworks.
+
+Python official web site [python website](https://www.python.org) and [python documentation](https://docs.python.org/) documentation to learn more.
```
kytrinyx: This needs to be a full sentence.

ramiroluz:
Please take a look @kytrinyx, let me know if you have any improvement suggestions.

behrtam:
@kytrinyx as you already took a look at this, maybe you could check it again. If you don't have enough time right know, I will read through this end of next week.

docs/ABOUT.md
```diff
@@ -20,28 +20,12 @@ Python can be applied to many different classes of problems [what python is good
...
 
 - It is easy to learn. Python was designed to help the developer not the machine.
 - It helps the novice programmer. Concepts used in python can be applied to other programming languages.
 - It is fun. Some projects like [Raspberry Pi](http://www.raspberrypi.org/faqs#introWhatIs) uses Python as the main language and can be used to create robots, remote controlled cars and video game consoles.
-- There are many job opportunities for Python developers in companies like Google, Yahoo!, Disney, Nokia, and IBM. Not to mention a huge amount of startups. 
-- It have a wide range of open source web application frameworks.
 
-Python official web site [python website](https://www.python.org) and [python documentation](https://docs.python.org/) documentation to learn more.
+There are many job opportunities for Python developers in companies like Google, Yahoo!, Disney, Nokia, and IBM. Not to mention a huge amount of startups. It have a wide range of open source web application frameworks. Visit the Python official web site [python website](https://www.python.org) and [python documentation](https://docs.python.org/) documentation to learn more.
```
kytrinyx: I would use "number of" instead of "amount of" when talking about startups, since startups are countable rather than measurable.

PR #392, opened 2016-11-03
Change to config.json based testing

behrtam:
So `./bin/configlet` only checks that for every exercises in the `config.json` there is also an example implementation and vice versa. It doesn't check if there is also a test file for every exercise.
The current test runner (`./test/check-exercises.py`) searches for all existing `*_test.py` files, so we won't get an error if we are missing a test file, put it in the wrong path or did name the file incorrectly (#294).

An attempt to fix this was started (#296) but not finished. So I adapted the new config format and changed the test runner in such a way, that it will detect missing test files.
With this new error case I also added more details to the failure message, so that it is clear `FAILURES:  hello-world (TestFailed), bob (FileNotFound)` what caused the problem.

### DIFF
--- a/test/check-exercises.py
+++ b/test/check-exercises.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
+from __future__ import print_function
+
 import os
 import ast
 import imp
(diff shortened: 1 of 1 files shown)


behrtam:
@pheanex Is there any chance (time/interest) you can review this?

pheanex:
Sure, I'd like to point out the following suggestions for improvement:
* Remove duplicated code running the tests (check single exercise vs all from json.config). 
* Rename exercise-filename from generic "name" to make it a bite more expressive
* We could make the code Flake8 cleaner (It suggests another newline before (main)-function)
* Remove unused "deprecated_problems"-variable by assigning "_"
I created a PR to your repo+branch with the suggested improvements implemented :smile:

PR #396, opened 2016-11-22
rna_transcription: Add invalid input support

behrtam:
Now reflects the canonical test data defined in x-common (see [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/rna-transcription/canonical-data.json)).

To work with the 3 new test cases I had to add a way to detect not valid input to the example implementation. While there are many ways to do this (*regex* being one of them) I thought the `if any(char not in valid_chars ...` construct is maybe the most self explanatory.

**Note:** *It would be great to get this merged before this weekend (November 26th) as there is a [code hangout](https://codebuddies.org/hangout/8An374DNxwZ38PJeF) planned.*

### DIFF
--- a/exercises/rna-transcription/example.py
+++ b/exercises/rna-transcription/example.py
@@ -6,8 +6,13 @@ else:
     maketrans = str.maketrans
 
 
-DNA_TO_RNA = maketrans('AGCT', 'UCGA')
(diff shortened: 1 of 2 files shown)


kytrinyx:
I'll make sure this gets deployed today.

PR #398, opened 2016-11-24
Align header with README.md headers

mienaikage:
Header is not consistent with other headers in README.md

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,4 +1,4 @@
-### Submitting Exercises
+## Submitting Exercises
 
 Note that, when trying to submit an exercise, make sure the solution is in the `exercism/python/<exerciseName>` directory.
 


behrtam:
```
## Instructions

Submissions are encouraged to be general, within reason. Having said that, it's also important not to over-engineer a solution.

It's important to remember that the goal is to make code as expressive and readable as we can. However, solutions to the hello-world exercise will be not be reviewed by a person, but by rikki- the robot, who will offer an encouraging word.

### Submitting Exercises

Note that, when trying to submit an exercise, make sure the solution is in the `exercism/python/<exerciseName>` directory.

For example, if you're submitting `bob.py` for the Bob exercise, the submit command would be something like `exercism submit <path_to_exercism_dir>/python/bob/bob.py`.

## Source

This is a program to introduce users to using Exercism [view source](http://en.wikipedia.org/wiki/%22Hello,_world!%22_program)
```

To me it clearly looks like a sub-point of *Instructions*, so `###` would be the correct headline level.

kytrinyx:
I think this makes sense as a sub-point of _Instructions_ as well.

PR #406, opened 2017-01-26
atbash_cipher_test: Add 2 tests; leap: change all tests

jgomo3:
Add `test_decode_number` as there is a test that encode number. Check this _iteration_[1] for a solution that pass the tests without checking for alphanumeric in decode.

Add `test_encode_decode`. Having done this in the first place, would had highlighted the need of checking for alphanumeric in decoding.

[1] http://exercism.io/submissions/c7c9448e6a674dd6a11ea35dc32e29a4

### DIFF
--- a/exercises/atbash-cipher/atbash_cipher_test.py
+++ b/exercises/atbash-cipher/atbash_cipher_test.py
@@ -42,6 +42,18 @@ class AtbashCipherTest(unittest.TestCase):
             decode("zmlyh gzxov rhlug vmzhg vkkrm thglm v")
         )
 
+    def test_decode_numbers(self):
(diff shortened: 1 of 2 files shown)


behrtam:
Great changes! Would it be possible to split them, so that we have one PR per exercise?

jgomo3:
Sure. What do you suggest?
Should this PR be cancelled and I should create 2 branches, one with only one commit for each change, and make a PR for each one? Or is there a better way?

behrtam:
I don't know an easier way than to create two new branches/prs.

jgomo3:
I created 2 separate pull requests, one for each commit in this one. They are:

#418 and #419.

Issue #402, opened 2017-01-11
Changes to Luhn tests


IanWhitney:
We recently rewrote the test suite for the Luhn exercise. Since this track implements Luhn, please [take a look at the new canonical_data.json file](https://github.com/exercism/x-common/blob/master/exercises/luhn/canonical-data.json) and see if your track should update its implementation or tests.

Details on the test changes and the motivation can be found in [the PR](https://github.com/exercism/x-common/pull/474).

Tracking of these changes is being done at [the tracking issue](https://github.com/exercism/x-common/issues/491).

PR #409, opened 2017-01-29
luhn: Update tests and example

rootulp:
- References #402

### DIFF
--- a/exercises/luhn/luhn_test.py
+++ b/exercises/luhn/luhn_test.py
@@ -45,6 +45,15 @@ class LuhnTests(unittest.TestCase):
     def test_input_digit_9_is_correctly_converted_to_output_digit_9(self):
         self.assertTrue(Luhn("091").is_valid())
 
+    def test_is_valid_can_be_called_repeatedly(self):
+        # Additional track specific test case
+        # This test was added, because we saw many implementations
+        # in which the first call to is_valid() worked, but the
+        # second call failed().
+        number = Luhn(8739567)
+        self.assertTrue(number.is_valid())
+        self.assertTrue(number.is_valid())
+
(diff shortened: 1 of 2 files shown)


behrtam:
We should keep an eye on those `luhn` related issues:
https://github.com/exercism/x-common/pull/522
https://github.com/exercism/x-common/issues/523

rootulp:
Good point. I expect the canonical-data to change so we should keep this open until that gets resolved.

behrtam:
https://github.com/exercism/x-common/pull/547

rootulp:
@behrtam sorry for the slow response! I updated the canonical tests.

exercises/luhn/luhn_test.py
```diff
@@ -1,48 +1,49 @@
...
-        self.assertEqual(8739567, Luhn.create(873956))
+    def test_single_zero_with_space_is_invalid(self):
+        self.assertFalse(Luhn("0").is_valid())
 
-    def test_create_valid_number3(self):
-        self.assertEqual(8372637564, Luhn.create(837263756))
+    def test_more_than_a_single_zero_is_valid(self):
+        self.assertTrue(Luhn("0000 0").is_valid())
 
-    def test_is_valid_can_be_called_repeatedly(self):
```
behrtam: I would like to leave this mutability test case (#136) as it is.
Maybe we could put it as the last test with a comment (`# additional track specific test case`).

rootulp: Added back with the comment you suggested. I replaced the test value with `"055 444 285"` because it exists in the canonical data.

behrtam:
Thanks for your work!

Issue #389, opened 2016-10-28
Add special python methods to linked-list


behrtam:
In the original pr that added the **linked-list** problem it was suggested to additionally implement some of the special python functions (see _Extra Credit Ideas_ https://github.com/exercism/xpython/pull/212#issuecomment-154601145).

So it would be nice to add methods like `__iter__` and `__len__` (maybe more, if they make sense) with skipped tests (`@unittest.skip("advanced")`) so that people who want to do more can optionally work on those.

PR #408, opened 2017-01-28
linked-list: Extra credit iter() and len() 

rootulp:
- Add tests for `__iter__()` and `__len__()`
- References #389 
- Commits have additional context

PR #410, opened 2017-01-30
linked-list: Issues #389. Add special Python functions __iter__ and _…

leec1979:
…_len__ along with their tests. Tests are skipped by default so that the new functionality is treated as optional extra work, as per the issue.

### DIFF
--- a/exercises/linked-list/example.py
+++ b/exercises/linked-list/example.py
@@ -45,3 +45,20 @@ class LinkedList(object):
             new_node.next = self.head
             self.head.prev = new_node
             self.head = new_node
+    def __len__(self):
+        count = 0
+        current = self.head
+
+        while current is not None:
+            count += 1
+            current = current.next
+
+        return count
+
+    def __iter__(self):
+        current = self.head
+        while current is not None:
+            yield current
+            current = current.next
+
+        return
--- a/exercises/linked-list/example.py
+++ b/exercises/linked-list/example.py
@@ -45,6 +45,7 @@ class LinkedList(object):
             new_node.next = self.head
             self.head.prev = new_node
             self.head = new_node
+
     def __len__(self):
         count = 0
         current = self.head
--- a/exercises/linked-list/linked_list_test.py
+++ b/exercises/linked-list/linked_list_test.py
@@ -44,6 +44,25 @@ class LinkedListTests(unittest.TestCase):
         self.assertEqual(50, self.list.pop())
         self.assertEqual(30, self.list.shift())
 
+    @unittest.skip("advanced")
+    def test_len(self):
+        self.assertEqual(0, len(self.list))
+        self.list.push(10)
+        self.list.push(20)
+        self.assertEqual(2, len(self.list))
+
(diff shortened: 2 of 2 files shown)


exercises/linked-list/example.py
```diff
@@ -45,3 +45,21 @@ def unshift(self, value):
             new_node.next = self.head
             self.head.prev = new_node
             self.head = new_node
+
+    def __len__(self):
```
rootulp: I think this implementation of `__len__` is cleaner than mine #408 but it requires O(n) to run when a call to `__len__` on a linked list could be O(1). Fortunately, this exercise doesn't mention anything about minimizing algorithmic complexity.

exercises/linked-list/example.py
```diff
@@ -45,3 +45,21 @@ def unshift(self, value):
...
+
+        return count
+
+    def __iter__(self):
+        current = self.head
+        while current is not None:
+            yield current
+            current = current.next
+
+        return
```
rootulp: Why the final `return`?

exercises/linked-list/linked_list_test.py
```diff
@@ -44,6 +44,26 @@ def test_all(self):
...
+        self.assertEqual(2, len(self.list))
+
+    @unittest.skip("advanced")
+    def test_iter(self):
+        values = [10, 20, 30, 40]
+        for value in values:
+            self.list.push(value)
+
+        for index, node in enumerate(self.list):
+            self.assertEqual(values[index], node.value)
```
rootulp: Why are we making this assertion? It seems unrelated to testing the functionality of `__iter__`

leec1979: It's testing that iterating over the linkedlist matches the values in the original list used to populate the linkedlist. Possibly a cleaner way to write that test would have been:

self.assertEqual(values, [x.value for x in self.list])

It's redundant due to the explicit iterator tests that follow but I forgot to remove it.

exercises/linked-list/example.py
leec1979: That's a hangover from writing in other languages recently.

exercises/linked-list/example.py
leec1979: Thanks for the review. I didn't spot your pull request until after I'd submitted mine - I'm still getting used to github.

My approach is more 'self contained' but worse from a big O perspective. From a teaching perspective it would be good to include both and document the pros/cons of each approach.

rootulp: I definitely agree on your approach being more self contained

exercises/linked-list/example.py
rootulp: Sounds like you can remove it :)

exercises/linked-list/linked_list_test.py
rootulp: Also safe to remove this

behrtam:
Thanks for your work!

Unfortunately #408 already implemented those two methods. As it was earlier and conflict free I merged that PR instead.

PR #413, opened 2017-02-12
list-ops: fix reversed tuple test

pheanex:
* i'd expect a reversed empty tuple to be an empty tuple and not an empty list

### DIFF
--- a/exercises/list-ops/example.py
+++ b/exercises/list-ops/example.py
@@ -1,13 +1,13 @@
 def map_clone(function, xs):
-    return (function(elem) for elem in xs)
+    return [function(elem) for elem in xs]
 
 
 def length(xs):
-    return sum(1 for x in xs)
+    return sum(1 for _ in xs)
 
 
 def filter_clone(function, xs):
-    return (x for x in xs if function(x))
+    return [x for x in xs if function(x)]
 
 
 def reverse(xs):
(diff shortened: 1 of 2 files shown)


behrtam:
What makes you thing that the test is about tuples? Based on the canonical test data, the test case is about lists: [../list-ops/canonical-data.json#L133-L139](https://github.com/exercism/x-common/blob/master/exercises/list-ops/canonical-data.json#L133-L139) So, maybe we could or even should improve the test name.

pheanex:
ah ok sorry, then i guess it should throw an error, if you put in an emtpy tuple?

behrtam:
I just looked at the test again and realized that got it all wrong, I thought it takes an empty list and expects an empty list, but it doesn't. Sorry!

We could just copy the approach from an earlier test ...

```python
def test_map_empty(self):
  self.assertEqual([], list(list_ops.map_clone(operator.index, [])))
```
... and change the other empty tests (`test_reverse_empty`, `test_flatten_empty`, `test_concat_empty`) to cast the result to *list* as well.

Another option would be to remove the casting all ofter the test suite and just use lists as in-&output all the time. The exercise is called **list**-ops and the canonical test data is based only on lists.

What do you think?

pheanex:
I suggest option 2 because its simpler

behrtam:
Would you mind going the extra mile and also remove the *tuple stuff* from the first 3 map tests and the `test_pos_leng` test.
Also the implementation of `filter_clone` should return a list, instead of a tuple.

Than the exercise is 100% `lists` which hopefully reduces possible misunderstandings in the future.

pheanex:
Done, now the code is also PEP8 clean

PR #415, opened 2017-02-16
Add skeleton files for exercises

pheanex:


exercises/linked-list/linked_list.py
```diff
@@ -1,13 +1,8 @@
...
-        self.value = value
-        self.next = next
-        self.prev = prev
+class Node:
+    def __init__(self):
+        pass
 
 
-class LinkedList(object):
+class LinkedList:
```
behrtam: As we support Python 2 & 3 we should subclass from `object` to have new style classes in both versions. That goes for all other cases as well.

behrtam:
You could use one of the available keywords (https://help.github.com/articles/closing-issues-via-commit-messages/) in your commit message to auto close issue #272 when this gets merged.

exercises/linked-list/linked_list.py
pheanex: Done

pheanex:
Ah thank you for the keyword tip. I forgot about that one :-)

PR #420, opened 2017-03-01
refactor exercise word-search

pheanex:
* rewrite tests so that users have more freedom of implementation
* update example-implementation to fit test-interface
* reduce skeleton to minimal as discussed in #272

### DIFF
--- a/exercises/word-search/example.py
+++ b/exercises/word-search/example.py
@@ -1,57 +1,22 @@
-import copy
+def find_stop(row, column, word, puzzle):
+    directions = [(0, 1), (0, -1), (1, 0), (1, 1), (1, -1), (-1, 0), (-1, 1), (-1, -1)]
+    for d in directions:
+        row_number, column_number = row, column
+        for char_number, character in enumerate(word):
+            try:
+                if puzzle[row_number][column_number] != character:
+                    break
+            except IndexError:
+                break
+            if char_number == len(word) - 1:
+                return row_number, column_number
+            row_number += d[0]
+            column_number += d[1]
 
 
-class Point(object):
-    def __init__(self, x, y):
-        self.x = x
-        self.y = y
-
-    def __repr__(self):
-        return 'Point({}:{})'.format(self.x, self.y)
-
-    def __add__(self, other):
-        return Point(self.x + other.x, self.y + other.y)
-
-    def __sub__(self, other):
-        return Point(self.x - other.x, self.y - other.y)
-
-    def __eq__(self, other):
-        return self.x == other.x and self.y == other.y
-
-    def __ne__(self, other):
-        return not(self == other)
-
-
-DIRECTIONS = (Point(1, 0), Point(1, -1), Point(1, 1), Point(-1, -1),
-              Point(0, -1), Point(0, 1), Point(-1, 1), Point(-1, 0))
-
-
-class WordSearch(object):
-    def __init__(self, puzzle):
-        self.rows = puzzle.split()
-        self.width = len(self.rows[0])
-        self.height = len(self.rows)
-
-    def find_char(self, coordinate):
-        if coordinate.x < 0 or coordinate.x >= self.width:
-            return
-        if coordinate.y < 0 or coordinate.y >= self.height:
-            return
-        return self.rows[coordinate.y][coordinate.x]
-
-    def find(self, word, position, direction):
-            current = copy.copy(position)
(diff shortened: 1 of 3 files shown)


exercises/word-search/example.py
```diff
@@ -1,57 +1,22 @@
...
-        for pos in positions:
-            for d in DIRECTIONS:
-                result = self.find(word, pos, d)
-                if result:
-                    return result
-        return None
+def find_stop(row, column, word, puzzle):
+    directions = [(0, 1), (0, -1), (1, 0), (1, 1), (1, -1), (-1, 0), (-1, 1), (-1, -1)]
+    for d in directions:
+        row_nr, column_nr = row, column
```
rootulp: Does `nr` refer to number? If so, why not make this explicit with `row_number` and `column_number`?

pheanex: yes it refers to number, i always try to find the sweet spot between overview/compactness and expressiveness ;-). I'll just update it if you think the gain in expressiveness is worth the loss in compactness and brevity

rootulp: It's a small nitpick and it's entirely up to you. I was confused by the `_nr` and I'm only pointing it out in the hope that others can avoid similar confusion.

pheanex: okay then it absolutely makes sense to change it, thank you for the feedback/review.

exercises/word-search/word_search_test.py
```diff
@@ -1,83 +1,53 @@
...
-                  'oivokprjsm\n'
-                  'pbwasqroua\n'
-                  'rixilelhrs\n'
-                  'wolcqlirpc\n'
-                  'screeaumgr\n'
-                  'alxhpburyi\n'
-                  'jalaycalmp\n'
-                  'clojurermt')
-        self.example = WordSearch(puzzle)
+    puzzle = ['jefblpepre',
```
behrtam: Why did you change the input format? I mean it's only a `split()` away ... just would like to know the reasoning behind this.

exercises/word-search/word_search_test.py
```diff
@@ -1,83 +1,53 @@
...
+              'alxhpburyi',
+              'jalaycalmp',
+              'clojurermt']
 
     def test_horizontal_words_left_to_right(self):
-        self.assertEqual(
-            self.example.search('clojure'),
-            (Point(0, 9), Point(6, 9))
-        )
+        self.assertEqual(search(WordSearchTests.puzzle, 'clojure'), ((9, 0), (9, 6)))
```
behrtam: Why did you reverse the `x, y` order? Normally coordinate systems use `(X, Y)` notation if I'm not mistaken.

pheanex:
Is there any way we can get a third reviewer? @behrtam any suggestions?

behrtam:
Maybe @rootulp has some time to look at it again. (Or do you want a fourth?)
It would help every other reviewer if you could explain your motivations behind the changes I asked you about.

pheanex:
I thought i did explain my motivation (see above). If there is still something that you think i should motivate further, please specify.

behrtam:
I appreciate the effort you have put into improving this exercise from a heavy class based exercise into a lighter flexible function based exercise and hope my maybe annoying questions about some details have not discouraged you.

You switched `(0, 9) -> (9, 0)` so from the typical `(X, Y)` to `(Y, X)` order which is just a small detail, but why? The only explanation I can find ist "freedom of implementation" which for me is not relevant for the `X/Y order` and for the `list vs string input` topic.

pheanex:
No problem.
I tried to say it with "I had to change the order to make it consistent with the naming (row/column)."
and because it is a more straight forward representation to me. That is why i wanted to know how others felt about it. (Hence the question with the o x o x o o x ...).
Also i threw out the newlines (\n) since i thought it is cleaner and simpler without them. (So to just keep whats absolutely necessary and i felt newlines were not absolutely necessary.)

behrtam:
So, the canonical test data is on it's way: https://github.com/exercism/x-common/pull/755

behrtam:
Maybe we could show off `namedtuples` in this exercise?

```python
>>> from collections import namedtuple
>>> Cell = namedtuple('Cell', ['row', 'column'])
>>> start = Cell(2, 2)
>>> end = Cell(row=2, column=12)
```

pheanex:
I guess with canonical test data we can close this

PR #422, opened 2017-03-05
Bowling: Implement bowling exercise

### LINKED ISSUE #421: Implement bowling exercise

jdh747:
First PR (all going well) closes #421

Changes:
- Added exercises/bowling/bowling_test.py (generated from [x-common/exercises/bowling/canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/bowling/canonical-data.json))
- Added exercises/bowling/example.py
- Updated config.json to include exercise (using [xcsharp/config.json](https://github.com/exercism/xcsharp/blob/master/config.json) as a template)

Passing 28 unit tests using pytest:
screenshot from 2017-03-05 16-10-54

And using test/check-exercises.py:
screenshot from 2017-03-05 16-13-27

The test harness exceeded 99 characters on several lines, breaking pep8 compliance:
screenshot from 2017-03-05 16-26-00

Test names were generated directly from the canonical-data.json description of the exercise, but I can amend these if need be.

Let me know if amendments need to be made to tests, example or commit message and I'll push necessary updates.

### DIFF
--- a/config.json
+++ b/config.json
@@ -442,6 +442,14 @@
         "strings",
         "searching"
       ]
+    },
+    {
+      "slug": "bowling",
+      "difficulty": 6,
+      "topics": [
+        "Algorithms",
+        "Control-flow (loops)"
+      ]
     }
   ],
   "deprecated": [
--- a/exercises/bowling/bowling_test.py
+++ b/exercises/bowling/bowling_test.py
@@ -1,0 +1,180 @@
+import unittest
+from bowling import BowlingGame
+
+
+class BowlingTest(unittest.TestCase):
+    def test_should_be_able_to_score_a_game_with_all_zeros(self):
+            rolls = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+            game = self._roll_many(rolls)
+            self.assertEqual(game.score(), 0)
+
+    def test_should_be_able_to_score_a_game_with_no_strikes_or_spares(self):
+            rolls = [3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6, 3, 6]
+            game = self._roll_many(rolls)
+            self.assertEqual(game.score(), 90)
+
+    def test_a_spare_followed_by_zeros_is_worth_ten_points(self):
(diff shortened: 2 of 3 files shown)


behrtam:
Looks good, only some small things.

**exception tests**

I think asserting the exception message text goes a bit too far. Just testing for the exception should be enough or we could test that there is at least a message. Maybe we could group the exceptions some how and create better exceptions instead of relying on detailed messages like in the `circular-buffer` exercise.

**max line length**
The README still states that we try to stay below 100 characters (`flake8 --max-line-length=99`). But we want to go full PEP8 compliant in the future, so it would be nice if you could reformat the files to stay below 80 chars.

There is one small additional style problem: `bowling.py:2:9: W292 no newline at end of file`

**Edit:** Okay, so we don't get around solving this with a `class` because of the requirements mentioned in the [description](https://github.com/exercism/x-common/blob/master/exercises/bowling/description.md#requirements).

jdh747:
Ok, PEP8 compliant formatting and custom exceptions over exception messages are a relatively easy fix.

I chose the approach of exception messages because it was significantly easier to write a test generator from [x-common canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/bowling/canonical-data.json) that generalises well. I'm not sure whether that's a factor, or whether there's a way of submitting generators to the project...

On the topic of class- versus function-based architecture, I considered both. Being a game that maintains state, I figured a class encapsulating the interface mentioned in the [description](https://github.com/exercism/x-common/blob/master/exercises/bowling/description.md#requirements) with methods was a more elegant implementation than a highly coupled procedural implementation that required global state variables for access by "roll" and "score" functions. The [xcsharp implementation](https://github.com/exercism/xcsharp/tree/master/exercises/bowling) that I used as reference took a similar class-based approach; however, they pass the entire game and perform all game logic in the "score" method with a complete view. Since the "roll" method isn't state-aware throughout the game, they're unable to implement some of the test cases described in [canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/bowling/canonical-data.json). In utilising the BowlingGame class, I attempted to create a state machine that accommodates all canonical examples.

Out of curiosity, when making changes to PR, is it best to push additional commits, or to squash new commits into prior work with a rebase and force push back up to my fork?

behrtam:
Just push additional commits. We can easily squash at the end.

PR #426, opened 2017-03-08
Fix python version used to run tests

behrtam:
Even if the test runner itself is executed with a specific python version the subprocess used to execute the respective test script always uses the generic 'python' found in the PATH. On most systems that have python2 & 3 installed 'python' is a symlink to 'python2.X'. So this would mean that even if you would run 'python3 ./test/...py' the tests would still be run with python2.

### DIFF
--- a/test/check-exercises.py
+++ b/test/check-exercises.py
@@ -12,6 +12,10 @@ import tempfile
 import json
 
 
+def python_executable_name():
+    return 'python{}.{}'.format(sys.version_info.major, sys.version_info.minor)
+
+
 def check_assignment(name, test_file):
     # Returns the exit code of the tests
     workdir = tempfile.mkdtemp(name)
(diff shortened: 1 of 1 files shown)


rootulp:
Good catch! Do you think we could've caught this sooner if we included a DEBUG log line that prints the version of Python that is used to test? Do you think it is worth adding this log line now? I'm torn because the `check-exercises.py` output is very concise but we explicitly ask contributors to [test](https://github.com/exercism/xpython#working-on-the-exercises) with both versions and an additional log line might serve to validate that the version they have tested against did in fact work (ex. 2.7 and 3.5).

behrtam:
I like the idea of adding a log line. If we add one line right before the FAILURE/SUCCESS message I don't think that would ruin the concise output.

rootulp:
Looks good to me 👍

PR #433, opened 2017-03-19
Run skipped tests

### LINKED ISSUE #391: Run skipped tests

rootulp:
Resolves #391
Inspiration from [Skipping tests in unittests](https://sateeshkumarb.wordpress.com/2011/06/16/skipping-tests-in-unittests/)

### DIFF
--- a/exercises/allergies/allergies_test.py
+++ b/exercises/allergies/allergies_test.py
@@ -33,14 +33,14 @@ class AllergiesTests(unittest.TestCase):
                     'chocolate pollen cats').split()),
             sorted(Allergies(255).lst))
 
-    @unittest.skipUnless('-extra_credit' in sys.argv, 'Skipping extra_credit')
+    @unittest.skipUnless('--extra-credit' in sys.argv, 'Skipping Extra Credit')
     def test_ignore_non_allergen_score_parts(self):
         self.assertEqual(['eggs'], Allergies(257).lst)
 
 
 if __name__ == '__main__':
     try:
-        sys.argv.remove('-extra_credit')
+        sys.argv.remove('--extra-credit')
     except ValueError:
         pass
     unittest.main()
--- a/exercises/allergies/allergies_test.py
+++ b/exercises/allergies/allergies_test.py
@@ -1,3 +1,4 @@
+import sys
 import unittest
 
 from allergies import Allergies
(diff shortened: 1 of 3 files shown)


behrtam:
Oh, that looks great. I will look at it more closely tomorrow when there is more time.

*Note: The max line length in this repo is relaxing 99 instead of 79 ... (https://github.com/exercism/xpython/blob/master/.travis.yml#L21)

behrtam:
The convention (maybe because of GNU *getopt*) is that single letter options are 
preceded by single dash and multi-letter options are preceded by double dash.
So could you make a small cosmetic change `-extra_credit` -> `--extra-credit`

I played with `skipUnless` at some point but I missed the `sys.argv.remove` hack.

rootulp:
> So could you make a small cosmetic change -extra_credit -> --extra-credit

I think this is a great change!

> I played with skipUnless at some point but I missed the sys.argv.remove hack.

I agree this does seem hacky. Basically we don't want to pass any sys.args to unittest but I couldn't find a better way to do this. Please let me know if you have any ideas.

>Don't know but do you think we should add a blue print for this to the README or
somewhere else if someone in the future also wants to add skipped tests or do you
think they will find it in the existing exercises?
Ad

The README currently doesn't have any docs on writing tests so I'm not sure where this would go. In the short-term I think contributors can use these two exercises as a template but I agree - this should be documented somewhere.

behrtam:
Well instead of removing the value from `sys.argv` you could pass an explicit argument vector like this:
```python
if __name__ == '__main__':
    unittest.main(argv=[''])
```

behrtam:
There is also a way that works without changing the existing test code by monkey-patching the `unittest.skip` decorator and using the test *discovery* modus.

```
$ python exercises/allergies/allergies_test.py
...s...
Ran 7 tests in 0.001s

OK (skipped=1)

$ python -c "import unittest as u; u.skip = lambda r: lambda t: t; u.main(None, argv=['', 'discover', '-p', '*test.py', '/Users/tmo/github/exercism/xpython/exercises/allergies/'])"
.......
Ran 7 tests in 0.001s

OK
```

rootulp:
I'm comfortable with both approaches. The monkey patching approach has the benefit of not maintaining an `EXTRA_CREDIT_ASSIGNMENTS` list. Also exercise implementers are free to create optional tests using any pattern.

On the other hand, the `--extra-credit` approach forces exercise implementers to conform to using the following pattern:
```python
@unittest.skipUnless('--extra-credit' in sys.argv, 'Skipping Extra Credit')
```
which would lead to consistency between exercises. This way users can always run optional test with the `--extra-credit` flag.

rootulp:
@behrtam any updates here?

behrtam:
Should find some time at the weekend to look at this again.

PR #437, opened 2017-03-19
Update Installation with supported Python versions

rootulp:
- Include a header


behrtam:
LGTM

Was only wondering why we have the paragraph about packages. This `doc` is for exercise users (not track maintainer) if I'm not mistaken who don't need to install any package to solve the exercises if they don't want to.

rootulp:
Are you aware of any exercises that require a package installation? I'm happy to remove the packages section if you think it's not necessary.

behrtam:
We only use the standard library no external packages at all. So, I think that we should remove it.

Issue #450, opened 2017-03-23
Update grains exercise (tests and example)


behrtam:
The exercise **grains** is missing some exception tests that are part of the canonical test data.
https://github.com/exercism/x-common/blob/master/exercises/grains/canonical-data.json#L55-L72

To add those tests the example implementation also needs to be updated. This task might be similar to the changes done in PR 439: https://github.com/exercism/xpython/pull/439/files

cmlang42:
I will take a look at this.

PR #453, opened 2017-03-24
grains: update exercise with missing exceptions

cmlang42:
grains: update exercise with missing exceptions

From Update grains exercise (tests and example) issue #450,
it was noted that 3 tests for exceptions were missing
- adds the missing test to the grains_test.py
- adds the appropriate exception handling and definitions to example.py
- makes the input parameter on the on_square function consistent naming, square


cmlang42:
Local changes worked for me, but I see the CI build could not complete with complaints about an environment variable.

cmlang42:
I should probably close this PR and recreate a new one with a clean commit without the merge commits.

behrtam:
We can clean that up later in this PR, no problem.

cmlang42:
I have locally resolved the whitespace with the flake8 recommendations. Except one. Wanted to ask about the second from/import line I have added. Of course flake complains about the line being too long and I am not one to be too concerned with a line a few characters over the "limit". 

But I did have a question that might end up resolving that anyway.  In the example solution I created three separate exception types, one for each of the three new assert checks (board boundaries). Did we want this to be the three separate exceptions or just one exception indicating that board boundaries were exceeded?  i.e. should I replace the three custom exceptions with one like BoardBoundaryError?

behrtam:
Maybe the built-in `ValueError` with a good message that explains the error is enough like in [`all-your-base`](https://github.com/exercism/xpython/blob/c9307c50c24e9166835ba999b5bc151ba32cb5ed/exercises/all-your-base/example.py), [`hamming `](https://github.com/exercism/xpython/blob/6fa2bd8a3c857a770bd87cbf208e1fcced5dee18/exercises/hamming/example.py) or [`ocr-numbers`](https://github.com/exercism/xpython/blob/d94e9a6ab0ca200decebe4edf4dc502612731108/exercises/ocr-numbers/example.py).

Here is an example for your import problem. If there are a bit more things in your `from X import`: https://github.com/exercism/xpython/blob/master/exercises/pythagorean-triplet/pythagorean_triplet_test.py#L38-L42

cmlang42:
Cleaned up the whitespace issue and the from import statement. Thinking about this more and considering where this exercise is in the order, much further down than hamming, maybe it is worth leaving with custom exceptions as an advanced topic. Hamming already has the ValueError as you mention, so maybe this can introduce custom exceptions.

cmlang42:
Should I move forward with this PR?

behrtam:
Custom errors are over engineering in this case and a `ValueError` with a good message should be the preferred way for this exercise.

```python
>>> on_square(65)
  File "<stdin>", line 1, in <module>
  File "example.py", line 2, in on_square
    check_square_input(square)
  File "example.py", line 20, in check_square_input
    raise SquareGTSixtyFourError
example.SquareGTSixtyFourError
>>>
>>> on_square(65)
  File "<stdin>", line 1, in <module>
  File "example.py", line 2, in on_square
    check_square_input(square)
  File "example.py", line 20, in check_square_input
    raise ValueError('square must be in 1..64')
ValueError: square must be in 1..64
>>>

I also think the `assertRaises` with context manager ([as used in the hamming exercise](https://github.com/exercism/xpython/blob/6fa2bd8a3c857a770bd87cbf208e1fcced5dee18/exercises/hamming/hamming_test.py)) looks better than the old way.

cmlang42:
Removed the custom exceptions and replaced with ValueError exceptions.

cmlang42:
Thanks!  Curious, how did you collapse my many commits down into one?

behrtam:
The git term for collapse is squash. You can do this manually with `$ git rebase -i HEAD~4` (last four commits into one) – read more about that [here](http://blog.benjamin-encz.de/post/squashing-commits-with-git-rebase-guide/) – but GitHub also offers a simple button in the web interface to squash all commits of a PR and merge this as a single commit.

cmlang42:
Thanks @behrtam !   I knew about the squash already but suspected that you probably did not get my branch and do the squashing on your end, so thought there might be some magic feature I didn't know of. Good to hear of this button in GH.  And thanks for the opportunity to contribute.

behrtam:
There is a blog post about that: https://github.com/blog/2141-squash-your-commits

PR #457, opened 2017-03-26
isogram: Add test for duplicate letter in the middle of a word.

hop:
I found at least one solution that always breaks out of the loop after testing the very first character of the input.

    for char in input:
        return (False if input.count(char) > 1 else True)


hop:
I'm sorry, I don't have the time to work through all of the contribution guidelines right now.

Is the Python track test suite generated from the common data?

hop:
I'm in a hurry, so I hope I've done at least some good with this: exercism/x-common#739

Thanks for your patience.

Issue #440, opened 2017-03-22
Parameter order in assertEqual() statements


behrtam:
We have both ways of ordering the parameters actual, expected result of the `assertEqual()` method. 

While the Python implementation uses indifferent names (first, second - [code](https://github.com/python/cpython/blob/master/Lib/unittest/case.py#L816)) all examples use the order actual followed by expected ([python doc – basic-example](https://docs.python.org/3/library/unittest.html#basic-example)).
```python
self.assertEqual('foo'.upper(), 'FOO')
self.assertEqual(s.split(), ['hello', 'world'])
```
I think it would be better to be consistent within this track. As I personally don't have any preferences I would just follow the official python documentation examples and default to `self.assertEqual(actual, expected)`.

@rootulp @pheanex do you have any opinions about this?

pheanex:
(actual, expected) looks good to me. I am in favour of consistency.

rootulp:
I think conforming to the Python doc basic example is a god idea. (Actual, expected) seems right.

nikclayton:
FYI -- if you run the tests in PyCharm (as I'm doing with my students) PyCharm shows the Expected/Actual results in the wrong order. The JetBrains bug for this behaviour is https://youtrack.jetbrains.com/issue/PY-27267.

cmccandless:
@nikclayton thank you for adding that info here.

We've run into that issue before here, but it's good to document it as a known issue.

PR #459, opened 2017-03-30
 transpose : Modified example.py to avoid exception

nithin-vijayan:
Added example.py and exercise test files. Updated config.json. Please review

### DIFF
--- a/exercises/transpose/example.py
+++ b/exercises/transpose/example.py
@@ -4,7 +4,3 @@ def transpose(input_lines):
                  [line.ljust(len(max(lines, key=len)))
                   for line in lines])
     return "\n".join("".join(line) for line in zip(*zipped)).strip()
-
-
-if __name__ == '__main__':
-    transpose()
--- a/exercises/transpose/example.py
+++ b/exercises/transpose/example.py
@@ -1,0 +1,9 @@
+
+def transpose(input_lines):
+    lines = input_lines.split("\n")
+    zipped = map(list, [line.ljust(len(max(lines, key=len))) for line in lines])
+    return "\n".join("".join(line) for line in zip(*zipped)).strip()
+
+
+if __name__ == '__main__':
+    transpose()
--- a/exercises/transpose/example.py
+++ b/exercises/transpose/example.py
@@ -1,7 +1,8 @@
-
 def transpose(input_lines):
     lines = input_lines.split("\n")
-    zipped = map(list, [line.ljust(len(max(lines, key=len))) for line in lines])
+    zipped = map(list,
+                 [line.ljust(len(max(lines, key=len)))
+                  for line in lines])
     return "\n".join("".join(line) for line in zip(*zipped)).strip()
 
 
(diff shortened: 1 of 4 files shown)


exercises/transpose/transpose_test.py
```diff
@@ -0,0 +1,275 @@
+import unittest
+from transpose import transpose
+
+
+class TransposeTests(unittest.TestCase):
+    def test_empty_string(self):
+        input_line = ""
+        expected = ""
+        self.assertEqual(
+            expected,
```
behrtam: The parameters in all `assertEqual()` calls need to be switched. Details and the background can be found here: #440

exercises/transpose/transpose_test.py
```diff
@@ -0,0 +1,275 @@
+import unittest
+from transpose import transpose
+
+
```
behrtam: Could you add the test data version (`# test cases adapted from `x-common//canonical-data.json` @ version: 1.0.0`) here, like here: https://github.com/exercism/xpython/blob/master/exercises/hello-world/hello_world_test.py#L6

exercises/transpose/transpose_test.py
```diff
@@ -0,0 +1,275 @@
...
+            "       f, r   ",
+            "       e  e  m",
+            "       .  m  e",
+            "          o  n",
+            "          v  d",
+            "          e  .",
+            "          ,"
+        ]
+        self.assertEqual(
+            '\n'.join(expected),
```
behrtam: Most of the times you use double quotes (`"\n".joi..`). This should be uniform. There are more lines than this.

exercises/transpose/transpose_test.py
```diff
@@ -0,0 +1,275 @@
...
+        input_line = ""
+        expected = ""
+        self.assertEqual(
+            expected,
+            transpose(input_line)
+        )
+
+    def test_two_characters_in_a_row(self):
+        input_line = ["A1"]
+        expected = [
```
behrtam: I think it would be more readable if this is in one line `expected = ["A", "1"]`
... or when the input is that small maybe even without the variables.
```python
self.assertEqual("\n".join(["A", "1"]), transpose("A1"))
```

nithin-vijayan:
sure, will rectify those issues soon. thanks

nithin-vijayan:
I have modified the code to follow design guidelines. Please review and report back in case of issues.

exercises/transpose/example.py
```diff
@@ -0,0 +1,10 @@
+def transpose(input_lines):
+    lines = input_lines.split("\n")
+    zipped = map(list,
+                 [line.ljust(len(max(lines, key=len)))
+                  for line in lines])
+    return "\n".join("".join(line) for line in zip(*zipped)).strip()
+
+
+if __name__ == '__main__':
```
behrtam: Sorry I missed this the last time, but we don't need that here and the call would even throw an exception.

behrtam: Sorry I missed this the last time, but we don't need that here and the call would even throw an exception.

nithin-vijayan:
Done. Sorry about that. Making a lot of mistakes as it is my first contribution :smile:

PR #460, opened 2017-03-30
Rotational Cipher: Exercise Added

krapes:
I will port the Rotational Cipher excerise

### DIFF
--- a/exercises/rotationa-cipher
+++ b/exercises/rotationa-cipher
@@ -1,0 +1,2 @@
+def rotational_cipher():
+    pass
--- a/exercises/rotationa-cipher
+++ b/exercises/rotationa-cipher
@@ -1,2 +1,0 @@
-def rotational_cipher():
-    pass
--- a/exercises/rotational-cipher/example.py
+++ b/exercises/rotational-cipher/example.py
@@ -1,0 +1,14 @@
+def rotational_cipher(message,key):
+    import string
+    alpha_lower = list(string.ascii_lowercase)
+    alpha_upper = list(string.ascii_uppercase)
+    coded_message = ""
+    for char in list(message):
+        if char in alpha_lower:
+            char = alpha_lower[(alpha_lower.index(char) + key)%26]
+        elif char in alpha_upper:
+            char = alpha_upper[(alpha_upper.index(char) + key)%26]
+        coded_message = coded_message + char
+    return coded_message
+
+
--- a/exercises/rotational-cipher/example.py
+++ b/exercises/rotational-cipher/example.py
@@ -1,12 +1,12 @@
-def rotational_cipher(message, key):
-    import string
-    alpha_lower = list(string.ascii_lowercase)
-    alpha_upper = list(string.ascii_uppercase)
+def rotate(message, key):
+    from string import ascii_lowercase as alpha_lower
+    from string import ascii_uppercase as alpha_upper
+    ALPHA_LEN = len(alpha_lower)
     coded_message = ""
-    for char in list(message):
+    for char in message:
         if char in alpha_lower:
-            char = alpha_lower[(alpha_lower.index(char) + key) % 26]
+            char = alpha_lower[(alpha_lower.index(char) + key) % ALPHA_LEN]
         elif char in alpha_upper:
-            char = alpha_upper[(alpha_upper.index(char) + key) % 26]
-        coded_message = coded_message + char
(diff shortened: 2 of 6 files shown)


krapes:
@behrtam This is ready for a review when you have time.

exercises/rotational-cipher/rotational_cipher.py
```diff
@@ -0,0 +1,14 @@
+
```
behrtam: There is already a description in the markdown file, so I wouldn't add any doc string here. We tend to keep the stub files very limited, so we don't force any specific way of implementing an exercise or naming parameters onto the users. This should be enough:
```python
def rotate():
    pass
```

exercises/rotational-cipher/example.py
```diff
@@ -0,0 +1,12 @@
+def rotational_cipher(message, key):
+    import string
+    alpha_lower = list(string.ascii_lowercase)
+    alpha_upper = list(string.ascii_uppercase)
```
behrtam: Instead of creating the alpha variables, you could rename them inside the import statement like this:
```python
from string import ascii_lowercase as alpha_lower
from string import ascii_uppercase as alpha_upper
``

File: exercises/rotational-cipher/example.py [hunk 3]
```diff
@@ -0,0 +1,12 @@
+def rotational_cipher(message, key):
```
behrtam: The convention is to use nouns for classes and verbs for functions. So in this case I would prefer `rotate()`.

exercises/rotational-cipher/example.py
```diff
@@ -0,0 +1,12 @@
...
+    import string
+    alpha_lower = list(string.ascii_lowercase)
+    alpha_upper = list(string.ascii_uppercase)
+    coded_message = ""
+    for char in list(message):
```
behrtam: The python strings already support the sequence type methods, so you don't need to convert them to lists to use `for` or `index()`.

exercises/rotational-cipher/example.py
```diff
@@ -0,0 +1,12 @@
...
+        if char in alpha_lower:
+            char = alpha_lower[(alpha_lower.index(char) + key) % 26]
```
behrtam: You could create a constant for the length of alphabet `ALPHA_LEN = len(alpha_lower)`. In this case it would also work if the ascii alphabet is changed, which is totally unlikely to happen. So, using 26 is fine as well.

exercises/rotational-cipher/example.py
```diff
@@ -0,0 +1,12 @@
...
+        elif char in alpha_upper:
+            char = alpha_upper[(alpha_upper.index(char) + key) % 26]
+        coded_message = coded_message + char
```
behrtam: This could be shortened with `coded_message += char`.

exercises/rotationa-cipher
```diff
@@ -0,0 +1,2 @@
+def rotational_cipher():
```
behrtam: I assume this file was exceedingly created (at least it's not needed) and should be removed.

exercises/rotational-cipher/rotational_cipher_test.py
```diff
@@ -0,0 +1,55 @@
+import unittest
+
+import rotational_cipher
+
+
+class RotationalCipher(unittest.TestCase):
```
behrtam: We try to stick to the [canonical test data](https://github.com/exercism/x-common/blob/master/exercises/rotational-cipher/canonical-data.json) provided in the `x-common` repository.

If you are missing a test case that would make sense for all tracks, than it would be best to also add it there.
If you think we need an additional test case that would be python specific it would be good to mark those with a comment (`# additional track specific tests`) to make it easy to compare it against the common test cases.

krapes:
@behrtam the updates have been applied.

behrtam:
I can see the finish line. There are only two things left to do.

1. There is still the one file that needs to be removed.

```
rotationa-cipher  <-- this file
rotational-cipher
├── example.py
├── rotational_cipher.py
└── rotational_cipher_test.py
```

2. While PEP8 only advises to put module level imports at the top of the file and there is only one function, I still would put the two import statements and the constant outside the function block and at the top of the file.

krapes:
Haha, my track career was pretty short in High School, I gave it up for something less painful.

The file has been deleted and the import/constant statements moved. Fingers crossed.

behrtam:
Great, thanks for your work!

PR #461, opened 2017-04-03
sublist: Add test cases for edge cases

csabella:
Two cases:
1.  One list has a double digit number and the second list has those numbers as single digits.  For example, 10 vs 1 and 0.
2.  Embedded spaces within a list entry vs individual letters.  For example, 'a c' vs 'a' 'c'.

One way to think about solving this in Python is to use string join, but that gives incorrect results for these edge cases.

### DIFF
--- a/exercises/sublist/sublist_test.py
+++ b/exercises/sublist/sublist_test.py
@@ -76,6 +76,16 @@ class SublistTest(unittest.TestCase):
         multiples_of_15 = list(range(15, 200, 15))
         self.assertEqual(check_lists(multiples_of_15, multiples_of_3), UNEQUAL)
 
+    def test_double_digits(self):
+        l1 = [10]
+        l2 = [1, 0]
+        self.assertEqual(check_lists(l1, l2), UNEQUAL)
+
+    def test_inner_spaces(self):
+        l1 = ['a c']
+        l2 = ['a', 'c']
+        self.assertEqual(check_lists(l1, l2), UNEQUAL)
+
     def test_avoid_sets(self):
         self.assertEqual(check_lists([1, 3], [1, 2, 3]), UNEQUAL)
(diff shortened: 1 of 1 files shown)


behrtam:
But if they use strings it doesn't really matter if there are digits or characters inside, so why do we need both test cases?

Also I don't understand the whitespace in the 2nd case if it's about using the join method of string.

```python
>>> ''.join(['a c']) == ''.join(['a', 'c'])
False
>>>
```

Maybe we could change the digits test case to look like the one in the `x-common` test data: https://github.com/exercism/x-common/blob/master/exercises/sublist/canonical-data.json#L130-L136

csabella:
If you look at the submissions made, most people do the ''.join method, but some people use ' '.join (to put a blank between the strings) to pass the ['10'], ['1', '0'] scenario.  I know that it would also be possible to do '~'.join or something else that wasn't expected in the list and that the tests wouldn't account for all of those.  I don't think the readme specified that the lists would only be numbers, so that's one reason I included the one with strings and an embedded blank. 

I'll change the digits case to match x-common.

behrtam:
Thanks for the explanation.

csabella:
You're welcome and thanks for asking about it.  I'm not trying to be annoying by adding more tests.  I'm using this platform to learn to solve the problems, but also to learn to build good test cases.  I just had created these tests which seemed to be gotchas if you solved it one way.

behrtam:
Now we only have to make it easier for people to discover new tests and update the test files: https://github.com/exercism/cli/issues/362

csabella:
That's interesting.  I hadn't even considered going back and looking for new tests on the exercises I already completed.  Would it be possible to send a notification if a new test is added to an exercise that has already been submitted?

behrtam:
I would like that. But I'm not sure where and how to add that to exercism.

PR #467, opened 2017-04-17
grade-school: catch false positives due to test case being sorted already

hop:
From Python 3.6 on, dictionaries retain the order in which items are
added. If the test case puts the items into the tested class in sorted
order, even a naive implementation (one that does no sorting) can pass
by giving back all items in the original order.

### DIFF
--- a/exercises/grade-school/grade_school_test.py
+++ b/exercises/grade-school/grade_school_test.py
@@ -46,7 +46,7 @@ class SchoolTest(unittest.TestCase):
                     (6, ("Kareem", ))]
 
         for grade, students_in_grade in students[::-1]:
-            for student in students_in_grade:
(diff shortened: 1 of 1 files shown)


behrtam:
> Get a sorted list of all students in all grades.  Grades should sort as 1, 2, 3, etc.,
and students within a grade should be sorted alphabetically by name.

```python
sorted_students = [(3, ("Kyle", )), ...
unsorted_students = 
[(4, ("Jennifer", "Christopher", )), (3, ("Kyle", )), (6, ("Kareem", ))]
```

Good catch! As both levels need to be sorted, maybe we should also switch `Jennifer` and `Christopher`.

hop:
Here you go!

I have no idea why that didn't occur to me in the first place.

PR #473, opened 2017-05-29
dibs: implement protein-translation

cptjackson:
Hi! This is my first time ever making a pull request. Thought I would start learning to contribute to projects, and porting some code looks like a fun place to start. I'm not very experienced with github.

I'm trying to follow the instructions for contributing code here:

https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#porting-an-exercise-to-another-language-track

Please let me know if I'm doing it right, and if not, what I need to change. Thanks :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -488,6 +488,14 @@
       "topics": [
       ]
     }
+    {
+      "slug": "protein-translation",
+      "difficulty": 3,
+      "topics": [
+        "strings",
+        "pattern matching"
+      ]
+    }
   ],
   "deprecated": [
     "accumulate",
--- a/config.json
+++ b/config.json
@@ -487,7 +487,7 @@
       "difficulty": 1,
       "topics": [
       ]
-    }
+    },
     {
       "slug": "protein-translation",
       "difficulty": 3,
--- a/exercises/protein-translation/example.py
+++ b/exercises/protein-translation/example.py
@@ -1,0 +1,39 @@
+#!/usr/bin/env python
+class InvalidCodonError(Exception):
+    def __init__(self,codon):
+        self.codon = codon
+        Exception.__init__(self, 'Invalid Codon exception: %s is not a valid \
+                           codon' % codon)
+
+class Translation(object):
+
+    # Lookup table
+    codonList = {'AUG': "Methionine", 'UUU': "Phenylalanine",
+                'UUC': "Phenylalanine", 'UUA': "Leucine", 'UUG': "Leucine",
+                'UCU': "Serine", 'UCC': "Serine", 'UCA': "Serine",
+                'UCG': "Serine", 'UAU': "Tyrosine", 'UAC': "Tyrosine",
+                'UGU': "Cysteine", 'UGC': "Cysteine", 'UGG': "Tryptophan",
+                'UAA': "STOP", 'UAG': "STOP", 'UGA': "STOP"}
+
+    def __init__(self):
+        self.codons = self.codonList
+
+    def ofCodon(self,codon):
+        try:
+            return self.codons[codon]
+        except KeyError:
+            raise(InvalidCodonError(codon))
+
+
+    def ofRNA(self,strand):
+        output = []
+        for i in range(0,len(strand),3):
+            codon = strand[i:i+3]
+            try:
+                if self.codons[codon] == 'STOP':
+                    return output
+                output.append(self.codons[codon])
+            except KeyError:
+                raise(InvalidCodonError(codon))
+
+        return output
--- a/exercises/protein-translation/example.py
+++ b/exercises/protein-translation/example.py
@@ -1,7 +1,4 @@
-#!/usr/bin/env python
-
-# Lookup table
-codons = {'AUG': "Methionine", 'UUU': "Phenylalanine",
+CODONS = {'AUG': "Methionine", 'UUU': "Phenylalanine",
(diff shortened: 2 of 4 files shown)


behrtam:
👍 If you have any questions don't hesitate to ask or commit non working code so we can have a look and help you.

cptjackson:
Yes I've been looking at the ruby tests and implementing those.

cptjackson:
So I'm running into a problem: I can't get the tests to work. This is my first time doing unit testing as well. I understand the theory but I keep getting errors. This isn't just with my code either; it's the code in the other python tests.

For example, from the `xpython/exercises/linked-list/` directory:

`python linked_list_test.py`

or

`py.test linked_list_test.py`

gives me huge AttributeErrors for everything, e.g.:

```
_____________________________ LinkedListTests.test_all _____________________________

self = <linked_list_test.LinkedListTests testMethod=test_all>

    def test_all(self):
>       self.list.push(10)
E       AttributeError: 'LinkedList' object has no attribute 'push'

linked_list_test.py:35: AttributeError
```

This is the same thing that happens with my own tests. Can you shed any light?

behrtam:
Sure, that happens because you are testing the stub file and not the example implementation.
You could either rename `example.py` or use the [check-exercises.py](https://github.com/exercism/xpython/blob/master/test/check-exercises.py) script that automatically (with a tmp file) does it for you. It's also mentioned in the `README.md`.

```
$ cat ./exercises/linked-list/linked_list.py
class Node(object):
    def __init__(self):
        pass


class LinkedList(object):
    def __init__(self):
        pass

$ python ./exercises/linked-list/linked_list_test.py -f
E
ERROR: test_all (__main__.LinkedListTests)
  File "./exercises/linked-list/linked_list_test.py", line 35, in test_all
    self.list.push(10)
AttributeError: 'LinkedList' object has no attribute 'push'

Ran 1 test in 0.000s

FAILED (errors=1)

$ python ./test/check-exercises.py linked-list
#  linked-list
.ss....
Ran 7 tests in 0.000s

OK (skipped=2)

TestEnvironment: Python2.7


SUCCESS!
```

cptjackson:
Ok well that worked perfectly. Thank you! I guess the lesson is to always read the `README.md`. And now I have some working solutions, I'll push them up shortly.

cptjackson:
I think that's it. Do I need to do anything else?

behrtam:
As you can see here the travis-ci build looks good: https://travis-ci.org/exercism/xpython/builds/237577149

That basically means that your code/tests work on all the different Python versions (2.7, 3.4-6) and that they meet the style requirements (flake8). Nicely done!

Now, I will try to find some time to do a review. Maybe today, but I can't promise you anything.

cptjackson:
Oh I'm not in a rush. Maybe I'll start another one.

behrtam:
You can find unimplemented exercises here: http://exercism.io/languages/python/todo

behrtam:
The tests look good to me. I personally favor the context manger style of testing exceptions because it let's you make the call yourself. But this is just a question of taste, as both ways do the same thing.

```python
def test_invalid_codons(self):
  strand = 'CARROT'
  self.assertRaises(InvalidCodonError, self.translate.ofRNA, strand)

def test_invalid_codons(self):
  with self.assertRaises(InvalidCodonError):
    self.translate.ofRNA('CARROT')
```

My feeling is that the `class` based solution doesn't really add any value here and a functional approach would be clearer.
Some examples where exercises have been moved from a class to functions: [pull/85/files](https://github.com/exercism/xpython/pull/85/files), [pull/93/files](https://github.com/exercism/xpython/pull/93/files)

In general in Python snake case `of_codon(param)` is preferred over camel case `ofCordon(param)` for variables, methods and functions. Class names instead use camel case.
Maybe have a look at: http://pep8.org

cptjackson:
Thanks for the feedback! I used the class-based solution because that's what the Ruby code I based it on does, but it's easily switchable. Also I think the exception is `KeyError` rather than `ValueError`. Will revise and update.

behrtam:
That looks much better. Only, some small details and than we should be able to merge this into master.

**A) shebang**
I don't really know why, but the only Python file with `#!/usr/bin/env python` is `test/check-exercises.py`. So I would remove that line to be consistent.

Maybe we should open an issue to investigate why we don't have it and/or if maybe it should be added.

**B) codons dict**
The name of a constant like `codons` should be upper case and the comment `# Lookup table` is redundant because thats basically what a Python dictionary is.

**C) exception**
I think the `ValueError` is better than `KeyError` as it's more flexible because we don't force the user to use a dictionary to implement this, even though I think it's the best option.

**D) extract the chunk logic**
So, this is up for debate. Do you think it helps to understand this, if we extract the logic to cut the string into chunks into its own function and by this are able to give it an explanatory name?

```python
def of_codon(codon):
    if codon not in CODONS:
        raise ValueError('Invalid codon: %s' % codon)
    return CODONS[codon]


def of_rna(strand):
    proteins = []
    for codon in map(of_codon, _chunkstring(strand, 3)):
        if codon == "STOP":
            break
        proteins.append(codon)
    return proteins


def _chunkstring(string, n):
    return (string[i:n + i] for i in range(0, len(string), n))
```

cptjackson:
Fixed **A)** and **B)**, and **C)**.

As for **D)**, I guess it could go either way. I think I like your logic though. It does make it a bit more straightforward.

behrtam:
Thank you for your work. It will be live with the next update.

PR #497, opened 2017-07-22
Update the enigmatic scale case

nikamirrr:
http://exercism.io/submissions/3c67c65da6e34871946a955fca7ff2ae
There seem to be a mistake in the enigmatic scale: enigmatic = Scale('G', 'enigmatic', 'mAMMMmM')

https://en.wikipedia.org/wiki/Enigmatic_scale Semitone, Tone and a half, Tone, Tone, Tone, Semitone, Semitone. Should be enigmatic = Scale('G', 'enigmatic', 'mAMMMmm')

### DIFF
--- a/exercises/scale-generator/scale_generator_test.py
+++ b/exercises/scale-generator/scale_generator_test.py
@@ -110,7 +110,7 @@ class ScaleGeneratorTest(unittest.TestCase):
         self.assertEqual(expected, actual)
 
     def test_enigmatic(self):
-        enigmatic = Scale('G', 'enigmatic', 'mAMMMmM')
+        enigmatic = Scale('G', 'enigmatic', 'mAMMMmm')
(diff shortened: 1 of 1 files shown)


behrtam:
Thanks! We should probably change the test, so this is actually validated and your change does really make a difference.
If we check `G G# B C# D# F F# G` instead of only `G G# B C# D# F F#` than also the last pattern is validated.

nikamirrr:
Agreed, but you should then change ALL tests to repeat the first tune at the end.
This way you will have N tunes and N-1 intervals
Or add one more test that will confirm that the user makes sure the last step is valid against the last and the first tune.

I think:
1) Fix the enigmatic test case like I proposed
2) re-add this broken enigmatic test as the newand expect a user to raise ValueError

PR #498, opened 2017-07-23
Update scale_generator_test.py

nikamirrr:
re-added the broken enigmatic case to actually make sure the developer flags it with the exception


behrtam:
You would at least need to also update the example implantation to make this test pass.

But this would mean to put a lot of information into that class so the code knows what a pentatonic or enigmatic scale is. Do you think this is a good way to go?

nikamirrr:
I don't think it is necessary. Just need to tweak the sample solution to test for the last tune.
I will see if I can tweak it

nikamirrr:
close

PR #499, opened 2017-07-23
scale-generator: a new test added

nikamirrr:
Done

### DIFF
--- a/exercises/scale-generator/example.py
+++ b/exercises/scale-generator/example.py
@@ -26,6 +26,8 @@ class Scale(object):
         for i, interval in enumerate(self.pattern):
             pitches.append(scale[last_index])
             last_index += self.ASCENDING_INTERVALS.index(interval) + 1
+        if pitches[0] != scale[last_index % len(scale)]:
+            raise ValueError()
         return pitches
 
     def _reorder_chromatic_scale(self):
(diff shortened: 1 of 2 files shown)


nikamirrr:
done

m-a-ge:
@nikamirrr, is this PR related to any issue or test case in question is in `canonical-data.json` for the exercise and that's the reason you've added it?

nikamirrr:
If the test is auto-generated, then likely the cause is there. Not sure how to check.

m-a-ge:
Ok, I've checked `problem-specification` repo for this exercise and `canonical-data.json` is missing there - https://github.com/exercism/problem-specifications/tree/master/exercises/scale-generator
I am not familiar with that exercise to be honest, should get more context.

@nikamirrr, why adding this test case is important?

nikamirrr:
Because the generated scales need to be in the closed loop.
Meaning the last step needs to wrap the scale from the last note to the first.
In the current test codes the last step is simply ignored.
Alternatively one may change the exercise and require that an extra note is generated at the end and matches the first note.

m-a-ge:
@nikamirrr, thanks for the explanation, I'm starting understanding the issue.
Give me some time to check a few things before proceeding with the merge

m-a-ge:
@nikamirrr, thank you for working on this :hammer_and_wrench:

PR #506, opened 2017-08-06
Update "TESTS.md" to Show That There Is a Version of "pytest" for Both Python and Python3

TheUltraCode:
Provides information on the different types of "pytest" available and how to install each, as well as switched "py.test" to "pytest", as recommended here (https://github.com/pytest-dev/pytest/issues/1629#issue-161422224).


TheUltraCode:
Although what I have written is most likely far from clean or professional, if the difference between "pytest" packages from pip and pip3 is explained somehow by someone and that explanation is interwoven into this file, then I would consider this attempted addition to the documentation to be a success. The other changes I made concerning "how to run 'pytest' (`pytest` instead of `py.test`)" are not really that important, but I believe using the new way to run "pytest" might be better in the long run.

m-a-ge:
@TheUltraCode, so there is no difference between `py.text` and `pytest`?

docs/TESTS.md
```diff
@@ -107,10 +115,16 @@ compliance to the style guide, install
 pip install pytest-pep8
 ```
 
+or, for Python 3.X syntax compliance...
```
behrtam: I don't understand this new part.

TheUltraCode:
@m-a-ge Yes, there is no functional difference between `py.test` and `pytest`. It is just as of June 16, 2016, that the developers of "pytest" now urge the use of the new `pytest` command instead of the older `py.test` command ( https://github.com/pytest-dev/pytest/issues/1629#issue-161422224 ). Besides, one of the benefits of `pytest` over `py.test` is that you do not have to type that pesky ".".

TheUltraCode:
@behrtam I have you to thank for finding a typo (meant "pip3" instead of "pip" after "or, for Python 3.X syntax compliance..."). Thanks!

Now then, what do you do not understand? If you are referring to "Python 3.X syntax compliance", I can help you.

Okay, let's start off with the fact that there are differences between Python 2.X and 3.X. For example, in Python 3.X, the `print` statement must be used more like a regular function:

    # Python 2.X "print" Statement
    print "Hello World"

    # Python 3.X "print" Statement
    print("Hello World")

Now, this is one of the simplest differences between the different versions of Python. There are much more, most of which are minute, but this should act as an example, one which I will use later below.

Okay, now that we know that "syntax" can differ between Python versions, generations, branches, what-have-you, let's now look at "pip", the Python package manager. There are currently two versions of "pip", "pip" (which installs Python packages and/or programs that employ and follow Python 2.X syntax), and "pip3" (which does the same thing, but for Python 3.X syntax).

Okay, we are a few steps hopefully from solving your misunderstanding. Let's say for one of "exercism's" Python exercises you write code that relies heavily on Python 3.X, using features not in 2.X. Well, if you installed the "pip" version of "pytest" and tried to run `pytest`, you would most likely get errors, because that version of "pytest" is meant to understand only Python 2.X code. However, if you were to install the "pip3" version of "pytest", there would most likely (assuming you made like no mistakes) *zero* errors produced from running "pytest", because this version of pytest is meant to understand Python 3.X code.

Using the "print" statement mentioned above, if your goal in a Python exercise from "exercism" was to write using 3.X syntax, and you typed `print "Hello World"` for example and you used the "pip" version of "pytest" to test your work, it would come back jim-dandy instead of with a syntax error because this "pytest" version only knows Python 2.X. Therefore, you are doing yourself a disfavor and potentially holding yourself back in terms of understanding the Python version you are trying to learn!

I'm sorry if *none* of this makes sense. I'm not good at explaining things, but I hope, you do indeed do not understand, someone catches my drift and then explains it better than me. Hope this helps!

TheUltraCode:
Not trying to sound mean. Just curious. Where are we with this Pull Request? Is it a strong, potential Go or a No Go? It's been a while.

m-a-ge:
@TheUltraCode sorry for the delay and thanks for your patience!

I like your proposal, `py.test` might be deprecated in future and it's better to use `pytest`.
Could you please clarify the following addition - https://github.com/exercism/python/pull/506/files#r134086795?

TheUltraCode:
@m-a-ge  Oh, I already addressed that above in this "Conversation", where it starts with "@behrtam I have you to thank for finding a typo (meant "pip3" instead of "pip" after "or, for Python 3.X syntax compliance..."). Thanks!". If you would like me to repost that entire post into the "Reply" field, I can do that as well. If you would like me to clarify anything, I can do that as well. :)

To say the least, though, it's just like the difference between `pip3 install pytest pytest-cache` and `pip install pytest pytest-cache`: the former installs the package `pytest`, which adheres to Python 3.X syntax,  while the latter adheres to Python 2.7.X syntax. Reapplying that concept, `pip3 install pytest-pep8` installs the package `pytest-pep8` which also adheres to Python 3.X syntax, and the opposite is true for `pip install pytest-pep8`. While these pairs of packages are extremely similar, they only differ in what version of Python they are meant to work with. You could send Python 3.X code, for example, through to the Python 2.7.X `pytest`, and while there might be a good chance of that Python 3.X code being run "correctly", you are just as, if not more likely, to end up with errors spat out by `pytest`, depending how much of your Python 3.X code "deviants" from Python 2.7.X syntax.

docs/TESTS.md
```diff
@@ -4,9 +4,17 @@ We recommend you install [pytest](http://pytest.org/latest/) and
...
+
 ```bash
 pip install pytest pytest-cache
 ```
+If you instead want the vesrion of pytest that obeys Python 3.X syntax, then use `pip3`:
+
+```bash
+pip3 install pytest pytest-cache
+```
+**Note:** Whichever version of pytest you install first will be the default one used whenever `pytest` is executed, regardless of whether you have installed both versions.
```
N-Parsons: It might be worth mentioning that you can check the default version by running `pytest --version` (on Linux and presumably OSX?, not sure what the equivalent is in Windows).

On my system it returns:
```shell
$ pytest --version
This is pytest version 3.2.1, imported from /usr/lib/python3.6/site-packages/pytest.py
```

docs/TESTS.md
```diff
@@ -4,9 +4,17 @@ We recommend you install [pytest](http://pytest.org/latest/) and
 [pytest-cache](http://pythonhosted.org/pytest-cache/). `pytest` is a testing
 tool that will give you more flexibility over running your unit tests.
 
+If you want to install the version of pytest that obeys Python 2.X syntax, then use `pip`:
+
 ```bash
 pip install pytest pytest-cache
 ```
+If you instead want the vesrion of pytest that obeys Python 3.X syntax, then use `pip3`:
```
N-Parsons: Looks like you have a typo here: "vesrion" -> "version"

docs/TESTS.md
```diff
@@ -4,9 +4,17 @@ We recommend you install [pytest](http://pytest.org/latest/) and
 [pytest-cache](http://pythonhosted.org/pytest-cache/). `pytest` is a testing
 tool that will give you more flexibility over running your unit tests.
 
+If you want to install the version of pytest that obeys Python 2.X syntax, then use `pip`:
```
N-Parsons: This might be better worded as: "If you want to install `pytest` for Python 2, then use `pip`:"?
Similar wording can be used for Python 3.

docs/TESTS.md
N-Parsons: Perhaps this might be better worded as "or, if you're using Python 3:"?

TheUltraCode:
Thanks for the feedback @N-Parsons! I took your suggestions and applied them to "TESTS.md" the best I could. I also cleaned up the file a teensy bit and added some new information to it which I deemed (and hope to be) helpful.

I would be happy to know your thoughts on my changes!

N-Parsons:
Thanks for your work on this, @TheUltraCode!

TheUltraCode:
You're welcome, everyone!
And thanks for giving me the chance to complete my first pull request! It was a pleasure!

Issue #509, opened 2017-08-11
Function in templates should have parameters?


JanKaifer:
I feel better if I have the parameters in template (maybe because I'm used to it from other sites). When I've been solving first problem with input I didn't knew how I'm supposed to get input. I had to look in the tester script.

behrtam:
But shouldn't choosing a good name for parameters be part of solving the exercises?

JanKaifer:
You can always rename it 😄

m-a-ge:
Fixed in https://github.com/exercism/python/pull/510

JanKaifer:
@m-a-ge well its not fixed yet because I fixed just one exercise. But others templates ar still without parameters. If you find this as a good idea then all exercises should be modified.
I don't know where to write it to let others know that new exercises should have parameters in template. And to tell contributors that someone could upgrade other exercises as well.

PS. Thanks for patience and sorry for any mismatches or errors. This is my first contribution and I don't know how these hings exactly work.

m-a-ge:
@JanKaifer, got it now
I'll reopen the issue then

JanKaifer:
Thx

JanKaifer:
Hi, how can I add label to this issue? I'm contributor so I should have the rights. But I don't have theses checkboxes for adding labels. I'd love to mark this as beginner friendly or even first-timers-only.
Thx

m-a-ge:
@JanKaifer, I've added requested label.

lilislilit:
#521 Mostly covers it, I hope :)

JanKaifer:
Thanks for changing all exercises.
I'm wondering If this issue should be closed (because it's changed everywhere) or left open (as info for new exercises).

JanKaifer:
@m-a-ge could you please un-merge my PR #514 because as @behrtam says in #521 there should be just one exercise per PR as in #510.

(Maybe we could make on exception with this rule because this will create lot of PRs. But @m-a-ge @behrtam  are the leaders here :smile: .)

m-a-ge:
First of all, @lilislilit and @JanKaifer thanks for working on this!
To be honest I don't want to revert #514, we can always create a new PR to improve params naming if required
And I also agree that creating PRs for each exercise can be time consuming and unnecessary

lilislilit:
@m-a-ge I agree, I don't have that much of a preference, but it would be like...a few dozen PR's. We should come to agreement with @behrtam then.

lilislilit:
[...]

PR #510, opened 2017-08-11
isogram: added parameter to template question

JanKaifer:
I added parameter string to template question. viz #509


m-a-ge:
Looks reasonable to me.
Thanks, @JanKaifer

m-a-ge:
@JanKaifer, thanks for opening #509 and eventually fixing it 👍

PR #514, opened 2017-08-22
adding parameters to template  A - B

JanKaifer:
viz issue #509

### DIFF
--- a/exercises/accumulate/accumulate.py
+++ b/exercises/accumulate/accumulate.py
@@ -1,2 +1,2 @@
-def accumulate():
+def accumulate(array, function):
     pass
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 tmp
 bin/configlet
 bin/configlet.exe
+.idea/
--- a/exercises/allergies/allergies.py
+++ b/exercises/allergies/allergies.py
@@ -1,3 +1,11 @@
 class Allergies(object):
-    def __init__(self):
+
+    def __init__(self, number):
+        pass
+
+    def is_allergic_to(self, string):
+        pass
+
+    @property
+    def lst(self):
         pass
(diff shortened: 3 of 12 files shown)


exercises/accumulate/accumulate.py
```diff
@@ -1,2 +1,2 @@
-def accumulate():
+def accumulate(array, function):
```
m-a-ge: This looks Ok to me, but some people would argue that you will force people to certain implementations by giving exact params.

I personally think it's a good idea to do so anyway, if one still doesn't want to follow suggested method signature he/she can ignore it and write his/her own implementation.

Moreover, sometimes forcing certain implementation helps people to use features of the language they aren't used to (like passing functions to a method, not all languages allow that etc)

As an improvement, I would name params with meaningful names rather than generic ones. But it's up for you to decide

JanKaifer: > As an improvement, I would name params with meaningful names rather than generic ones. But it's up to you to decide

I wanted to use names from example.py, but as @behrtam said in issue #509 choosing good name is important thing too. So I chose generic ones. But that is just detail which depends on exercise creator.

m-a-ge: @JanKaifer, got it, thanks for clarification

m-a-ge:
@JanKaifer,  I've merged your changes, thanks. 

And I think it's a good idea to give a learner a little clue of what they should by providing expected parameters

behrtam:
We explicitly decided against *parameters/arguments* when we added stups ...

m-a-ge:
@behrtam, we can always revert this changes. 

Why are you against of having the params in example code? I find it somewhat confusing to not have them. And tests still expect that from you, take a look at https://github.com/exercism/python/blob/master/exercises/leap/leap_test.py for example

As I said before, the learner can ignore them if he/she thinks the suggest signature is not correct or convenient

PR #516, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

Note that we're aiming for 10 lines and a max-width of 40 columns.
This solution has a max-width of 41, so we may want to adjust it.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,10 @@
+prices = {'apple': 0.75, 'egg': 0.50}
+cart = {
+  'apple': 1,
+  'egg': 6
+}
+
+bill = sum(prices[item] * cart[item]
+           for item in cart)
+
+print(f'I have to pay {bill:.2f}')


docs/SNIPPET.txt
```diff
@@ -0,0 +1,5 @@
+def hello(name=''):
+    if name:
+        return u'Hello, {}!'.format(name)
```
m-a-ge: I don't like this `u''` part since Strings in Python 3 are Unicode ones - https://docs.python.org/3/howto/unicode.html#python-s-unicode-support
And we should highlight the current version of the language.

Plus this code doesn't reflect OOP portion of Python

kytrinyx: These are all good observations. @m-a-ge would you be willing to find a better example for this?

behrtam:
```python
def hello(name='World'):
    return f'Hello, {name}!'

>>> hello()
'Hello, World!'
>>> hello('Katrina')
'Hello, Katrina!'
>>>
```

If it's about making the `helloworld` more pythonic, this could be a way. But maybe we could do more in the 10x40 space ...

docs/SNIPPET.txt
m-a-ge: @kytrinyx,
I was thinking about something like this:
```python
class Greeter(object):
    name = ''
    def __init__(this, name):
        this.name = name
        
    def greet(this):
        print('Hello, {}!'.format(this.name))
        
name = input('Enter your name: ')
Greeter(name).greet()
```
@behrtam, what do you think?

Side note: I understand that extending `object` is not necessary in Python 3, but I intentionally put it there for demonstration purposes

kytrinyx:
I like the _prices_ example! It shows all sorts of bits and pieces of the language.

cmlang42:
I might be jumping too far on this, and please tell me if I am. If this is in reference to a very first exercism exercise, I do not think classes, functions, or anything beyond a simple print statement is a good first step. I thought we were aiming to have the first exercise be more about setting up the exercism environment and making something in the specific language work?  If so, especially for those completely new to Python, having the solutions include anything beyond print() is taking the user into unfamiliar territory in addition to verifying the setup and learning to use exercism.

behrtam:
I think this `code example` which is used [here](https://v2.exercism.io/tracks/python) is more about demoing the language and not about being the very first exercise.

kytrinyx:
> this [...] is more about demoing the language and not about being the very first exercise.

Yes—sorry about being unclear. This is a quick visual taste/impression. It should give an idea of what the language looks like, but nobody needs to actually understand it.

By default if we don't have a good example, we pull in the solution to the first exercise, but since that tends to be very, very simple, it often doesn't give a good idea of what the language actually looks like.

kytrinyx:
Alright. I've updated this to use prices. @behrtam if it looks good would you merge it?

PR #517, opened 2017-08-30
simple-cipher: 'ArgumentError' -> 'ValueError'

twsh:
The test looks for a `ValueError`.


m-a-ge:
Thanks for bringing this up. 
Judging by test suite https://github.com/exercism/python/blob/master/exercises/simple-cipher/simple_cipher_test.py only `ValueError` is expected which is Ok -
 https://docs.python.org/3/library/exceptions.html#ValueError

I guess in reality something like this will be used:
```python
class IllegalArgumentError(ValueError):
    pass
```

PR #519, opened 2017-09-05
Update word count to match canonical tests

### LINKED ISSUE #445: Rework word-count exercise to be consistent
In 2015 we decided to make the `word-count` exercise more interesting and not only split on white spaces (https://github.com/exercism/xpython/pull/177) which was inspired by the Haskell track. Since than other tracks where against it and so Haskell moved back to white space only (https://github.com/exercism/xhaskell/issues/501). This is also the way the canonical test data looks like. So maybe we should drop the python specialty to be more uniform.

matthewmorgan:
Closes #445 

The regex is not my favorite, but it gets the job done.  Also, the tests now match the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/word-count/canonical-data.json).

I also took the liberty of skipping all but the first test.  If #518 gets merged, these tests will run in CI ;)

### DIFF
--- a/exercises/word-count/example.py
+++ b/exercises/word-count/example.py
@@ -1,8 +1,13 @@
 from collections import Counter
+import re
+
+regex = r"'([\w]+)?'|\s?([\w']+)\s?"
 
 
 def word_count(text):
-    def replace_nonalpha(char):
-        return char.lower() if char.isalnum() else ' '
-    text = ''.join(replace_nonalpha(c) for c in text)
-    return Counter(text.split())
+    words = [match_and_lower(match) for match in re.finditer(regex, text)]
+    return Counter(words)
+
+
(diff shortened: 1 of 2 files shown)


behrtam:
This quote stuff is somehow overcomplicating this otherwise simple exercise. I agree that a regex is the best solution in this case ... otherwise one would have to write a lot just to handle the special cases.

matthewmorgan:
@behrtam yes, the quote thing is thorny.  Solving the exercise was trivial except for that.

m-a-ge:
While implementing this (or rather improving) example solution for this exercise for `bash` track I simply ignored the case with `'` from `canonical-data.json`
So, it isn't necessary to include all the tests, but it's generally a good idea

Thanks for the PR, @matthewmorgan

PR #523, opened 2017-09-07
Implement exercise complex-numbers

lilislilit:
Trying to implement complex-numbers according to guideline. This is initial commit for WIP pull-request. Tests and example.py still needs to be finished.

### DIFF
--- a/config.json
+++ b/config.json
@@ -824,6 +824,17 @@
       ]
     },
     {
+      "uuid": "7f4d5743-7ab8-42ca-b393-767f7e9a4e97",
+      "slug": "complex-numbers",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 6,
+      "topics": [
+        "tuples",
+        "mathematics"
+      ]
+    },
+    {
       "uuid": "e7351e8e-d3ff-4621-b818-cd55cf05bffd",
       "slug": "accumulate",
       "deprecated": true
--- a/exercises/complex-numbers/complex_numbers.py
+++ b/exercises/complex-numbers/complex_numbers.py
@@ -1,0 +1,24 @@
+class ComplexNumber(object):
+    def __init__(self, real, imaginary):
+        pass
+
+    def add(self, other):
+        pass
+
+    def mul(self, other):
+        pass
+
+    def sub(self, other):
+        pass
+
+    def div(self, other):
+        pass
+
+    def abs(self):
+        pass
+
+    def conjugate(self):
+        pass
+
+    def exp(self):
+        pass
--- a/exercises/complex-numbers/README.md
+++ b/exercises/complex-numbers/README.md
@@ -1,0 +1,22 @@
+# Complex Numbers
+
+A complex number is a number in the form `a + b * i` where `a` and `b` are real and `i` satisfies `i^2 = -1`.
+
+Assume the programming language you are using does not have an implementation of complex numbers.
+
+### Submitting Exercises
+
+Note that, when trying to submit an exercise, make sure the solution is in the `exercism/python/<exerciseName>` directory.
+
+For example, if you're submitting `bob.py` for the Bob exercise, the submit command would be something like `exercism submit <path_to_exercism_dir>/python/bob/bob.py`.
+
+
+For more detailed information about running tests, code style and linting,
+please see the [help page](http://exercism.io/languages/python).
+
+## Source
+
+Wikipedia [https://en.wikipedia.org/wiki/Complex_number](https://en.wikipedia.org/wiki/Complex_number)
+
+## Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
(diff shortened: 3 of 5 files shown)


m-a-ge:
Good luck with this one and have fun 👍 @lilislilit

lilislilit:
tests and example.py are both finished, so there is that. Humbly waiting for nitpicking and/or approval.

config.json
```diff
@@ -824,6 +824,17 @@
       ]
     },
     {
+      "uuid": "7f4d5743-7ab8-42ca-b393-767f7e9a4e97",
+      "slug": "complex-numbers",
+      "core": false,
+      "unlocked_by": "leap",
```
m-a-ge: Could you please share with us the reasoning behind this decision?
Don't get me wrong, I am not against it, just curios

And it turned out (while testing a new version of Exercism) that non-core exercise should be unlocked by only a core one.
We should revisit the whole track curriculum, perhaps you can help with a fresh view

config.json
```diff
@@ -824,6 +824,17 @@
...
+      "difficulty": 6,
```
m-a-ge: The same for difficulty.
What affected the score to be 6?

config.json
```diff
@@ -824,6 +824,17 @@
...
+      "topics": [
+        "Tuples",
```
m-a-ge: Good choice.

We are trying to standardize topics across all tracks and decided to use `snake_case` notation as a naming convention. List of of common topics can be found here - https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt 
Could you please convert them to 'snake_case'

exercises/complex-numbers/complex_numbers.py
```diff
@@ -0,0 +1,3 @@
+class ComplexNumber(object):
+    def __init__(self):
```
m-a-ge: As it was discussed in #509 it would be a good idea to include expected parameters to not force the user to go through a test class

What are your thoughts on this?

lilislilit: Yeah, I think it is a wise thing to do

config.json
lilislilit: To be completely honest, I was looking at a c# track, it looks somewhat finished, so I figured that their unlocking graph is finished too. If we will be redoing the whole of dependency graph, maybe it warrants separate issue and separate PR?

config.json
lilislilit: Yeah, same hear. Took it from C# track. As I understood, difficulty grade is kinda... subjective.

exercises/complex-numbers/complex_numbers.py
lilislilit: Although, should I also include method signatures? Or only params for constructor? If we don't want people to go through test suite too much, including method signatures makes sense, some tracks here do this.

config.json
m-a-ge: Yeah, this should be addressed in a separate issue\PR.
I didn't have a chance to create it, I'll create it later today
But still, it's better to leave `unlocked_by` empty for (until we'll figure out a list of `core` exercises)

config.json
m-a-ge: Agreed. Maybe we can come with some sort of formula to compute it, what do you think?

exercises/complex-numbers/complex_numbers.py
m-a-ge: I guess it makes sense to include method signatures since the user uses the provided test suite anyway.
And if someone wants to create their own unique solution, they are free to change provided class placeholder or even test suite (rewrite it, add or remove tests)

I don't see how providing method signatures can influence the user's overall implementation. The most important thing for me is convenience

config.json
lilislilit: I think this question maybe hurried somewhere in general discussions repo. Btw, will difficulty be used at all in version 2, or there would be relying on "unlocked_by" prerequisites?

In general, it could be done, of course, simplest heuristic would be to split topics that we have in to tiers, from simplest to hardest and add the difficulty level for each additional topic for exercise. So if basically tier_level1+....tier_level2

m-a-ge: @lilislilit, hm, this a really good point, thanks! I haven't thought about it this way
I've created the following issue https://github.com/exercism/discussions/issues/193 in general discussion, feel free to add your thoughts there

m-a-ge:
@lilislilit, everything looks good to me, thanks for taking the time and implementing this!

Generally it would be a good idea to [squash commits](https://github.com/exercism/docs/blob/master/contributing/git-basics.md#squashing), but it's up to you

lilislilit:
@m-a-ge From the looks of history, commits look squashed :)

cmccandless:
Test cases `test_multiply_numbers_with_real_and_imaginary_part` is incorrect. Please stick to the [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/complex-numbers/canonical-data.json#L77-L102) when implementing established exercises in a new language.

m-a-ge:
@cmccandless could please create an issue describing the details?

cmccandless:
Done. See #527.

m-a-ge:
@cmccandless thanks!

PR #529, opened 2017-09-23
phone-number: Add validation of area code and exchange code

N-Parsons:
* Update tests to use current version of canonical data (v1.2.0). These tests
  require validation of the area and exchange codes, which is already mentioned
  in README.md - the first digit of each must be between 2 and 9

* Update example code to pass new tests

### DIFF
--- a/exercises/phone-number/example.py
+++ b/exercises/phone-number/example.py
@@ -27,8 +27,10 @@ class Phone(object):
         )
 
     def _normalize(self, number):
-        valid = len(number) == 10 or \
-            len(number) == 11 and number.startswith('1')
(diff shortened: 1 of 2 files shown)


behrtam:
Great work! Could you add all missing test cases that are present in the canonical data (v1.2.0)?

N-Parsons:
Sorry, I hadn't realised that there were missing tests - these are now all implemented. I reordered all of the tests to match the order in the canonical data (only one was out of place), but if you think there's a better order, I'm happy to reorganise them.

behrtam:
Thanks!

PR #533, opened 2017-09-25
hello-world: Add a 'running the tests' section

shaleh:
Document how to run the tests and two popular flags.


N-Parsons:
Instructions for running tests with Python exercises are already available on the [language page](http://exercism.io/languages/python/tests), which comes from [python/docs/TEST.md](https://github.com/exercism/python/blob/master/docs/TESTS.md).

I'm not opposed to also including them in the `hello-world` exercise as you've proposed, but I think it would be better to be consistent with the existing documentation surrounding tests, and therefore use `py.test` rather than running the `*_test.py` file directly.

shaleh:
I pulled the python track after finishing the Go track. I never so those docs. Some consistency between tracks would be nice. Go has a nice bit of docs.

On September 26, 2017 4:13:16 AM PDT, Nathan Parsons <notifications@github.com> wrote:
>Instructions for running tests with Python exercises are already
>available on the [language
>page](http://exercism.io/languages/python/tests), which comes from
>[python/docs/TEST.md](https://github.com/exercism/python/blob/master/docs/TESTS.md).
>
>I'm not opposed to also including them in the `hello-world` exercise as
>you've proposed, but I think it would be better to be consistent with
>the existing documentation surrounding tests, and therefore use
>`py.test` rather than running the `*_test.py` file directly.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/exercism/python/pull/533#issuecomment-332166002

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

N-Parsons:
They seem to also have such a section in the READMEs for the Rust track as well. Perhaps they would be a good general addition for all Python exercises? As you say, some people will switch between tracks without looking at the language page on the website, and of course some people might take a break and come back later, so it might be good to have them throughout.

*(As a general side-note, I'm not sure why the "Submitting Exercises" heading is at level 3 and not level 2 for all of the Python exercises - it seems an odd choice.)*

One of the maintainers should be able to give more clarity, but it looks like READMEs are generated from a [template](https://github.com/exercism/python/blob/master/config/exercise_readme.go.tmpl), and have an [extra insert](https://github.com/exercism/python/blob/master/docs/EXERCISE_README_INSERT.md) from the docs folder. I don't know how/when the README is generated from this template though.

shaleh:
I would be happy to make a PR that adds to the inserts if that is what the maintainers would like. So far I have not heard from them on any of the issues I have raised.

m-a-ge:
@shaleh don't take it personally, we didn't mean to ignore your PRs

At this point I am not sure what is the best course of action is to include this info for each exercise. There is an ability to generate these doc files from special templates. This feature (among other things) helps staying in sync with https://github.com/exercism/problem-specifications/ and stuff like that. Looks like this track uses old approach and I'll have to do a bit of research on what the best way off adding this info to all exercises is and come back to you. Is it Ok?

tleen:
We're [a bit behind](https://github.com/exercism/configlet/issues/91) on configlet' documentation for the README generation. It is on the top of my todo list. I have [some docs](https://github.com/exercism/go#exercism-go-style) on the Go track that explain the anatomy of an exercise and the `.meta/*` files that configlet can use to generate the README for an exercise. Still some work to do on that! 

Basically [configlet](https://github.com/exercism/configlet) is a command line tool that can help with a lot of track maintenance tasks. One of which is generating the README's for an exercise using a combination of local track templating, `problem-specifications` repo data, and exercise specific overrides/extra information.

shaleh:
@m-a-ge I can submit a PR adding support for configlet to hello-world if that is helpful to the team.

shaleh:
Adding the .meta directories can be done exercise by exercise. You do not have to move over whole cloth. We can start by updating hello-world. Once everyone has seen what is involved PRs can be made to update the rest of the exercises.

m-a-ge:
@shaleh, ah, I see, it makes sense.

I believe the following file https://github.com/exercism/python/blob/master/docs/TESTS.md is also deserving an update 😉

cmccandless:
@shaleh are you still working on this?

PR #663, opened 2017-10-03
queen-attack: added parameters to function

sharno:
Solving issue #616

### DIFF
--- a/exercises/queen-attack/queen_attack.py
+++ b/exercises/queen-attack/queen_attack.py
@@ -1,6 +1,6 @@
-def board():
+def board(pos1, pos2):
     pass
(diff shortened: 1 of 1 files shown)


exercises/queen-attack/queen_attack.py
```diff
@@ -1,6 +1,6 @@
-def board():
+def board(pos1, pos2):
```
m-a-ge: What about `first_position` and `second_position`? It's more readable to me, don't you agree?

m-a-ge:
Closing this one in favor of #697

@sharno I'm sure you can find something else to work on

PR #670, opened 2017-10-03
Added topics for questions up till Roman numerals

### LINKED ISSUE #826: grade-school: add topics

dvatsav:
fixes #826

### DIFF
--- a/config.json
+++ b/config.json
@@ -104,7 +104,8 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-
+        "mathematics",
+        "logic"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


m-a-ge:
Hi @dvatsav!
Thanks for the PR.
Adding topics is definitely in my TODO list for this track, but I would like to do it in separate PRs to reduce the pain of reviewing suggestions. I'll creating corresponding issues shortly

config.json
```diff
@@ -134,7 +139,9 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-
+        "strings",
+        "mathematics",
+        "text formatting",
```
m-a-ge: Trailing comma is not allowed in JSON

PR #685, opened 2017-10-03
Add Python 3.7 to travis build matrix

behrtam:


behrtam:
Sorry, overlocked that it's only in alpha state ...

PR #691, opened 2017-10-03
collatz-conjecture: Add parameter to placeholder file

### LINKED ISSUE #552: collatz-conjecture: add parameters to exercise placeholder

smt923:
Closes #552, I decided on `n` instead of `number` this time as it matches the README which calls the integer n

### DIFF
--- a/exercises/collatz-conjecture/collatz_conjecture.py
+++ b/exercises/collatz-conjecture/collatz_conjecture.py
@@ -1,2 +1,2 @@
-def collatz_steps():
+def collatz_steps(n):
     pass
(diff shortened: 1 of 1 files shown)


exercises/collatz-conjecture/collatz_conjecture.py
```diff
@@ -1,2 +1,2 @@
-def collatz_steps():
+def collatz_steps(n):
```
m-a-ge: I like `number` better than `n` because it is too ambiguous and we as programmers should avoid such names,
Do you agree?

smt923: Yeah I suppose it is more clear, and regarding my previous stuff about it matching the readme's, I think in this case n == number is pretty clear

m-a-ge:
Thanks, @smt923!

You can find this Python style guide from Google helpful https://google.github.io/styleguide/pyguide.html?showone=Naming#Naming

PR #698, opened 2017-10-04
strain: Add parameters to exercise placeholder #642

### LINKED ISSUE #642: strain: add parameters to exercise placeholder

jnaulty:
Fixes #642


exercises/strain/strain.py
```diff
@@ -1,6 +1,6 @@
-def keep():
+def keep(sequence, predicate):
```
m-a-ge: I like your variant, it's a good one. 
But I think not everyone knows a concept of `predicate`. Maybe it's better to change it to `lambda_expression`?

The other option is to add hits for the exercise explaining a concept of `predicate` (preferable)

What are your thoughts on this?

jnaulty: Well, It seems from the problem statement that a predicate is requried. I don't really think it's necessary for the predicate to be a lambda function, it just has to be any function that takes the item from the list and returns a boolean.

I agree that a hint would be better, I found this little snippet from wikipedia:
```
a predicate P(x) will be true or false, depending on whether x belongs to a set.
```
https://en.wikipedia.org/wiki/Predicate_(mathematical_logic)

jnaulty: Ahh, perhaps just add the predicate definition in the README.md?

m-a-ge: I am not sure actually, since there is an ability to generate these files from special templates. It helps staying in sync with https://github.com/exercism/problem-specifications/ and stuff like that.

Looks like this track uses old approach and I'll have to do a bit of research on what is the best way to add hints to exercises and come back to you. Is it Ok?

jnaulty: ya, when I get some more time I can take a look too.

m-a-ge:
@jnaulty sorry for the late reply.

Following our previous conversation, will it be possible to put hints in [`.meta/hint.md`](https://github.com/exercism/configlet#hints)?

We use [`configlet`](https://github.com/exercism/configlet) to generate `README.md ` files for exercises.

m-a-ge:
@jnaulty I've merged your PR, thanks for your help!

I'll revisit our process of generating READMEs soon (hopefully) and move your definition you `predicate` in 'hints' section

jnaulty:
thanks @m-a-ge

PR #702, opened 2017-10-04
robot-simulator: add parameters to exercise placeholder

### LINKED ISSUE #622: robot-simulator: add parameters to exercise placeholder

parinporecha:
Resolves #622

### DIFF
--- a/exercises/robot-simulator/robot_simulator.py
+++ b/exercises/robot-simulator/robot_simulator.py
@@ -1,3 +1,3 @@
 class Robot(object):
-    def __init__(self):
+    def __init__(self, bearing=NORTH, x=0, y=0):
         pass
--- a/exercises/robot-simulator/robot_simulator.py
+++ b/exercises/robot-simulator/robot_simulator.py
@@ -1,8 +1,9 @@
 # Globals for the bearings
-EAST = "East"
-NORTH = "North"
-WEST = "West"
-SOUTH = "South"
+# Change the values as you see fit
+EAST = None
+NORTH = None
+WEST = None
+SOUTH = None
 
 
 class Robot(object):
--- a/exercises/robot-simulator/robot_simulator.py
+++ b/exercises/robot-simulator/robot_simulator.py
@@ -1,3 +1,5 @@
+NORTH, EAST, SOUTH, WEST = range(4)
+
 class Robot(object):
     def __init__(self, bearing=NORTH, x=0, y=0):
         pass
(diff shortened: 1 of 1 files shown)


N-Parsons:
I'm really not keen on having default values for the bearings. There are other ways to implement bearings that can actually be quite useful, so I don't think that we should push people down one route.

For an example of an alternative, see [my solution](http://exercism.io/submissions/ea15ff22414b4c8d971a0aef6a4d43db).

parinporecha:
@N-Parsons I think you're right. I've made a commit setting the bearing enums to strings -

```
# Globals for the bearings
EAST = "East"
NORTH = "North"
WEST = "West"
SOUTH = "South"
```
Let me know if this is okay.

N-Parsons:
@parinporecha, my point was more that giving any value to the bearings in the exercise suggests that that's how they should be handled, particularly to new programmers.

In this exercise, there are actually a number of ways that the bearing values can be useful, so I think part of the exercise should be deciding what these values should be.

For example, I used the following for my solution (with sine and cosine):
```
EAST = 0
NORTH = 90
WEST = 180
SOUTH = 270
```

But I've also seen equally valid and sensible solutions using tuples of (dx, dy):
```
EAST = (1, 0)
NORTH = (0, 1)
WEST = (-1, 0)
SOUTH = (0, -1)
```

I don't think that the exercise should favour any one of these solutions.

Ultimately, it's not the end of the world if there are pre-assigned values (people can just change them), but I still feel like it takes away some freedom and creativity.

m-a-ge:
This is a good discussion.
I personally believe it's more important to provide default params, even though it might encourage people to use certain way of solving the exercise over the other. Sometimes you just out of ideas how to organize things and need a direction, without direction you'll tend to drop the exercise.

In this particular case, I suggest declaring variables as:
```python
EAST = None
NORTH = None
WEST = None
SOUTH = None
```
and let the learner decide on their representation.

What do you think?

N-Parsons:
@m-a-ge, I think your solution strikes a good balance. It gives people the variable names in the way that the tests require, while leaving it to them to choose the values.

parinporecha:
Sounds good @m-a-ge @N-Parsons

m-a-ge:
@N-Parsons thanks for your insightful comments

@parinporecha thanks for your work. It's a best practice to squash your commits https://github.com/exercism/docs/blob/master/contributing/git-basics.md#squashing. It isn't a big deal since I can do it myself, but it would be awesome to just click on 'Merge pull request' button 😃 

It makes our work much easier. I hope this hint will help you to make your future PRs a bit better

parinporecha:
@m-a-ge Sure, I'll keep that in mind for the future. Thanks for reviewing!

PR #723, opened 2017-10-05
leap: replace assertFalse with assertIs

olufotebig:
* replace assertFalse with assertIs because the test passes if the tested method returns None
* replace assertTrue with assertIs for consistency.


m-a-ge:
Good catch, @olufotebig 
Thanks for taking the time and fixing this! 🛠

PR #756, opened 2017-10-06
Implement exercise diffie-hellman

### LINKED ISSUE #747: diffie-hellman: implement exercise

kusti8:
-----------
Fixes #747


kusti8:
Done!

exercises/diffie-hellman/example.py
```diff
@@ -0,0 +1,13 @@
+import random
+
+
+def private_key(p):
+    return random.randint(2, p-1)
```
N-Parsons: I think that this should use `secrets` rather than `random`, since the exercise is a cryptographic one. Perhaps we should also have a note in README.md that solutions should avoid using `random` since it's not cryptographically secure?

```
def private_key(p):
    return 2 + secrets.randbelow(p-2)
```

kusti8: I did look at that, but according to the [PEP](https://www.python.org/dev/peps/pep-0506/) it was introduced in version 3.6, so it isn't compatible with Python 2 or most Python 3 versions on stable distros. It seems that exercism wants cross compatibility which is why I chose to use the default random instead.

m-a-ge: Good points!

These are just training exercises, so `random` is good enough I think.
But it would be a really good idea to mention `secrets` in `HINTS.md` for such exercises.

@N-Parsons, could you please create an issue for it to not forget?

kusti8: I'll add a note to the readme about it and pseudo randomness.

exercises/diffie-hellman/README.md
```diff
@@ -0,0 +1,58 @@
...
+Bob calculates
+
+    s = A**b mod p
+
+The calculations produce the same result!  Alice and Bob now share
+secret s.
+
+## Notes
+
+Python, as of version 3.6, includes two different random modules. The module called `random` is pseudo-random, meaning it does not generate true randomness, but follows and algorithm that simulates randomness. Since random numbers are generated through a known algorithm, they are not truly random. The `random` module is not correctly suited for crypotography and should not be used, because it is pseudo-random. In version 3.6, Python introduced the `secrets` module which is more cryptographically secure and produces more random numbers suited for cryptography. Since this is only an exercise, `random` is fine to use, but note that it would be very insecure if actually used for crypotgraphy.
```
N-Parsons: @kusti8 Looks good. My only suggestion would be to reword the penultimate sentence to:

"In version 3.6, Python introduced the `secrets` module, which generates cryptographically strong random numbers that provide the greater security required for cryptography."

exercises/diffie-hellman/example.py
N-Parsons: Sorry, @kusti8, I hadn't realised that `secrets` was only in Python 3.6.

@m-a-ge, I'll create an issue for it now :)

exercises/diffie-hellman/README.md
N-Parsons: There are also a typos in the middle and at the end:
"crypotography" --> "cryptography"
"crypotgraphy" --> "cryptography"

kusti8: Fixed.

m-a-ge:
@N-Parsons thanks for the review

@kusti8 thanks a lot for working on this!

PR #757, opened 2017-10-07
[WIP] Implement exercise bank-account

### LINKED ISSUE #730: bank-account: implement exercise

aes421:
---------
Resolves #730

Closes #1260 (mutual-exclusion)

### DIFF
--- a/config.json
+++ b/config.json
@@ -500,6 +500,17 @@
       ]
     },
     {
+      "uuid": "83a3ff95-c043-401c-bc2c-547d52344b02",
+      "slug": "bank-account",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "Classes",
+        "Concurrency"
+      ]
+    },
+    {
       "uuid": "d41238ce-359c-4a9a-81ea-ca5d2c4bb50d",
       "slug": "twelve-days",
       "core": false,
--- a/config.json
+++ b/config.json
@@ -506,8 +506,9 @@
       "unlocked_by": null,
       "difficulty": 4,
       "topics": [
-        "Classes",
-        "Concurrency"
+        "classes",
+        "concurrency",
+        "conditionals"
       ]
     },
     {
--- a/exercises/bank-account/README.md
+++ b/exercises/bank-account/README.md
@@ -1,0 +1,22 @@
+# Bank Account
+
+Simulate a bank account supporting opening/closing, withdrawals, and deposits of money. Watch out for concurrent transactions!
+
+A bank account can be accessed in multiple ways. Clients can make deposits and withdrawals using the internet, mobile phones, etc. Shops can charge against the account.
+
+Create an account that can be accessed from multiple threads/processes (terminology depends on your programming language).
+
+It should be possible to close an account; operations against a closed account must fail.
+
+### Submitting Exercises
+
+Note that, when trying to submit an exercise, make sure the solution is in the `exercism/python/<exerciseName>` directory.
+
+For example, if you're submitting `bob.py` for the Bob exercise, the submit command would be something like `exercism submit <path_to_exercism_dir>/python/bob/bob.py`.
+
+
+For more detailed information about running tests, code style and linting,
+please see the [help page](http://exercism.io/languages/python).
+
+## Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/bank-account/bank_account.py
+++ b/exercises/bank-account/bank_account.py
@@ -1,3 +1,4 @@
+<<<<<<< HEAD
 
 
 class BankAccount(object):
@@ -19,3 +20,29 @@ class BankAccount(object):
 
     def close(self):
         pass
+=======
+import threading
+
+class BankAccount(object):
+	def __init__(self):
+		pass
+
+	def getBalance(self):
+		pass
+
+	def open(self):
+		pass
+
+	def deposit(self, amount):
+		pass
+
+
+	def withdraw(self, amount):
+		pass
+
+	def close(self):
+		pass
+
+
+
+>>>>>>> b8e7ac1... bank-account: exercise configuration
--- a/exercises/bank-account/bank_account.py
+++ b/exercises/bank-account/bank_account.py
@@ -1,0 +1,21 @@
+
+
+class BankAccount(object):
+
+    def __init__(self):
+        pass
+
+    def getBalance(self):
+        pass
+
+    def open(self):
+        pass
+
+    def deposit(self, amount):
+        pass
+
+    def withdraw(self, amount):
+        pass
+
+    def close(self):
+        pass
--- a/exercises/bank-account/bank_account.py
+++ b/exercises/bank-account/bank_account.py
@@ -1,4 +1,3 @@
-<<<<<<< HEAD
 
 
 class BankAccount(object):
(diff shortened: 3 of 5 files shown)


config.json
```diff
@@ -500,6 +500,17 @@
       ]
     },
     {
+      "uuid": "83a3ff95-c043-401c-bc2c-547d52344b02",
+      "slug": "bank-account",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "Classes",
```
m-a-ge: We've switched to `snake_case` naming convention for topics to increase consistency among tracks, so could you please choose them from the following list - https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt

exercises/bank-account/example.py
```diff
@@ -0,0 +1,44 @@
+import threading
+
+
+class BankAccount(object):
+    def __init__(self):
+        self.lock = threading.Lock()
+
+    def getBalance(self):
+        self.lock.acquire()
+        if (self.open is True):
```
m-a-ge: Probably just: 
```python
if self.open:
    # do stuff
```

exercises/bank-account/example.py
```diff
@@ -0,0 +1,44 @@
...
+            self.lock.release()
+            return self.balance
+        else:
+            self.lock.release()
+            raise Exception
```
m-a-ge: Could it you please throw a narrower Exception, or at least state the root cause by:
```python
raise Exception('Describe what went wrong') 
```

exercises/bank-account/example.py
```diff
@@ -0,0 +1,44 @@
+import threading
+
+
+class BankAccount(object):
+    def __init__(self):
+        self.lock = threading.Lock()
```
m-a-ge: I think it would be a good idea to initialize `open` flag in constructor as well:
```python
self.open = False
```

aes421:
Thank you @m-a-ge  I will get on these as soon as I have a chance :)

exercises/bank-account/bank_account_test.py
```diff
@@ -0,0 +1,128 @@
...
+        self.account.withdraw(50)
+
+        self.assertEqual(self.account.getBalance(), 50)
+
+    def test_checking_balance_of_closed_account_throws_error(self):
+        self.account = BankAccount()
+        self.account.open()
+        self.account.close()
+
+        with self.assertRaises(Exception):
```
cmccandless: I would suggest something more specific than `Exception`; perhaps [`ValueError`](https://docs.python.org/2/library/exceptions.html#exceptions.ValueError)?

exercises/bank-account/example.py
cmccandless: Perhaps [`ValueError`](https://docs.python.org/2/library/exceptions.html#exceptions.ValueError) would be better?

N-Parsons:
Hi @aes421, are you still interesting in working on this?

aes421:
@N-Parsons Life got in the way of coding for a few weeks.  About to work on some of these fixes right now!

aes421:
@cmccandless Is there anything else I need to do?

cmccandless:
@aes421 Not as far as I can tell, but since @m-a-ge previously had comments on this, I'd like to give him the chance to comment again.

exercises/bank-account/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Simulate a bank account supporting opening/closing, withdrawals, and deposits of money. Watch out for concurrent transactions!
+
+A bank account can be accessed in multiple ways. Clients can make deposits and withdrawals using the internet, mobile phones, etc. Shops can charge against the account.
+
+Create an account that can be accessed from multiple threads/processes (terminology depends on your programming language).
+
+It should be possible to close an account; operations against a closed account must fail.
+
+### Submitting Exercises
```
N-Parsons: This should be at heading level 2 (`##`).

exercises/bank-account/bank_account.py
```diff
@@ -0,0 +1,21 @@
+
+
+class BankAccount(object):
+
+    def __init__(self):
+        pass
+
+    def getBalance(self):
```
N-Parsons: Following [PEP8](https://www.python.org/dev/peps/pep-0008/) we use `snake_case` for function and method names, so this should be `get_balance`.

exercises/bank-account/bank_account_test.py
```diff
@@ -0,0 +1,125 @@
...
+        self.account.open()
+
+        threadlist = []
+        threads = 100
+        i = 0
+        while (i < threads):
+            thread = threading.Thread(target=increment_and_decrement,
+                                      args=(self, ))
+            threadlist.append(thread)
+            i += 1
```
N-Parsons: This loop would be better implemented using `for i in range(100):`.

exercises/bank-account/bank_account_test.py
```diff
@@ -0,0 +1,125 @@
...
+            self.account.withdraw(-50)
+
+    def test_cannot_deposit_negative(self):
+        self.account = BankAccount()
+        self.account.open()
+
+        with self.assertRaises(ValueError):
+            self.account.deposit(-50)
+
+    def test_can_handle_concurrent_transactions(self):
```
N-Parsons: @aes421 (cc: @cmccandless, @m-a-ge)

I'm not sure that this test is truly testing for concurrency - depositing and then immediately withdrawing inherently safe values within one function is never going to break as far as I can see. Further, I can remove all instances of `self.lock` from the example code and the tests still pass. Part of the issue is probably that the Global Interpreter Lock (GIL) already locks the `BankAccount` instance when a thread accesses it, and releases it once the thread ends. I'm not sure of the best way to get around this.

Currently, Python seems to be running these threads in order, depositing `10` and withdrawing `1`, then moving on to the next thread. We can see this by storing a list of transaction values and using `i+1` as a second argument for the `increment_and_decrement` function, depositing and withdrawing this value, and then printing the transaction list after all threads are done.

One possible solution to get disorder seems to be to add `time.sleep(random.random())` before deposits or withdrawals, but I think the deposits and withdrawal should be decoupled and not part of the same function call - just joining a list of threads for each should be sufficient if there are random delays.

Concurrency was discussed a little in #1106. If we want to keep concurrency in this exercise (I think the exercise would be ok without it, but only as an early exercise about classes), I think that some consideration needs to be made for how to make the tests only pass if it is safe concurrent implementation.

Ultimately, for concurrency to make sense here, we need to be able to make it so that the concurrency matters in some way.

cmccandless: @N-Parsons (cc: @aes421, @m-a-ge)

>I think the exercise would be ok without it

I disagree. It's a core part of the exercise. From the [problem description](https://github.com/exercism/problem-specifications/blob/4d2489f8502614e49a8f041f23dd85600d65658f/exercises/bank-account/description.md):
>Create an account that can be accessed from **multiple threads/processes** (terminology depends on your programming language).

If we marked another exercise as foregone simply because of the challenges of testing parallelism in Python, then I think we might need to consider foregoing this exercise as well. Either that, or we need to come up with another solution and apply it both to this exercise and the foregone `parallel-letter-frequency`.

cmccandless:
Pinning this (pending further discussion on concurrency).

exercises/bank-account/bank_account_test.py
Dog: If we want to be pedantic with this exercise, we could do something similar to NumPy and write an extension in C that runs outside of the GIL. However we can only safely do that if we use no Python C API function used.

I also want to quickly put a note here because I don't think concurrency and parallelism have been discussed much in this repo. There is a difference between concurrency and parallelism.

"In programming, concurrency is the composition of independently executing processes, while parallelism is the simultaneous execution of (possibly related) computations." - from golang.org

https://vimeo.com/49718712 (Video of Rob Pike talking about the difference at Heroku)

Dog: We also might be able to use Pools:

https://docs.python.org/2/library/multiprocessing.html

PR #762, opened 2017-10-07
connect: Implement the exercise "connect"

yunchih:
Implement https://github.com/exercism/python/issues/733


exercises/connect/connect_test.py
```diff
@@ -0,0 +1,113 @@
+import unittest
+import connect
+
```
m-a-ge: Could you please also leave a comment stating what version of `canonical-data.json` the tests were adopted as discussed in #784 if aplicable?
The format is:
```text
```

exercises/connect/connect.py
```diff
@@ -0,0 +1,4 @@
+
+class ConnectGame:
+    def __init__(self, board):
+        pass
```
cmccandless: The solution template doesn't match its usage in `connect_test.py`. The solution template should only contain placeholders for all classes and methods referenced in the tests. `connect_test.py` does not reference `connect.ConnectGame`, but it does reference `connect.play()`, so that should be in the template.

exercises/connect/connect_test.py
yunchih: updated!

exercises/connect/connect.py
yunchih: Thanks for pointing this out! (-:
Please check the updated one.

exercises/connect/connect_test.py
```diff
@@ -0,0 +1,116 @@
...
+                     O X O O O X O X O
+                      O X X X X X O X O
+                       O O O O O O O X O
+                        X X X X X X X X O """,
+        "winner": "X"
+    },
+]
+
+
+class SieveTest(unittest.TestCase):
```
cmccandless: This should be `class ConnectTest(unittest.TestCase):`

exercises/connect/connect_test.py
```diff
@@ -0,0 +1,116 @@
...
+    def test_game(self):
+        for t in testcases:
```
cmccandless: Can you please replace `t` with a more meaningful name? One common method of naming foreach iterator variables is to just take the singular form of the name of the collection being iterated (i.e. `for testcase in testcases`).

yunchih:
@cmccandless updated as requested.

exercises/connect/connect_test.py
```diff
@@ -0,0 +1,116 @@
+import unittest
+import connect
+
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.1.0
```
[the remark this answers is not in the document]
yunchih: done

m-a-ge:
Merged, thanks @yunchih!

PR #765, opened 2017-10-07
Update sieve.py

teohxuen:


exercises/sieve/sieve.py
```diff
@@ -1,2 +1,9 @@
 def sieve(limit):
-    pass
+    isprimes = [True] * (limit+1)
```
m-a-ge: I'm afraid that `sieve` already has example implementation - https://github.com/exercism/python/blob/master/exercises/sieve/example.py

And `sieve.py` is used a placeholder template which is deliver by CLI to the learners 

Feel free to ask any questions

N-Parsons:
Hi @teohxuen, thanks for your interest in contributing. I'm going to close this PR because the file you've edited is only intended to be a skeleton for learners to build on at [exercism.io](http://exercism.io/languages/python).

There are plenty of ways you can contribute to this repo, and a good starting point would be to tackle one of our open [issues](https://github.com/exercism/python/issues).

PR #772, opened 2017-10-08
list-ops: Update test cases to v2.0.0

### LINKED ISSUE #570: list-ops: add parameters to exercise placeholder

N-Parsons:
I've implemented the tests from the most recent version of the [canonical data (v2.0.0)](https://github.com/exercism/problem-specifications/blob/master/exercises/list-ops/canonical-data.json), and I kept a couple of the old tests that I thought were good additions. This PR also fixes #570.

There are a couple of points that I'm not too sure about though and would invite comment.

**Key questions:**
- Is the new implementation of `append` and `concat` sensible? Or should we revert their behaviour?
- Do we want the suggested optional tests for long lists?

### Altered behaviour of `append` and `concat`

The canonical data seems to suggest implementing `append` and `concat` in a different way to the previous tests, and I'm not sure that I like it. The new `append` function is essentially just the old `concat` function, and the new `concat` takes a list of lists. I've included examples below to demonstrate this change.

**Old version:**
```
append([1, 2], 3) == [1, 2, 3]

concat([1, 2], [3, 4]) == [1, 2, 3, 4]
```

**New version:**
```
append([1, 2], [3]) == [1, 2, 3]
append([1, 2], [3, 4]) == [1, 2, 3, 4]

concat([[1, 2], [3, 4]]) = [1, 2, 3, 4]
concat([[1, 2], [3, 4], [5, 6]]) = [1, 2, 3, 4, 5, 6]
```

My personal preference would be to use the old form of `append` and `concat`, but change `concat` to accept an arbitrary number of arguments so that a list of lists can be unpacked into it, but `concat(list1, list2)` would still work.

### Tests with long lists

There is also the suggestion in the canonical data that we could include tests with long lists to give people a reason to think about the efficiency of their approach. Do you think that these would be a good idea? I'm struggling to think of any particularly inefficient solutions that it would discourage.

### DIFF
--- a/exercises/list-ops/example.py
+++ b/exercises/list-ops/example.py
@@ -1,15 +1,9 @@
-def append(xs, y):
-    xs[len(xs):] = [y]
-    return xs
+def append(xs, ys):
+    return concat([xs, ys])
 
 
-def concat(xs, ys):
-    if not ys:
-        return xs
-    else:
-        for item in ys:
-            xs.append(item)
-        return xs
+def concat(lists):
+    return [elem for lst in lists for elem in lst]
 
 
 def filter_clone(function, xs):
@@ -22,7 +16,6 @@ def length(xs):
 
 def map_clone(function, xs):
     return [function(elem) for elem in xs]
-
 
 
 def foldl(function, xs, acc):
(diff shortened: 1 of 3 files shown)


m-a-ge:
@N-Parsons thanks a ton for working on this 👍

PR #768, opened 2017-10-08
[WIP] Implement exercise binary-search-tree

rodrigocam:
--------------
Resolves #740

PR #773, opened 2017-10-08
Implement exercise binary-search-tree

### LINKED ISSUE #740: binary-search-tree: implement exercise

rodrigocam:
Resolved #740


m-a-ge:
You've created two PRs #773 and #768 
Could you please close one to avoid confusion

exercises/binary-search-tree/example.py
```diff
@@ -0,0 +1,69 @@
...
+                        return inserted
+                elif(value > cur_node.value):
+                    if(cur_node.right_node):
+                        cur_node = cur_node.right_node
+                    else:
+                        cur_node.right_node = TreeNode(value)
+                        inserted = True
+                        return inserted
+
+    def search(self, searched_number):
```
kabiir: I recommend changing searched_number to either `number` or `search_number`, searched is in past tense which means you've already searched for it.

exercises/binary-search-tree/example.py
```diff
@@ -0,0 +1,69 @@
...
+        cur_node = self.root
```
kabiir: Also add a check for `if(self.root is None): return false` which does an early exit if the tree hasn't been constructed.

exercises/binary-search-tree/example.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def print_preorder(self, node):
+        if(node is not None):
+            print(node.value)
+            self.print_preorder(node.left_node)
+            self.print_preorder(node.right_node)
+
+    def print_postorder(self, node):
+        if(node is not None):
+            self.print_preorder(node.left_node)
```
kabiir: Post order should be
```python
self.print_postorder(node.right_node) # Right comes first
self.print_postorder(node.left_node)
print(node.value)
```
Also you're calling `print_preorder()` .

exercises/binary-search-tree/example.py
```diff
@@ -0,0 +1,69 @@
...
+                        cur_node.right_node = TreeNode(value)
+                        inserted = True
+                        return inserted
+
+    def search(self, searched_number):
+        cur_node = self.root
+        found = False
+        while not found:
+            if(cur_node is None):
+                raise ValueError("Element not found")
```
kabiir: You won't need to raise an exception if you do an early `return false`

exercises/binary-search-tree/example.py
rodrigocam: Ok, changed!

exercises/binary-search-tree/example.py
rodrigocam: Ok, checked!

config.json
```diff
@@ -888,9 +888,22 @@
       "uuid": "6b7f7b1f-c388-4f4c-b924-84535cc5e1a0",
       "slug": "hexadecimal",
       "deprecated": true
+    },
```
cmccandless: It would probably be best to move the `config.json` entry for this exercise before all deprecated exercises.

exercises/binary-search-tree/README.md
```diff
@@ -0,0 +1,52 @@
+Insert and search for numbers in a binary tree.
```
N-Parsons: This needs the title of the exercise as a heading: `# Binary Search Tree`

exercises/binary-search-tree/binary_search_tree.py
```diff
@@ -0,0 +1,8 @@
+class TreeNode(object):
+    def __init__(self, value):
+        pass
+
+
+class BinarySearchTree(object):
+    def __init__(self):
+        pass
```
N-Parsons: Can you also include skeletons for the `add` and `search` methods here?

exercises/binary-search-tree/binary_search_tree.py
```diff
@@ -0,0 +1,8 @@
+class TreeNode(object):
+    def __init__(self, value):
+        pass
```
N-Parsons: Can you add `self.value = value` here? I feel like if we assume that something exists in the tests, then we should make it clear in the solution template.

config.json
```diff
@@ -1169,6 +1183,5 @@
     }
   ],
   "foregone": [
-
```
N-Parsons: Can you add this blank line back in?

exercises/binary-search-tree/README.md
```diff
@@ -0,0 +1,52 @@
...
+      3
+
+And if we then added 1, 5, and 7, it would look like this
+
+          4
+        /   \
+       /     \
+      2       6
+     / \     / \
+    1   3   5   7
```
N-Parsons: There should also be some other stuff that goes here. Check another exercise and copy it over, or regenerate this README with the `configlet`.

exercises/binary-search-tree/binary_search_tree_test.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+
+from binary_search_tree import BinarySearchTree
+
+
+class BinarySearchTreeTests(unittest.TestCase):
+    bst = BinarySearchTree()
+
+    def test_add_integer_number1(self):
+        self.assertTrue(self.bst.add(7))
```
cmccandless: We use `assertIs(actual, True)` and `assertIs(actual, False)` rather than `assertTrue(actual)` or `assertFalse(actual)` ([#419](https://github.com/exercism/python/pull/419)).

exercises/binary-search-tree/example.py
```diff
@@ -0,0 +1,69 @@
...
+                        cur_node.left_node = TreeNode(value)
+                        inserted = True
+                        return inserted
+                elif(value > cur_node.value):
+                    if(cur_node.right_node):
+                        cur_node = cur_node.right_node
+                    else:
+                        cur_node.right_node = TreeNode(value)
+                        inserted = True
+                        return inserted
```
N-Parsons: Looking at this again, I just noticed that this method is returning `True`/`False` to indicate success/failure. Under what condition is failure expected to occur? From what I can see, this code will never return `False`, and would actually just enter an infinite loop if `value == cur_node.value`.

I think it would be worth adding a `BinarySearchTree.list` method to this exercise that returns a sorted list of values in the tree. A `collections.deque` would be most efficient for this, but I don't think that the tests should enforce its use, so the tests should probably convert the actual output to a list (eg. `assertEqual(list(self.bst.list()), [1, 2, 3])`).

Doing this would allow us to check that values are placed in the right order - each test for adding a number should contain a second assert for the current list of stored values. We should also test duplicate values, because the README specifies a way for these to be handled.

**From the `README`:**
> All data in the left subtree is less than or equal to the current node's data, and all data in the right subtree is greater than the current node's data.

cmccandless: @N-Parsons Returning `True` from an add or insert method is a common practice for container classes. I agree that it doesn't make the most sense, but that is likely why @rodrigocam did this.

cmccandless:
@rodrigocam Just FYI, canonical-data for this exercise is being created at [/exercism/problem-specifications#940](https://github.com/exercism/problem-specifications/pull/940).

config.json
rodrigocam: Done

exercises/binary-search-tree/README.md
rodrigocam: Done

exercises/binary-search-tree/README.md
rodrigocam: Done

exercises/binary-search-tree/binary_search_tree.py
rodrigocam: Done

exercises/binary-search-tree/binary_search_tree.py
rodrigocam: Done

exercises/binary-search-tree/binary_search_tree_test.py
rodrigocam: Done

exercises/binary-search-tree/example.py
rodrigocam: @cmccandless is right, I thought this could be good to give more safety. I made the list method that @N-Parsons asked and change the tests to use this new method. I fix the infinite loop condition that you point but I keep the bool return, if it was a problem I can remove (:

N-Parsons: Given that there's no way that `False` can be returned, and there is not test for this method returning `False`, I think it's unnecessary and likely to cause some confusion/frustration. I think it's fine for this example code to return `True`/`False`, but I don't think that it should be tested for.

exercises/binary-search-tree/binary_search_tree_test.py
```diff
@@ -0,0 +1,58 @@
...
+        bst.add(7.5)
+        self.assertEqual(bst.search(1).value, 1)
+        self.assertEqual(bst.search(7.5).value, 7.5)
+        
+    def test_search_invalid_numbers(self):
+        bst = BinarySearchTree()
+        bst.add(1)
+        bst.add(7.5)
+        self.assertIs(bst.search(6), False)
+        self.assertIs(bst.search(8.8), False)
```
N-Parsons: I don't like that `search` will sometimes return a `bool` and sometimes return a `TreeNode`. I think that this should either be `TreeNode`/`None`, or `True`/`False` - the former is perhaps more generally useful, since nodes are truthy and `None` is falsey, and returning the actual node could be useful in some instances.

rodrigocam: Done! I change to return None when the node is not found.

exercises/binary-search-tree/example.py
rodrigocam: Done! I take out the return of the method.

config.json
N-Parsons: I looks like you've got some unnecessary white-space here now (see below) - can you please remove it?

exercises/binary-search-tree/binary_search_tree.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class BinarySearchTree(object):
+    def __init__(self):
+        pass
+
+    def add(self, value):
+        pass
+
+    def search(self, search_number):
```
N-Parsons: I think this should use `value` rather than `search_number` as an argument, since we put values in and then search for them, and `search_` is probably redundant.

exercises/binary-search-tree/.cache/v/cache/lastfailed
```diff
@@ -0,0 +1,10 @@
+{
+  "binary_search_tree_test.py::BinarySearchTreeTests::test_add_float_number": true,
+  "binary_search_tree_test.py::BinarySearchTreeTests::test_add_integer_number1": true,
+  "binary_search_tree_test.py::BinarySearchTreeTests::test_add_integer_number2": true,
+  "binary_search_tree_test.py::BinarySearchTreeTests::test_add_integer_number3": true,
+  "binary_search_tree_test.py::BinarySearchTreeTests::test_add_integer_number4": true,
+  "binary_search_tree_test.py::BinarySearchTreeTests::test_search_valid_number1": true,
+  "binary_search_tree_test.py::BinarySearchTreeTests::test_search_valid_number2": true,
+  "binary_search_tree_test.py::BinarySearchTreeTests::test_search_valid_number3": true
+}
```
N-Parsons: You need to remove the `.cache` directory.

N-Parsons:
Hi @rodrigocam, sorry for taking a while to get back to this - I've had a really busy couple of weeks!

There are a couple of minor issues remaining that need to be dealt with, and then this can be merged.

exercises/binary-search-tree/binary_search_tree.py
rodrigocam: Done!

exercises/binary-search-tree/.cache/v/cache/lastfailed
rodrigocam: Done!

rodrigocam: Done!

exercises/binary-search-tree/binary_search_tree.py
N-Parsons: @rodrigocam, it looks like you've updated this in `example.py` but not here.

config.json
```diff
@@ -1185,6 +1199,6 @@
...
+
```
N-Parsons: @rodrigocam, the extra spaces here also need to be removed. There should be a blank line, but there should be nothing in it.

N-Parsons:
@cmccandless, can you review this again? There are a couple of minor changes pending and then I think this will be ready to merge.

exercises/binary-search-tree/binary_search_tree.py
rodrigocam: Ok, done!

exercises/binary-search-tree/example.py
```diff
@@ -0,0 +1,79 @@
...
+        self.trav_inorder(self.root, elements)
+        return elements
+
+    def trav_inorder(self, node, elements):
+        if(node is not None):
+            self.trav_inorder(node.left_node, elements)
+            elements.append(node.value)
+            self.trav_inorder(node.right_node, elements)
+
+    def print_inorder(self, node):
```
cmccandless: We should probably remove all `print*` methods, as they are not used by the tests.

rodrigocam: Done!

N-Parsons:
Looks good. Thanks @rodrigocam!

PR #793, opened 2017-10-12
Change execution order of verification scripts for Travis CI

m-a-ge:
By reading logs from https://travis-ci.org/exercism/python/jobs/286835621 I've noticed that our test script run tests before verifiying high level things with `configlet` which can be improved
This PR is meant to fix that

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,6 @@ before_script:
   - flake8
 
 script:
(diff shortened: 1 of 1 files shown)


m-a-ge:
I don't usually merge my own PRs (because it's a sin of course), but in this particular case I'll make an exception considering nature of the change.

I hope you understand my decision

PR #795, opened 2017-10-12
two-bucket: fix incorrect test case

### LINKED ISSUE #794: two-bucket: Fix incorrect test case
Canonical data was recently changed for this exercise to [fix an incorrect test case](/exercism/problem-specifications/issues/911). The [implementation](https://github.com/exercism/python/blob/master/exercises/two-bucket/two_bucket_test.py#L28) for the updated test data should be updated as well.

cmccandless:
Resolves #794

### DIFF
--- a/exercises/two-bucket/example.py
+++ b/exercises/two-bucket/example.py
@@ -1,59 +1,54 @@
 '''
-The solution follows the following approach:
-
-->Pour from first bucket to second
-  1.Fill the first bucket and empty it into the second bucket.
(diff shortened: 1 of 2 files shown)


m-a-ge:
@cmccandless, thanks for updating this one. Example implementation should also pass the new test

Could you also check to what version of `canonical-data.json` is based on and state it in the comment  inside the file

m-a-ge:
Merged, thanks for fixing 🛠

PR #796, opened 2017-10-12
tree-building : implement exercise

### LINKED ISSUE #741: tree-building: implement exercise

clapmyhands:
resolves #741

### DIFF
--- a/exercises/tree-building/example.py
+++ b/exercises/tree-building/example.py
@@ -1,0 +1,46 @@
+class Record():
+    def __init__(self, record_id, parent_id):
+        self.record_id = record_id
+        self.parent_id = parent_id
+
+    def equal_id(self):
+        return self.record_id == self.parent_id
+
+
+class Node():
+    def __init__(self, node_id):
+        self.node_id = node_id
+        self.children = []
+
+
+def validateRecord(record):
+    if record.equal_id() and record.record_id != 0:
+        # only root(id=0) should have equal id
+        raise ValueError
+    elif not record.equal_id() and record.parent_id >= record.record_id:
+        # non-root node id should be smaller than its parent_id
+        raise ValueError
+
+
+def BuildTree(records):
+    parent_dict = {}
+    node_dict = {}
+    ordered_id = sorted((i.record_id for i in records))
+    for record in records:
+        validateRecord(record)
+        parent_dict[record.record_id] = record.parent_id
+        node_dict[record.record_id] = Node(record.record_id)
+
+    root_id = 0
+    root = None
+
+    for index, record_id in enumerate(ordered_id):
+        if index != record_id:
+            raise ValueError
+        if record_id == root_id:
+            root = node_dict[record_id]
+        else:
+            parent_id = parent_dict[record_id]
+            node_dict[parent_id].children.append(node_dict[record_id])
+
+    return root
--- a/exercises/tree-building/tree_building_test.py
+++ b/exercises/tree-building/tree_building_test.py
@@ -1,0 +1,183 @@
+import unittest
+
+from tree_building import Record, BuildTree
+
+
+class TestBuildingTest(unittest.TestCase):
+    """
+        Record(record_id, parent_id): records given to be processed
+        Node(node_id): Node in tree
+        BuildTree(records): records as argument and returns tree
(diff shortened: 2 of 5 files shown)


exercises/tree-building/example.py
```diff
@@ -0,0 +1,44 @@
...
+
+class Node():
+    def __init__(self, node_id):
+        self.node_id = node_id
+        self.children = []
+
+
+def validateRecord(record):
+    if record.equal_id() and record.record_id != 0:
+        raise ValueError
```
m-a-ge: Should probably provide the cause of the exception

exercises/tree-building/example.py
```diff
@@ -0,0 +1,44 @@
...
+    elif (not record.equal_id()) and record.parent_id >= record.record_id:
+        raise ValueError
```
m-a-ge: the same

exercises/tree-building/example.py
```diff
@@ -0,0 +1,44 @@
...
+class Node():
+    def __init__(self, node_id):
+        self.node_id = node_id
+        self.children = []
+
+
+def validateRecord(record):
+    if record.equal_id() and record.record_id != 0:
+        raise ValueError
+    elif (not record.equal_id()) and record.parent_id >= record.record_id:
```
m-a-ge: redundant parentheses

clapmyhands:
@m-a-ge I've change it according to the review. Please check again and let me know if there's anything else to change 😄 .

exercises/tree-building/tree_building_test.py
```diff
@@ -0,0 +1,178 @@
...
+            Record(0, 0),
+            Record(2, 0),
+            Record(1, 2)
+        ]
+        with self.assertRaises(ValueError):
+            BuildTree(records)
+
+    def assert_node_is_branch(self, node, node_id, children_count):
+        self.assertEqual(node.node_id, node_id)
+        self.assertFalse(len(node.children) == 0)
```
cmccandless: `assertNotEqual(len(node.children), 0)` would be clearer.

exercises/tree-building/tree_building_test.py
```diff
@@ -0,0 +1,178 @@
...
+        self.assertEqual(len(node.children), children_count)
+
+    def assert_node_is_leaf(self, node, node_id):
+        self.assertEqual(node.node_id, node_id)
+        self.assertTrue(len(node.children) == 0)
```
cmccandless: `assertEqual(len(node.children), 0)` would be clearer.

clapmyhands:
@cmccandless I've changed according to your review. Also I've been thinking, do I need to put some reason on the test where a `ValueError` need to be raised?

cmccandless:
@clapmyhands If I understand you correctly, do you mean something like this?
``` Python
with self.assertRaisesRegexp(ValueError, 'No root node'):
    BuildTree(records)
```

If that's what you mean, that would be a good idea :+1:

clapmyhands:
@cmccandless I don't think it's possible. At least in my current code it can't since the checking of 'no root node' is the same as checking of 'non continuous'

reason being:
```python
raise ValueError
```
 can cover multiple failing cases.

I was thinking more in line of comments like so:
```python
# no root node
with self.assertRaises(ValueError):
            BuildTree(records)
```

the name of the test might already be self-explanatory but for some tests such as
```python
def test_cycle_directly(self):
```
it might be good to have further explanation on why that gives an error.

p.s: what do you think about changing
```python
self.assertIs(root, None)
```
to
```python
self.assertIsNone(root)
```

clapmyhands:
@cmccandless I've changed it to use `self.assertIsNone` and added comment where I think is needed. Please review once more and let me know if there's anything else I need to change 😄

cmccandless:
@clapmyhands I've left a couple comments. Just some consistency stuff.

clapmyhands:
@cmccandless I've added comment for consistency.

m-a-ge:
Thanks for working on this, @clapmyhands!

clapmyhands:
@m-a-ge @cmccandless Thanks for being patient with me through it all. New contributor and hopefully a lasting one 😄

Issue #784, opened 2017-10-09
All exercise test files should reference canonical data version


N-Parsons:
Currently some of the test files reference `x-common//canonical-data.json` with a version number, and a few test files lack any such a reference. Additionally, `x-common` was recently renamed to `problem-specifications`, so references to `x-common` are inherently outdated and potentially confusing.

Having a consistent version string should help with maintenance, since a script can fairly easily be written to compare the version numbers in the test file and the canonical data, and manual checks for how up to date it is are similarly easy. It also makes it easier for exercism users to find and check the canonical data, and thus to be able to contribute any changes they think are necessary if they identify problems or additional tests that would be useful.

### Suggested wording

The currently adopted form for the version string seems to be:
``# test cases adapted from `x-common//canonical-data.json` @ version: 1.0.1``

To conform to PEP8, we should aim to limit the string to 79 characters, so I propose the following wording (75 chars):
``# Tests adapted from `problem-specifications//canonical-data.json` @ v2.0.0``

Does anyone have any alternatives or suggestions for improvements to this wording?

### How to proceed

There are a couple of options for how to go about updating the test files:
1. Only update version strings when the tests are next updated, and update tests when the canonical data change.
1. Systematically update each test file to match the most recent canonical data version (or verify that it's already up-to-date), and update the version strings in the process.

Option 1 is obviously the less work-intensive one, but it would leave some inconsistency for quite a while. Option 2 is more work-intensive and should get everything up-to-date quicker, but it might burn people out a bit.

m-a-ge:
Thanks for raising the issue, it's important one.
I think the second option is a way to go and it should be automated. Unfortunately, I have a lot of going on right now, so I can't do it right now

The other thing is that some tests can be trak specific, so it can be also stated in the comment giving the reader a sense of the reason for such decision

N-Parsons:
I had some free time this evening, so I put together a [fairly rough script for checking the state of version numbering](https://gist.github.com/N-Parsons/9a99d9be6df07e470faf7a9310abbd23). I put [the results into a gist](https://gist.github.com/N-Parsons/b68a02c4413099d0d49531daaa7bcc62) to save making a very long post here, and I have summarised the results below.

**Up-to-date:** 33
**Out-of-date:** 8
**Unknown:** 31 *(no version tag)*
**Unimplemented:** 27
**No canonical data:** 14
**No data needed:** 5 *(maybe more - requires manual checks)*

m-a-ge:
[...]

PR #798, opened 2017-10-13
error-handling: implement exercise

### LINKED ISSUE #742: error-handling: implement exercise

cmccandless:
Resolves #742 

TODO:
- [x] update `config.json`
- [x] add README
- [x] create solution template
- [x] write test cases
- [x] write example solution

### DIFF
--- a/config.json
+++ b/config.json
@@ -861,7 +861,7 @@
       "unlocked_by": null,
       "difficulty": 3,
       "topics": [
-        "exception handling"
+        "exception_handling"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -855,6 +855,16 @@
       ]
     },
     {
+      "uuid": "3a2a947a-01b3-1e80-e32b-de1756fd88365adf12e",
+      "slug": "error-handling",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 3,
+      "topics": [
+        "exception handling"
+      ]
+    },
+    {
       "uuid": "e7351e8e-d3ff-4621-b818-cd55cf05bffd",
       "slug": "accumulate",
       "deprecated": true
--- a/exercises/error-handling/README.md
+++ b/exercises/error-handling/README.md
@@ -1,0 +1,27 @@
+# Error Handling
+
+Implement various kinds of error handling and resource management.
+
+An important point of programming is how to handle errors and close
+resources even if errors occur.
+
+This exercise requires you to handle various errors. Because error handling
+is rather programming language specific you'll have to refer to the tests
+for your track to see what's exactly required.
+
+### Submitting Exercises
+
+Note that, when trying to submit an exercise, make sure the solution is in the `exercism/python/<exerciseName>` directory.
+
+For example, if you're submitting `bob.py` for the Bob exercise, the submit command would be something like `exercism submit <path_to_exercism_dir>/python/bob/bob.py`.
+
+
+For more detailed information about running tests, code style and linting,
+please see the [help page](http://exercism.io/languages/python).
+
+## Source
+
+Problem 6 at Project Euler [http://projecteuler.net/problem=6](http://projecteuler.net/problem=6)
+
+## Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/error-handling/README.md
+++ b/exercises/error-handling/README.md
@@ -19,9 +19,5 @@ For example, if you're submitting `bob.py` for the Bob exercise, the submit comm
 For more detailed information about running tests, code style and linting,
 please see the [help page](http://exercism.io/languages/python).
 
-## Source
-
-Problem 6 at Project Euler [http://projecteuler.net/problem=6](http://projecteuler.net/problem=6)
-
 ## Submitting Incomplete Solutions
 It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/error-handling/error_handling_test.py
+++ b/exercises/error-handling/error_handling_test.py
@@ -29,7 +29,7 @@ class ErrorHandlingTest(unittest.TestCase):
         self.assertEqual(1, er.handle_error_by_returning_none('1'),
                          'Result of valid input should not be None')
         self.assertIsNone(er.handle_error_by_returning_none('a'),
-                        'Result of invalid input should be None')
+                          'Result of invalid input should be None')
 
     def test_return_tuple(self):
         successful_result, result = er.handle_error_by_returning_tuple('1')
--- a/exercises/error-handling/error_handling_test.py
+++ b/exercises/error-handling/error_handling_test.py
@@ -19,8 +19,9 @@ class FileLike(object):
 
     def __enter__(self):
         self.open()
+        return self
 
-    def __exit__(self):
+    def __exit__(self, *args):
         self.close()
 
     def do_something(self):
@@ -50,7 +51,6 @@ class ErrorHandlingTest(unittest.TestCase):
 
     def test_filelike_objects_are_closed_on_exception(self):
         filelike_object = FileLike()
-        filelike_object.open()
         with self.assertRaises(Exception):
             er.filelike_objects_are_closed_on_exception(filelike_object)
         self.assertFalse(filelike_object.is_open,
@@ -58,7 +58,7 @@ class ErrorHandlingTest(unittest.TestCase):
         self.assertTrue(filelike_object.was_open,
                         'filelike_object should have been opened')
         self.assertTrue(filelike_object.did_something,
-                        'filelike_object should call did_something()')
+                        'filelike_object should call do_something()')
 
 
 if __name__ == '__main__':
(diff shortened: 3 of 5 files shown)


config.json
```diff
@@ -855,6 +855,16 @@
       ]
     },
     {
+      "uuid": "3a2a947a-01b3-1e80-e32b-de1756fd88365adf12e",
+      "slug": "error-handling",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 3,
+      "topics": [
+        "exception handling"
```
m-a-ge: We've switched to `snake_case` naming convention for topics to increase consistency among tracks, so could you please choose them from the following list - https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt

exercises/error-handling/error_handling_test.py
```diff
@@ -0,0 +1,52 @@
+import unittest
+
+import error_handling as er
+
```
m-a-ge: Could you please also leave a comment stating what version of `canonical-data.json` the tests were adopted as discussed in #784 if aplicable?
The format is:
```text
```

exercises/error-handling/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+For example, if you're submitting `bob.py` for the Bob exercise, the submit command would be something like `exercism submit <path_to_exercism_dir>/python/bob/bob.py`.
+
+
+For more detailed information about running tests, code style and linting,
+please see the [help page](http://exercism.io/languages/python).
+
+## Source
+
+Problem 6 at Project Euler [http://projecteuler.net/problem=6](http://projecteuler.net/problem=6)
```
N-Parsons: This source isn't right - I presume that there's actually no source needed for this exercise?

exercises/error-handling/example.py
```diff
@@ -0,0 +1,22 @@
...
+def handle_error_by_returning_tuple(input_data):
+    try:
+        return (True, int(input_data))
+    except ValueError:
+        return (False, None)
+
+
+def filelike_objects_are_closed_on_exception(filelike_object):
+    with filelike_object:
+        filelike_object.close()
```
N-Parsons: This isn't closing the filelike object on an exception, it's closing the filelike object and then raising an exception afterwards.

A better solution might be to have a `was_open` property and `do_something` method in the `FileLike` class, with the latter just raising some form of exception. Then it should be up to the user to call `open`, `do_something`, and `close` in that order. The tests can check that an exception was raised, that `is_open` is `False`, and that `was_open` is `True`.

It'd probably be worth using the exercise's `.meta/hints.md` file to include something in the README about what methods exist in the `filelike_object`. The `.meta/hints.md` file is then included into the README when it's generated using the `configlet` - it's fine to just use copy-and-paste to put together the README, but also putting it in the hints file makes sure that it won't accidentally get lost in future.

N-Parsons: Also, as a learner I would expect that a context manager could be used here (eg. `with open(filelike_object) as f`:) - it's the Pythonic way to handle files. I'm not sure what would need to be implemented in the `FileLike` class to make this work though. If we can't get a context manager working, then I think that the README needs to tell learners that context managers would be the typical way to do it with files, but that they won't work with the custom class here, because otherwise there is a massive amount of room for confusion and frustration.

exercises/error-handling/error_handling_test.py
cmccandless: At this time, there is no canonical data for this exercise; these tests were adapted from the [C# implemetation of this exercise](https://github.com/exercism/csharp/blob/f838767eab422738c8d8517b910dfcd926602969/exercises/error-handling/ErrorHandlingTest.cs). Is there a standard way of noting this?

exercises/error-handling/README.md
cmccandless: Correct; I shall remove it. The included source section was in the README from another exercise from which I created this one.

exercises/error-handling/example.py
cmccandless: The built-in `open()` method only takes a file name; you cannot implement a class to support `open(filelike_object)`. However, I like the idea of `do_something()` raising an exception. See my latest changes.

exercises/error-handling/README.md
m-a-ge: @cmccandless if it was based on `canonical-data`, it will be good to open PR there to fix it

exercises/error-handling/error_handling_test.py
m-a-ge: If there is no tests in problem-specifications repo for this exercise, then it's Ok to leave it as is.

We agreed with @N-Parsons to use comment above to denote the version of `canonical-data.json` from which tests were adopted

exercises/error-handling/README.md
cmccandless: It was not. I copied the exercise-specific sections from the C# README and the track-specific sections (submitting, incomplete, etc) from another Python exercise. The source section was accidentally copied with the info from the other Python exercise.

exercises/error-handling/error_handling_test.py
```diff
@@ -6,18 +6,26 @@
...
-    def __enter__(self):
+        self.was_open = False
         self.is_open = True
 
     def close(self):
         self.is_open = False
+        self.was_open = True
+
+    def __enter__(self):
+        self.open()
```
N-Parsons: If you add `return self` here (plus make the change to `__exit__`), then this class will work with a context manager.

exercises/error-handling/error_handling_test.py
```diff
@@ -6,18 +6,26 @@
...
 
     def __exit__(self):
```
N-Parsons: `__exit__` needs to accept 4 arguments, but we don't need them here, so we can just use `def __exit__(self, *args):`

exercises/error-handling/error_handling_test.py
```diff
@@ -45,7 +53,12 @@ def test_filelike_objects_are_closed_on_exception(self):
         filelike_object.open()
```
N-Parsons: I don't think that we should be passing an open object to the user's function. It seems strange to me that we'd pass something open to a function that (if we weren't just throwing an exception) may or may not close it - I would expect closing to be handled in the place that it's opened. I think it should be up to the user to open and close the filelike object however they see fit.

exercises/error-handling/example.py
```diff
@@ -18,5 +18,8 @@ def handle_error_by_returning_tuple(input_data):
 
 def filelike_objects_are_closed_on_exception(filelike_object):
     with filelike_object:
-        filelike_object.close()
-        raise Exception()
+        try:
+            filelike_object.do_something()
+        except Exception:
+            filelike_object.close()
+            raise
```
N-Parsons: You're actually mixing two error handling methods here - you have a context manager, but then you are explicitly closing an object using `try ... except` blocks to handle the exception behaviour.

Using `try ... finally`: (the `finally` block allows you to close the connection regardless of whether an exception is raised or not, and just doing `except: raise` is redundant)

```python
filelike_object.open()
try:
    filelike_object.do_something()
finally:
    filelike_object.close()
```

Alternatively, a context manager can be used:

```python
with filelike_object:
    filelike_object.do_something()
```

Or a context manager with an assignment:

```python
with filelike_object as infile:
    infile.do_something()
```

cmccandless: I had forgotten about `finally`; thanks.

exercises/error-handling/example.py
```diff
@@ -17,9 +17,8 @@ def handle_error_by_returning_tuple(input_data):
...
-    with filelike_object:
+    with filelike_object as fobj:
         try:
-            filelike_object.do_something()
-        except Exception:
-            filelike_object.close()
-            raise
+            fobj.do_something()
+        finally:
+            fobj.close()
```
N-Parsons: As I stated previously, you don't need both a context manager and a `try` block for this - they are two different methods of error handling. Either you use a context manager, which will call the `__enter__` and `__exit__` methods that will open and close the object, OR you handle opening and closing explicitly, using a `try ... finally` block to ensure that the object is closed once the `try` block has finished.

This function should either be:
```python
def filelike_objects_are_closed_on_exception(filelike_object):
    filelike_object.open()
    try:
        filelike_object.do_something()
    finally:
        filelike_object.close()
```

Or:
```python
def filelike_objects_are_closed_on_exception(filelike_object):
    with filelike_object as fobj:
        fobj.do_something()
```

Or:
```python
def filelike_objects_are_closed_on_exception(filelike_object):
    with filelike_object:
        filelike_object.do_something()
```

cmccandless: Apologies, I had somehow managed to convince myself that when an exception is raised from within the context manager object itself, `__exit__()` would not necessarily be called correctly, thus the need for `try... finally` inside `with`. Fix forthcoming...

cmccandless:
@N-Parsons Replaced assertTrue and assertFalse with assertIs.

cmccandless:
@N-Parsons Sorry, missed that one. I should be covered now.

N-Parsons:
Thanks for your work on this, @cmccandless

PR #805, opened 2017-10-14
allergies: replace assertFalse with assertIs

olufotebig:
* replace assertFalse with assertIs because the test passes if the tested method returns None
* replace assertTrue with assertIs for consistency.


N-Parsons:
In case you're wondering what the error is (I got curious and checked :P), it's that `example.py` is returning an `int` instead of a `bool`, which works fine with `assertFalse`, but not `assertIs`. The easy solution is to add `bool()` on the return statement for `Allergies.is_allergic_to`.

N-Parsons:
@olufotebig, don't forget to update `example.py` so that this passes the tests and can be merged :)

olufotebig:
Let do that now

m-a-ge:
Thanks, @olufotebig

PR #807, opened 2017-10-14
Update version strings to reference problem-specifications

N-Parsons:
Update version strings as per #784.


m-a-ge:
Thanks a lot for this  ❤️ , @N-Parsons

PR #841, opened 2017-10-14
queen_attack: replace assertFalse with assertIs

olufotebig:
replace assertFalse with assertIs because the test passes if the tested method returns None
replace assertTrue with assertIs for consistency.


m-a-ge:
Thanks, @olufotebig 

After restarting the build it passed

PR #890, opened 2017-10-14
[WIP] isbn-verifier: implement exercise

eulercamposbarros:
A detailed description of this exercise can be found [here](https://github.com/exercism/problem-specifications/blob/master/exercises/isbn-verifier/description.md).

TODO:
- [x] ISBN-10 verifier
- [x] ISBN-10 generator from a starting ISBN (bonus task)
- [x] ISBN-13 generator (bonus task)
- [ ] Suggested changes

### DIFF
--- a/config.json
+++ b/config.json
@@ -1112,7 +1112,27 @@
       "unlocked_by": null,
       "difficulty": 3,
       "topics": [
+        "type_conversion",
         "exception_handling"
+        "strings",
+        "arrays",
+        "integers",
+        "parsing"
+      ]
+    },
+    {
+      "uuid": "7961c852-c87a-44b0-b152-efea3ac8555c",
+      "slug": "isbn-verifier",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 3,
+      "topics": [
+        "type_conversion",
+        "conditionals",
+        "strings",
+        "arrays",
+        "integers",
+        "parsing"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -70,6 +70,21 @@
       ]
     },
     {
+      "uuid": "7961c852-c87a-44b0-b152-efea3ac8555c",
+      "slug": "isbn-verifier",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "type_conversion",
+        "conditionals",
+        "strings",
+        "arrays",
+        "integers",
+        "parsing"
+      ]
+    },
+    {
       "uuid": "8648fa0c-d85f-471b-a3ae-0f8c05222c89",
       "slug": "hamming",
       "core": false,
--- a/config.json
+++ b/config.json
@@ -1128,22 +1128,7 @@
       "difficulty": 3,
       "topics": [
         "type_conversion",
-        "exception_handling"
-        "strings",
-        "arrays",
-        "integers",
-        "parsing"
-      ]
-    },
-    {
-      "uuid": "7961c852-c87a-44b0-b152-efea3ac8555c",
-      "slug": "isbn-verifier",
-      "core": false,
-      "unlocked_by": null,
-      "difficulty": 3,
-      "topics": [
-        "type_conversion",
-        "conditionals",
+        "exception_handling",
         "strings",
         "arrays",
         "integers",
--- a/exercises/isbn-verifier/README.md
+++ b/exercises/isbn-verifier/README.md
@@ -1,0 +1,43 @@
+# Isbn Verifier
+
+Check if a given ISBN-10 is valid.
+
+## Functionality
+
+Given an unkown string the program should check if the provided string is a valid ISBN-10.
+Putting this into place requires some thinking about preprocessing/parsing of the string prior to calculating the check digit for the ISBN.
+
+The program should allow for ISBN-10 without the separating dashes to be verified as well.
+
+## ISBN
+
+Let's take a random ISBN-10 number, say `3-598-21508-8` for this.
+The first digit block indicates the group where the ISBN belongs. Groups can consist of shared languages, geographic regions or countries. The leading '3' signals this ISBN is from a german speaking country.
+The following number block is to identify the publisher. Since this is a three digit publisher number there is a 5 digit title number for this book.
+The last digit in the ISBN is the check digit which is used to detect read errors.
+
+The first 9 digits in the ISBN have to be between 0 and 9.
+The check digit can additionally be an 'X' to allow 10 to be a valid check digit as well.
+
+A valid ISBN-10 is calculated with this formula `(x1 * 10 + x2 * 9 + x3 * 8 + x4 * 7 + x5 * 6 + x6 * 5 + x7 * 4 + x8 * 3 + x9 * 2 + x10 * 1) mod 11 == 0`
+So for our example ISBN this means:
+(3 * 10 + 5 * 9 + 9 * 8 + 8 * 7 + 2 * 6 + 1 * 5 + 5 * 4 + 0 * 3 + 8 * 2 + 8 * 1) mod 11 = 0
+
+Which proves that the ISBN is valid.
+
+### Submitting Exercises
+
+Note that, when trying to submit an exercise, make sure the solution is in the `exercism/python/<exerciseName>` directory.
+
+For example, if you're submitting `bob.py` for the Bob exercise, the submit command would be something like `exercism submit <path_to_exercism_dir>/python/bob/bob.py`.
+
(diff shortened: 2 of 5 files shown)


exercises/isbn-verifier/example.py
```diff
@@ -0,0 +1,46 @@
...
+                        int(isbn[2:3]) * 8 +
+                        int(isbn[3:4]) * 7 +
+                        int(isbn[4:5]) * 6 +
+                        int(isbn[5:6]) * 5 +
+                        int(isbn[6:7]) * 4 +
+                        int(isbn[7:8]) * 3 +
+                        int(isbn[8:9]) * 2) % 11
+
+    if check_digit == 10:
+        return 'X'
```
jamesmcm: This is quite a lot of code for a small task, I'd recommend something like:

```python
import string

def verify_isbn(isbn):
    handleX = lambda x: 10 if x == 'X' else x
    nums = [int(handleX(x)) for x in isbn if x in string.digits+'X']
    return(sum([x*(10-ind) for ind, x in enumerate(nums)]) % 11 == 0)
```

exercises/isbn-verifier/isbn_verifier_test.py
```diff
@@ -0,0 +1,51 @@
...
+        self.assertIs(verify('359821507X'), True)
+
+    def test_invalid_isbn_without_check_digit_and_dashes(self):
+        self.assertIs(verify('359821507'), False)
+
+    def test_invalid_too_long_isbn_with_no_dashes(self):
+        self.assertIs(verify('3598215078X'), False)
+
+    def test_invalid_isbn_without_check_digit(self):
+        self.assertIs(verify('3-598-21507'), False)
```
jamesmcm: Here you can use AssertTrue and AssertFalse instead of AssertIs.

I.e.:
```python
    def test_verify_isbn_example_with_X_spaces(self):
        self.assertTrue(isbn_verifier.verify_isbn('955 20 3051 X'))

```

config.json
```diff
@@ -902,6 +902,21 @@
       ]
     },
     {
+      "uuid": "7961c852-c87a-44b0-b152-efea3ac8555c",
+      "slug": "isbn-verifier",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 3,
```
jamesmcm: I would consider this a much, much easier difficulty. Perhaps put it after the RNA Transcription exercise?

exercises/isbn-verifier/isbn_verifier_test.py
m-a-ge: `assertFalse` won't fail the test if the tested method returns `None` in comparison with `assertIs`

Consider this:
```python
import unittest

tc = unittest.TestCase()

tc.assertFalse(None) # passes
tc.assertIs(None, True) # fails as expected
```

Do you see the difference, @jamesmcm?

So, I think it's better to use `assertIs` even if readability suffers a bit

jamesmcm: That's a good point, seems like strange behaviour from the unittest module.

m-a-ge: Yeah, especially considering that `bool(None)` returns `False` meaning that `None` is Falsy

N-Parsons: @m-a-ge, your example code is asserting that `None` is Falsey, so passing is the expected behaviour - `bool(None) is False`.

```
import unittest

tc = unittest.TestCase()

tc.assertFalse(None)  # Passes, since None is Falsey
tc.assertTrue(None)   # Fails because None is not Truthy

tc.assertIs(None, False)  # Fails because None is not False
tc.assertIs(None, True)   # Fails because None is not True
```

In theory, there's nothing wrong with using `assertTrue` and `assertFalse` in exercise tests, but the presence of the latter will cause some tests to pass when no code is implemented, and so these could confuse some people. Using `assertIs` creates another slight issue, in that it forces learners to return booleans, even though they may expect to be able to use Truthy and Falsey statements instead. I'd consider `assertIs` the lesser of two evils.

N-Parsons:
@jamesmcm, the implementation of the two additional tests is fairly simple, so I don't think they should go into a separate exercise.

**ISBN-10 to ISBN-13: ** Prepend "978" and calculate the new checksum (by the ISBN-13 method).
**ISBN-10 generator: ** Calculate the checksum, if the ISBN is valid, return it, otherwise correct the last digit and return it.

These tests are clearly quite easy extensions if the checksum calculation is implemented as a separate function to the validation.

exercises/isbn-verifier/isbn_verifier_test.py
m-a-ge: @N-Parsons thanks for clarification

config.json
```diff
@@ -1112,7 +1127,12 @@
       "unlocked_by": null,
       "difficulty": 3,
       "topics": [
-        "exception_handling"
+        "type_conversion",
+        "exception_handling",
+        "strings",
+        "arrays",
+        "integers",
+        "parsing"
```
N-Parsons: Looks like you've accidentally included some changes for a different exercise.

exercises/isbn-verifier/README.md
```diff
@@ -0,0 +1,43 @@
...
+The first 9 digits in the ISBN have to be between 0 and 9.
+The check digit can additionally be an 'X' to allow 10 to be a valid check digit as well.
+
+A valid ISBN-10 is calculated with this formula `(x1 * 10 + x2 * 9 + x3 * 8 + x4 * 7 + x5 * 6 + x6 * 5 + x7 * 4 + x8 * 3 + x9 * 2 + x10 * 1) mod 11 == 0`
+So for our example ISBN this means:
+(3 * 10 + 5 * 9 + 9 * 8 + 8 * 7 + 2 * 6 + 1 * 5 + 5 * 4 + 0 * 3 + 8 * 2 + 8 * 1) mod 11 = 0
+
+Which proves that the ISBN is valid.
+
+### Submitting Exercises
```
N-Parsons: This should be at heading level 2 (`## Submitting Exercises`).

exercises/isbn-verifier/isbn_verifier.py
```diff
@@ -0,0 +1,10 @@
+def verify(isbn):
+    pass
+
+
+def isbn_generator(isbn):
+    pass
+
+
+def isbn13_generator_from_isbn10(isbn10):
```
N-Parsons: `generate_isbn13_from_isbn10` might be a better function name?

exercises/isbn-verifier/isbn_verifier.py
```diff
@@ -0,0 +1,10 @@
+def verify(isbn):
+    pass
+
+
+def isbn_generator(isbn):
```
N-Parsons: `generate_isbn` might be a better function name? I think function names should be more descriptive of what they do than what they are.

exercises/isbn-verifier/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+The first 9 digits in the ISBN have to be between 0 and 9.
+The check digit can additionally be an 'X' to allow 10 to be a valid check digit as well.
+
+A valid ISBN-10 is calculated with this formula `(x1 * 10 + x2 * 9 + x3 * 8 + x4 * 7 + x5 * 6 + x6 * 5 + x7 * 4 + x8 * 3 + x9 * 2 + x10 * 1) mod 11 == 0`
+So for our example ISBN this means:
+(3 * 10 + 5 * 9 + 9 * 8 + 8 * 7 + 2 * 6 + 1 * 5 + 5 * 4 + 0 * 3 + 8 * 2 + 8 * 1) mod 11 = 0
+
+Which proves that the ISBN is valid.
+
```
N-Parsons: Could you add something here (and to `.meta/hints.md`) to explain that we are also expecting implementations of an ISBN10 generator and an ISBN13 generator? It will also need to explain how the conversion from ISBN10 to ISBN13 works, and how to do the ISBN13 checksum calculation.

cmccandless:
@eulercamposbarros Canonical-data has been updated in [/exercism/problem-specifications/#993](https://github.com/exercism/problem-specifications/pull/993)

pheanex:
Would it be okay for everyone If I continue from where @jamesmcm left off?
I'd really like to see this exercise online.

cmccandless:
@pheanex Actually the PR was started by @eulercamposbarros. In any case, as this has been closed due to being abandoned, the original author has effectively forfeited his/her claim on the work, so feel free to pick it up!

Issue #745, opened 2017-10-06
hangman: implement exercise


m-a-ge:
There are exercises that are not yet implemented in Python track (the full list of them - https://github.com/exercism/python/issues/417#issuecomment-334841778).
You can help us change this situation by implementing this one. Ref #417

The guide ["Implement an exercise from specification"](https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md) will help you succeed.

If you still have questions, I encourage you to ask them before starting working on the issue.

I also suggest claiming the issue by creating an empty PR as stated [here](https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md#avoiding-duplicate-work)

Please note that there are some special ['closing issue' keywords](https://help.github.com/articles/closing-issues-using-keywords/) that you can put in description of your PR and Github will take care of closing the corresponding issue automatically after merge.
This simple thing helps us a lot, and makes our life a bit easier

vaibhavsingh97:
can i take up this issue?

m-a-ge:
@vaibhavsingh97 of course

I suggest creating an empty [WIP] PR - https://github.com/exercism/docs/blob/master/you-can-help/implement-an-exercise-from-specification.md#avoiding-duplicate-work to claim it

khoa102:
Hi, I wonder if I can take up this issues as my first contribution for open source?

cmccandless:
@khoa102 Absolutely! Issues can be claimed by creating a PR for them. Please remember to put `Fixes #745` in the PR description to link the PR to this issue.

khoa102:
Great! @cmccandless

khoa102:
Hi, can anyone take a look at the PR for this issue that I submit a few day ago? I would also want to know what should I put for the config of this exercise, like the difficulty, unlock or topic?

anshima1:
Hey is this issue solved? If not, I'll like to work on it.

cmccandless:
@anshima1 No, this issue is not yet resolved.

If you wish to work on this, I highly suggest reviewing the discussion in #1160.

junming403:
Hi is this issue still relevant? I would like to work on it.

cmccandless:
@junming403 It is still relevant, yes. 

Also, as stated above:

> If you wish to work on this, I highly suggest reviewing the discussion in #1160.

junming403:
[...]

PR #905, opened 2017-10-15
[WIP] Implement hangaman excercise

vaibhavsingh97:
Fixes: #745

### DIFF
--- a/exercises/hangman/README.md
+++ b/exercises/hangman/README.md
@@ -1,0 +1,1 @@
+# Hangman
--- a/exercises/hangman/example.py
+++ b/exercises/hangman/example.py
@@ -1,0 +1,60 @@
+import random
+
+
+def main():
+    while True:
+        words = ["python", "golang", "java", "assembly", "haskel", "emoji"]
+        chosen_word = random.choice(words).lower()
+        player_guess = None
+        guessed_letters = []
+        word_guessed = []
+        for letter in chosen_word:
+            word_guessed.append("*")
+        joined_word = None
+        chances = 10
+        attempts = chances - 1
+        while (attempts != 0 and "-" in word_guessed):
+            print(("\nYou have {} attempts remaining").format(attempts))
+            joined_word = "".join(word_guessed)
+            print(joined_word)
+
+            try:
+                player_guess = str(
+                    input("\nPlease select a letter between A-Z" + "\n> ")).lower()
+            except:
+                print("please enter valid input and try again")
+                continue
+            else:
+                if not player_guess.isalpha():
+                    print("That is not a letter. Please try again.")
+                    continue
+                elif len(player_guess) > 1:
+                    print("That is more than one letter. Please try again.")
+                    continue
+                elif player_guess in guessed_letters:
+                    print("You have already guessed that letter. Please try again.")
+                    continue
+
+            guessed_letters.append(player_guess)
+
+            for pos, letter in enumerate(chosen_word):
+                if player_guess == letter:
+                    word_guessed[pos] = player_guess
+
+            if player_guess not in chosen_word:
+                attempts -= 1
+
+        if "-" not in word_guessed:
+            print(("\nCongratulations! {} was the word").format(chosen_word))
+        else:
+            print(("\nUnlucky! The word was {}.").format(chosen_word))
+
+        print("\nWould you like to play again?")
+
+        response = input("> ").lower()
+        if response not in ("yes", "y"):
+            break
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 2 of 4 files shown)


vaibhavsingh97:
@m-a-ge I don't know much about writing test. Can you please guide me 😅

exercises/hangman/hangman_test.py
```diff
@@ -0,0 +1,11 @@
+import unittest
+import hangman
+
+
+class HangmanTest(unittest.TestCase):
+    def test(self):
+        self.assertEqual("hangman", "hangman")
+
```
m-a-ge: More tests should be added

exercises/hangman/example.py
```diff
@@ -0,0 +1,60 @@
+import random
+
+
+def main():
+    while True:
```
m-a-ge: I don't like this infinite loop

exercises/hangman/README.md
```diff
@@ -0,0 +1 @@
+# Hangman
```
m-a-ge: You can adopt `README.md` from here - https://github.com/exercism/problem-specifications/blob/master/exercises/hangman/description.md

m-a-ge:
@vaibhavsingh97, it's a best practice to write tests before working on actual implementation. It gives an opportunity to analyze the problem better, find edge cases etc.
We use `unittest` for writing tests, docs can be found here - https://docs.python.org/3/library/unittest.html 
You can also take a look at tests from different exercises as examples. 

You can also adopt tests from other tracks

exercises/hangman/example.py
vaibhavsingh97: @m-a-ge This loop is only if player wan't to play again

exercises/hangman/hangman_test.py
N-Parsons: @vaibhavsingh97, I would suggest basing the tests on the ones from the [C# track](https://github.com/exercism/csharp/blob/master/exercises/hangman/HangmanTest.cs) since they have fairly good test coverage.

N-Parsons:
The [problem-specifications//description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/hangman/description.md) suggests that this should be a functional reactive programming  (FRP) exercise using an external library. Such libraries seem to exist for Python, but I know nothing about them and I don't really understand what FRP really is. Unless you know what this is and can explain and implement it, I'm going to suggest that we just modify the exercise to use a more traditional paradigm. @m-a-ge, what are your thoughts on this?

Your current implementation is going to be really hard to write tests for, because it relies on user interaction, which can't be automated through `unittest` (as far as I know). A better approach might be to implement a `Hangman` class with methods for setting/picking a word and guessing letters, and write the tests based on this approach. It might be a cool extension to include something in the README about getting people to use their new class to write an interactive game as you've done - we can't test it, but people might find it quite fun.

It would be good to put whatever new description (based loosely on the `problem-specifications` description) into a `.meta/description.md` file in this exercise folder so that your new description is persistent when the `configlet` is used to regenerate the READMEs.

N-Parsons:
@vaibhavsingh97 

The [problem-specifications//description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/hangman/description.md) suggests that this should be a functional reactive programming  (FRP) exercise using an external library. Such libraries seem to exist for Python, but I know nothing about them and I don't really understand what FRP really is. Unless you know what this is and can explain and implement it, I'm going to suggest that we just modify the exercise to use a more traditional paradigm. @m-a-ge, what are your thoughts on this?

Your current implementation is going to be really hard to write tests for, because it relies on user interaction, which can't be automated through `unittest` (as far as I know). A better approach might be to implement a `Hangman` class with methods for setting/picking a word and guessing letters, and write the tests based on this approach. It might be a cool extension to include something in the README about getting people to use their new class to write an interactive game as you've done - we can't test it, but people might find it quite fun.

It would be good to put whatever new description (based loosely on the `problem-specifications` description) into a `.meta/description.md` file in this exercise folder so that your new description is persistent when the `configlet` is used to regenerate the READMEs.

m-a-ge:
Personally, I'm in favor of adding a new exercise even if it contradicts `problem-specification` in some minor way.
It'll allow users to work on it, and we can always rewrite the example solutions to meet certain standards later (when we have time) by putting it to our backlog. 
Such exercises shouldn't be included to `core` exercises though

@N-Parsons what do you think of such approach to the problems likes these?

vaibhavsingh97:
@m-a-ge What should I do know?

m-a-ge:
@vaibhavsingh97 we need more tests to be added

N-Parsons:
@m-a-ge, sounds good to me.

cmccandless:
@vaibhavsingh97 are you still working on this?

vaibhavsingh97:
Yes, I forgot now will update PR

PR #944, opened 2017-10-17
Add topics for secret-handshake exercise

### LINKED ISSUE #849: secret-handshake: add topics

rtfoley:
Add topics for secret-handshake exercise.

Fixes #849

### DIFF
--- a/config.json
+++ b/config.json
@@ -566,7 +566,15 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-
+        "conditionals",
+        "loops",
+        "arrays",
+        "integers"
+        "enumerations",
+        "bitwise_operations",
+        "logic",
+        "mathematics",
+        "transforming"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -568,9 +568,8 @@
       "topics": [
         "conditionals",
         "loops",
-        "arrays",
+        "lists",
         "integers",
-        "enumerations",
         "bitwise_operations",
         "logic",
         "mathematics",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -566,7 +566,15 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-
+        "conditionals",
+        "loops",
+        "arrays",
```
m-a-ge: Could you please change it to `lists`.
In my mind `arrays` can store data only of the same type (in Java for example), but `lists` can store data of any type (like in Python or any implementation of `List` interface in Java)

config.json
```diff
@@ -566,7 +566,15 @@
...
+        "integers",
+        "enumerations",
```
m-a-ge: This is controversial, since `Enum` class was introduced in Python 3 - https://docs.python.org/3/library/enum.html
But I believe it's Ok to have it (to show that it's possible to use it)

config.json
rtfoley: Good point, fixed in next commit.

config.json
rtfoley: That is a good point, someone learning on Python 2.x wouldn't have access to enums without a 3rd party package. I removed it in the next commit, for consistency with the rest of the exercises (only one of which lists enumerations as a topic).

m-a-ge: @rtfoley probably it slipped through somehow, we should revisit it later

m-a-ge:
Thanks, @rtfoley

PR #945, opened 2017-10-17
Improve exception tests

### LINKED ISSUE #477: Improve exception tests
Maybe we should change the way we test exceptions (see #476) to always use a `context manger` as this way is easier to understand.

```python
def test_it(self):
    self.assertRaises(SomeException, do_something, 42, True)

def test_it_with_context(self):
    with self.assertRaises(SomeException):
        do_something(42, True)
```

## Steps
- [ ]  update old tests
- [ ] document this in the readme for new exercises

cmccandless:
Resolves #477


m-a-ge:
Thanks a lot, @cmccandless 👍

PR #962, opened 2017-10-19
book-store: use book price const in calculation

ferhatelmas:


m-a-ge:
Thanks for fixing @ferhatelmas

PR #963, opened 2017-10-20
Inorder program

ironmanMA:


N-Parsons:
It looks like a similar exercise already exists in the `problem-specifications` and is implemented in PR #773.

For future reference, new exercises are generally proposed at [exercism/problem-specifications](https://github.com/exercism/problem-specifications) first, then implemented within the different language tracks (although this could be done simultaneously).

ironmanMA:
got, will propose first and then pickup.

cmccandless:
@N-Parsons if it is agreed that this concept is already implemented in #773 can we close this PR?

N-Parsons:
I'm going to close this PR because the concept is already implemented in #773. If you would like to propose a new exercise, please do so at [exercism/problem-specifications](https://github.com/exercism/problem-specifications), but I don't think that this exercise would be accepted, since it is so similar to another.

PR #965, opened 2017-10-20
Create webcam.py

AlphaBAT69:


m-a-ge:
Hi @AlphaBAT69,

What issue does this PR solves? Please add description or summary explaining why we should add it to the project.

JFYI: there is a special repo which declares exercises for tracks - https://github.com/exercism/problem-specifications
If you have a new idea for the exercise, please add it there first

PR #967, opened 2017-10-21
Hints on cryptography about "random" modules

### LINKED ISSUE #774: Cryptographic exercises should highlight the existence of the `secrets` module
Python 3.6 introduced the `secrets` module, which provides cryptographically stronger randomness than the `random` module. Since this is only available in Python 3.6+, `random` needs to be used in the example code in order maintain backwards compatibility.

I think it's important that any exercise involving cryptography should highlight the existence of the `secrets` module in Python 3.6+ and explain why it is highly preferable to `random` for creating secure cryptography.

During a discussion on PR #756, @m-a-ge suggested that this be raised as an issue, and that a HINTS.md drafted for cryptographic exercises. The suggested wording for the HINTS.md is given below and is based on the wording submitted by @kusti8.

**Suggested wording:**
> Python, as of version 3.6, includes two different random modules. The module called `random` is pseudo-random, meaning it does not generate true randomness, but follows and algorithm that simulates randomness. Since random numbers are generated through a known algorithm, they are not truly random. The `random` module is not correctly suited for cryptography and should not be used, because it is pseudo-random. In version 3.6, Python introduced the secrets module, which generates cryptographically strong random numbers that provide the greater security required for cryptography. Since this is only an exercise, `random` is fine to use, but note that it would be very insecure if actually used for cryptography.

lucasloisp:
Closes #774 by creating a HINTS.md file for every exercise involving cryptography or the use of random module.

The wording is as follows

> 
> # Hints
> 
> ## Should I use random or secrets?
> 
> Python, as of version 3.6, includes two different random modules.
> 
> The module called `random` is pseudo-random, meaning it does not generate
> true randomness, but follows an algorithm that simulates randomness. 
> Since random numbers are generated through a known algorithm, they are not truly random. 
> 
> The `random` module is not correctly suited for cryptography and should not be used, 
> precisely because it is pseudo-random. 
> 
> For this reason, in version 3.6, Python introduced the `secrets` module, which generates
> cryptographically strong random numbers that provide the greater security required for cryptography. 
> 
> Since this is only an exercise, `random` is fine to use, but note that **it would be 
> very insecure if actually used for cryptography.**

### DIFF
--- a/exercises/diffie-hellman/.meta/hints.md
+++ b/exercises/diffie-hellman/.meta/hints.md
@@ -1,0 +1,16 @@
+## Should I use random or secrets?
+
+Python, as of version 3.6, includes two different random modules.
+
+The module called `random` is pseudo-random, meaning it does not generate
+true randomness, but follows an algorithm that simulates randomness.
+Since random numbers are generated through a known algorithm, they are not truly random.
+
+The `random` module is not correctly suited for cryptography and should not be used,
+precisely because it is pseudo-random.
+
+For this reason, in version 3.6, Python introduced the `secrets` module, which generates
+cryptographically strong random numbers that provide the greater security required for cryptography.
+
+Since this is only an exercise, `random` is fine to use, but note that **it would be
+very insecure if actually used for cryptography.**
--- a/exercises/diffie-hellman/README.md
+++ b/exercises/diffie-hellman/README.md
@@ -37,9 +37,22 @@ Bob calculates
 The calculations produce the same result!  Alice and Bob now share
 secret s.
 
-## Notes
+## Should I use random or secrets?
 
-Python, as of version 3.6, includes two different random modules. The module called `random` is pseudo-random, meaning it does not generate true randomness, but follows and algorithm that simulates randomness. Since random numbers are generated through a known algorithm, they are not truly random. The `random` module is not correctly suited for cryptography and should not be used, because it is pseudo-random. In version 3.6, Python introduced the `secrets` module, which generates cryptographically strong random numbers that provide the greater security required for cryptography. Since this is only an exercise, `random` is fine to use, but note that it would be very insecure if actually used for cryptography.
+Python, as of version 3.6, includes two different random modules.
+
+The module called `random` is pseudo-random, meaning it does not generate
+true randomness, but follows an algorithm that simulates randomness.
+Since random numbers are generated through a known algorithm, they are not truly random.
+
+The `random` module is not correctly suited for cryptography and should not be used,
+precisely because it is pseudo-random.
+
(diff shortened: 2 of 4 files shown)


N-Parsons:
@lucasloisp, I believe that the hints file should be located at `.meta/hints.md` in order for the `configlet` to find it and incorporate it into the README - you can find details for this in [the configlet's README](https://github.com/exercism/configlet/blob/master/README.md).

Some of these exercises don't seem to require randomness, specifically: 
- atbash-cipher
- crypto-square
- rail-fence-cipher
- rotational-cipher

I also wouldn't consider `robot-name` to be a cryptographic exercise - there are times where `random` is the most suitable module.

lucasloisp:
Noted! I've made the necessary updates

exercises/diffie-hellman/.meta/HINTS.md
```diff
@@ -0,0 +1,18 @@
+# Hints
```
m-a-ge: Could you please rename the file to `hints.md` (lowercase) for consistency

lucasloisp: Done! I'd done it but apparently there was some error in renaming

exercises/diffie-hellman/.meta/hints.md
```diff
@@ -0,0 +1,18 @@
+# Hints
```
N-Parsons: Heading level 1 here is going to be a the wrong level for the README - the first level in this document should be level 2 (`##`). Personally, I would just delete this heading and keep the one below - two headings for one thing would seem strange in the flow of the README.

exercises/diffie-hellman/README.md
```diff
@@ -37,10 +37,6 @@ Bob calculates
 The calculations produce the same result!  Alice and Bob now share
 secret s.
 
-## Notes
```
N-Parsons: This README (and the one for `simple-cipher`) will need either regenerating to include the hint. Alternatively, you can just copy-and-paste it in at this location.

exercises/simple-cipher/.meta/hints.md
```diff
@@ -0,0 +1,18 @@
+# Hints
```
N-Parsons: As above, I think like line should just be removed.

exercises/diffie-hellman/README.md
lucasloisp: Ready

N-Parsons: @lucasloisp, the section needs to go before `### Submitting Exercises`, since that's where it goes if we regenerate the file.

PR #968, opened 2017-10-22
Update example.py

TudorMaiereanu:


cmccandless:
@TudorMaiereanu thanks for your pull request! Can you please indicate which test this new method solves?

N-Parsons:
I'm going to close this PR because it cannot be merged.

If you think that adding an `intersect` function is beneficial, then I would suggest submitting a PR in [exercism/problem-specifications](https://github.com/exercism/problem-specifications) first, and then returning to implement it here.

PR #970, opened 2017-10-23
Normalize the config file to minimize noise later

kytrinyx:
We will be scripting a number of changes to the track configuration.
This runs configlet fmt, which ensures that any further changes can be minimized.

See https://github.com/exercism/meta/issues/95

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,577 +1,568 @@
 {
-  "language": "Python",
   "active": true,
   "exercises": [
     {
-      "uuid": "f458c48a-4a05-4809-9168-8edd55179349",
+      "core": false,
+      "difficulty": 1,
       "slug": "hello-world",
-      "core": false,
-      "unlocked_by": null,
-      "difficulty": 1,
       "topics": [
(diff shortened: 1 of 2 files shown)


lucifer-angel:
update

N-Parsons:
I've had a look at the changes, and I think that these changes harm human-readability too much and should be re-evaluated. I'm not too bothered about the changes to the `maintainers.json`, since that file is significantly smaller and is interacted with far less than the `config.json`, but I would still prefer to maintain the current key ordering.

m-a-ge:
@cmccandless for me its format is fine, it isn't typically read by most people (other than maintainers).
I believe it's very important to have consistency and I don't think it's worth changing anything in how it's formatted

kytrinyx:
@N-Parsons @cmccandless @m-a-ge I agree that the readability is harmed by alphabetizing the keys. We've got a discussion going in exercism/meta#95 - please join in there / follow along. Once we get everything sorted, I'll resubmit this.

PR #979, opened 2017-10-24
Swap '2016' for '1996' to stop faulty logic from passing unit test in Leap

cruxicheiros:
[This Leap solution](http://exercism.io/submissions/7ef5b0ab93b540f79cdee9f153e0a21c) should not pass the unit test, since it checks whether `year % 100` and `year % 400` are equal in order to tell whether the year it has been passed is a leap year. This works for 2016 because `2016 % 100` and `2016 % 400` both evaluate to 16.

1996, another leap year that is divisible by 4 but not 100, does not have this property and, under that solution, would be falsely *not* classed as a leap year.

### DIFF
--- a/exercises/leap/leap_test.py
+++ b/exercises/leap/leap_test.py
@@ -10,7 +10,7 @@ class YearTest(unittest.TestCase):
         self.assertIs(is_leap_year(2015), False)
 
     def test_year_divisible_by_4_not_divisible_by_100(self):
-        self.assertIs(is_leap_year(2016), True)
+        self.assertIs(is_leap_year(1996), True)
 
     def test_year_divisible_by_100_not_divisible_by_400(self):
         self.assertIs(is_leap_year(2100), False)
--- a/exercises/leap/leap_test.py
+++ b/exercises/leap/leap_test.py
@@ -3,7 +3,7 @@ import unittest
 from leap import is_leap_year
 
 
-# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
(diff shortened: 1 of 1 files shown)


cmccandless:
@cruxicheiros Thanks for the pull request! Good catch; turns out this issue was already addressed recently by updating the [canonical data](https://github.com/exercism/problem-specifications/blob/2f6a515ed1b1a05477988ce18c3954e1669feda2/exercises/leap/canonical-data.json), but this track had not been updated yet. If you can modify your changes to conform to the new version of the canonical data, we should be good to go!

cruxicheiros:
@cmccandless This is actually a different corner case. The one the canonical data has been updated to fix is one where it's possible to pass all the tests by checking if the year is evenly divisible by 16. With this one, it's possible to pass all the corner tests by checking if `(year % 100) == (year % 400)` - something which is still true with 2020.

Coincidentally, 16 is also not a factor of 1996, so changing the value in the canonical data from 2020 to 1996 would address both issues.

cmccandless:
@cruxicheiros Yes, submitting a PR to problem-specifications was the correct action. Thanks! :smile:

Now that [that PR](/exercism/problem-specifications/pull/971) has been merged, please update the rest of `leap_test.py` to conform to new canonical-data version, and we should be good to go!

cruxicheiros:
@cmccandless Updated the version number. I think that's it done.

cmccandless:
@cruxicheiros Merged! Congrats and thanks for working on this!

PR #1023, opened 2017-10-25
[REF] grep

padznich:
add __doc__ for grep()

### DIFF
--- a/exercises/grep/example.py
+++ b/exercises/grep/example.py
@@ -53,9 +53,7 @@ def grep(pattern, files, flags=None):
     # Default value
     if flags is None:
         flags = ''
-    
     matched_lines = []
-
     for file_name in files:
         with open(file_name) as f:
             for line_number, line in enumerate(f.readlines(), start=1):
--- a/exercises/grep/example.py
+++ b/exercises/grep/example.py
@@ -42,7 +42,18 @@ def format_lines(matched_lines, files, flags):
     return ''.join(result)
 
 
-def grep(pattern, files, flags=''):
+def grep(pattern, files, flags=None):
+    """
+    Linux-like grep command.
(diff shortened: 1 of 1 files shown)


exercises/grep/example.py
```diff
@@ -42,9 +42,18 @@ def format_lines(matched_lines, files, flags):
...
+    """
+    Linux-like grep command.
+    :param pattern: [STRING]. Match pattern.
+    :param files: [LIST].
+    :param flags: [STRING]. Additional attributes.
+    :return: [STRING].
+    """
+    # Default value
+    if flags is None:
+        flags = ''
```
N-Parsons: What is the benefit of handling `flags` this way instead of just having a default value assigned in function arguments?

exercises/grep/example.py
```diff
@@ -42,9 +42,18 @@ def format_lines(matched_lines, files, flags):
...
 
-def grep(pattern, files, flags=''):
+def grep(pattern, files, flags=None):
+    """
+    Linux-like grep command.
+    :param pattern: [STRING]. Match pattern.
+    :param files: [LIST].
+    :param flags: [STRING]. Additional attributes.
+    :return: [STRING].
+    """
```
N-Parsons: I'm not sure how useful docstrings are in the example code - it's mostly here as a proof that the exercise can be done, and to give us something to run the test-suite against. Others might disagree with me though, so I would suggest that you create an issue if you think that it's something worth doing for all of the exercises.

m-a-ge: I don't think it's necessary to add docstrings to all example solutions. 
But my guess is that someone can learn from these example solutions.

I'm Ok to merge this PR after your suggestion regarding `flags` will be addressed

PR #1033, opened 2017-10-25
space-age: update tests to version 1.0.0

### LINKED ISSUE #1007: space-age: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `space-age` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/space-age/space_age_test.py](https://github.com/exercism/python/blob/master/exercises/space-age/space_age_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/space-age/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/space-age/canonical-data.json)

abhijitparida:
I reordered the tests according to the natural order of planets.

resolves #1007


exercises/space-age/space_age_test.py
```diff
@@ -3,49 +3,38 @@
 from space_age import SpaceAge
 
 
+# test cases adapted from `x-common//canonical-data.json` @ version: 1.0.0
```
N-Parsons: This should reference `problem-specifications`:
```
```

N-Parsons:
Thanks @abhijitparida!

PR #1034, opened 2017-10-25
pig-latin: update tests to version 1.1.0

### LINKED ISSUE #1019: pig-latin: update tests to version 1.1.0
As identified in #784, some of the exercise test-suites are out-of-date, `pig-latin` is one of them. It would therefore be really helpful for people to go through and update the test suite to match the canonical data.

```
```

**Test suite:** [exercism/python/exercises/pig-latin/pig_latin_test.py](https://github.com/exercism/python/blob/master/exercises/pig-latin/pig_latin_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/pig-latin/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/pig-latin/canonical-data.json)

kishan3:
Resolves #1019


cmccandless:
@kishan3 Thanks for your pull request! In order to merge your changes, the example solution (`example.py`) must also be updated to pass the new tests. Please make the appropriate changes.

kishan3:
@cmccandless not sure how can I do that. Can you help for it?

N-Parsons:
Thanks for your work @kishan3! It just needed a slight tweak to the regex for consonant sounds.

My solution was to treat `y` as a vowel in the simplest case (`[^aeiouy]`) and then add a test for `y` followed by a vowel (`y(?=[aeiou])`). The construct `(?=[aeiou])` does a lookahead for `[aeiou]` but is non-capturing, meaning that the vowels go into the second group. If you want to learn more about regex, I would suggest looking in the [Python docs](https://docs.python.org/3/library/re.html).

kishan3:
Thank you, sir @N-Parsons. Learned many things from you.

PR #1041, opened 2017-10-25
wordy: added test version and updated tests to match canonical

### LINKED ISSUE #1013: wordy: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `wordy` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/wordy/wordy_test.py](https://github.com/exercism/python/blob/master/exercises/wordy/wordy_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/wordy/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/wordy/canonical-data.json)

jackattack24:
Fix #1013

### DIFF
--- a/exercises/wordy/wordy_test.py
+++ b/exercises/wordy/wordy_test.py
@@ -3,54 +3,62 @@ import unittest
 from wordy import calculate
 
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
+
 class WordyTest(unittest.TestCase):
-    def test_simple_add_1(self):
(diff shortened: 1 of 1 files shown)


N-Parsons:
Looks good. Thanks for your work, @jackattack24!

PR #1043, opened 2017-10-26
robot-simulator: Updated test order to match canonical-data.json

### LINKED ISSUE #1004: robot-simulator: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `robot-simulator` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
# Tests adapted from `problem-specifications//canonical-data.json` @ v2.0.0
```

**Test suite:** [exercism/python/exercises/robot-simulator/robot_simulator_test.py](https://github.com/exercism/python/blob/master/exercises/robot-simulator/robot_simulator_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/robot-simulator/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/robot-simulator/canonical-data.json)

denislooby:
resolves #1004

### DIFF
--- a/exercises/robot-simulator/robot_simulator_test.py
+++ b/exercises/robot-simulator/robot_simulator_test.py
@@ -2,8 +2,8 @@ import unittest
 
 from robot_simulator import Robot, NORTH, EAST, SOUTH, WEST
 
-# Tests adapted from `problem-specifications//canonical-data.json` @ v2.0.0
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v2.0.0
 
 class RobotTests(unittest.TestCase):
     def test_init(self):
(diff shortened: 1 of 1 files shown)


exercises/robot-simulator/robot_simulator_test.py
```diff
@@ -2,6 +2,8 @@
 
 from robot_simulator import Robot, NORTH, EAST, SOUTH, WEST
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v2.0.0
+
```
N-Parsons: For consistency, can you move the version string down by a line? (ie. so that there are two blank lines before and one after the comment).

N-Parsons:
Thanks, @denislooby!

PR #1044, opened 2017-10-26
word-count: added test version and updated tests to match canonical 

### LINKED ISSUE #1011: word-count: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `word-count` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/word-count/word_count_test.py](https://github.com/exercism/python/blob/master/exercises/word-count/word_count_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/word-count/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/word-count/canonical-data.json)

jackattack24:
Fix #1011 

As I mentioned in issue #1045, there were two tests from the canonical data that failed with the example implementation, so I left them out.

### DIFF
--- a/exercises/word-count/word_count_test.py
+++ b/exercises/word-count/word_count_test.py
@@ -3,6 +3,8 @@ import unittest
 from word_count import word_count
 
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
+
 class WordCountTests(unittest.TestCase):
 
     def test_count_one_word(self):
@@ -23,6 +25,18 @@ class WordCountTests(unittest.TestCase):
             word_count('one fish two fish red fish blue fish')
         )
 
+    def test_cramped_lists(self):
+        self.assertEqual(
+            {'one': 1, 'two': 1, 'three': 1},
+            word_count("one,two,three")
+        )
+
+    def test_expanded_lists(self):
+        self.assertEqual(
+            {'one': 1, 'two': 1, 'three': 1},
+            word_count("one,\ntwo,\nthree")
+        )
+
     def test_ignores_punctuation(self):
         self.assertEqual(
             {'car': 1, 'carpet': 1, 'as': 1, 'java': 1, 'javascript': 1},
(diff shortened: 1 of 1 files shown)


m-a-ge:
I think it's better to add the failing tests but comment them for now and create an issue for fixing the example solution (you've already done that #1045)
What do you think, @jackattack24?

jackattack24:
I'll go ahead and add them back in and comment them out.

jackattack24:
Didn't mean for all of those other commits to be in here. Is there a way to undo that?

cmccandless:
@jackattack24 No worries, we can squash the commits when it's time to merge the PR.

cmccandless:
@m-a-ge On updating example solution with new tests: I've been telling PR authors to update the example solution as well. For the sake of consistency, should we should update the example solution here as well?

@jackattack24 Sorry, I just looked at what commits you were referring to. To fix this on your end, I would recommend the following:
```Bash
$ git checkout word_count_test
$ git remote add upstream https://github.com/exercism/python.git
$ git pull upstream master
$ git rebase -i master
```

When prompted, simply remove the commit lines for all commits you wish to remove, save file, and close. Then:
```Bash
$ git push --force
```

Normally, `git push --force` is strongly discouraged, but cases like this are why it's there.

exercises/word-count/word_count_test.py
```diff
@@ -41,6 +41,21 @@ def test_mixed_case(self):
             sorted(list(word_count('go Go GO Stop stop').values()))
         )
 
+    # def test_apostrophes(self):
+    #     self.assertEqual(
+    #         {'first': 1, "don't": 2, 'laugh': 1, 'then': 1, 'cry': 1},
+    #         word_count("First: don't laugh. Then: don't cry.")
```
N-Parsons: Can you update the order of these arguments (and the ones in the rest of the test-suite) to match the `assertEqual(actual, expected)` format?

exercises/word-count/word_count_test.py
```diff
@@ -7,81 +7,93 @@ class WordCountTests(unittest.TestCase):
...
+            word_count('testing 1 2 testing'),
+            {'testing': 2, '1': 1, '2': 1}
         )
 
     def test_mixed_case(self):
         self.assertEqual(
-            [2, 3],
-            sorted(list(word_count('go Go GO Stop stop').values()))
+            sorted(list(word_count('go Go GO Stop stop').values())),
+            [2, 3]
```
N-Parsons: Would you mind changing the format of this test to match the other ones for consistency?

I had a poke around and it looks like it was done differently so that words could be uppercase in the list (#207), but the tests added to this track since all assume that words are normalised to lowercase. I kind of like the original idea of giving learners freedom, but I think that practical applications would normalise to lowercase if they wanted a case-insensitive count, and I can't think of a clean and robust way to implement case-insensitive testing.

N-Parsons:
Perfect! Thanks, @jackattack24!

PR #1052, opened 2017-10-26
perfect-numbers: Update perfect-numbers to reflect changes to canonical data

### LINKED ISSUE #1001: perfect-numbers: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `perfect-numbers` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.1
```

**Test suite:** [exercism/python/exercises/perfect-numbers/perfect_numbers_test.py](https://github.com/exercism/python/blob/master/exercises/perfect-numbers/perfect_numbers_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/perfect-numbers/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/perfect-numbers/canonical-data.json)

emerali:
Resolves #1001

A few things to note:

- I added errors + error catching to both the example solution and the test cases (as these were specified in the canonical data)
- I refer to the range of valid inputs as "positive whole numbers" as the definition of natural numbers can either include or exclude 0. If this is accepted, I'll open an appropriate PR to update the canonical data accordingly
- I've put two blank lines below the version string as my PEP8 linter was complaining about it (it's possible my linter is not configured properly)

### DIFF
--- a/exercises/perfect-numbers/perfect_numbers_test.py
+++ b/exercises/perfect-numbers/perfect_numbers_test.py
@@ -1,42 +1,78 @@
 import unittest
 
-from perfect_numbers import is_perfect
+from perfect_numbers import classify
+
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.1
 
 
 class PerfectNumbersTest(unittest.TestCase):
+    def test_smallest_perfect_number(self):
+        self.assertIs(classify(6), "perfect")
 
-    def test_first_perfect_number(self):
-        self.assertIs(is_perfect(6), True)
+    def test_medium_perfect_number(self):
+        self.assertIs(classify(28), "perfect")
 
-    def test_no_perfect_number(self):
-        self.assertIs(is_perfect(8), False)
+    def test_large_perfect_number(self):
+        self.assertIs(classify(33550336), "perfect")
 
-    def test_second_perfect_number(self):
-        self.assertIs(is_perfect(28), True)
-
-    def test_abundant(self):
-        self.assertIs(is_perfect(20), False)
-
-    def test_answer_to_the_ultimate_question_of_life(self):
-        self.assertIs(is_perfect(42), False)
-
+    # Additional tests for this track
     def test_third_perfect_number(self):
-        self.assertIs(is_perfect(496), True)
-
-    def test_odd_abundant(self):
-        self.assertIs(is_perfect(945), False)
+        self.assertIs(classify(496), "perfect")
 
     def test_fourth_perfect_number(self):
-        self.assertIs(is_perfect(8128), True)
-
-    def test_fifth_perfect_number(self):
-        self.assertIs(is_perfect(33550336), True)
+        self.assertIs(classify(8128), "perfect")
 
     def test_sixth_perfect_number(self):
-        self.assertIs(is_perfect(8589869056), True)
+        self.assertIs(classify(8589869056), "perfect")
 
     def test_seventh_perfect_number(self):
-        self.assertIs(is_perfect(137438691328), True)
+        self.assertIs(classify(137438691328), "perfect")
+
+
+class AbundantNumbersTest(unittest.TestCase):
+    def test_smallest_abundant_number(self):
+        self.assertIs(classify(12), "abundant")
+
+    def test_medium_abundant_number(self):
+        self.assertIs(classify(30), "abundant")
+
+    def test_large_abundant_number(self):
+        self.assertIs(classify(33550335), "abundant")
+
+    # Additional tests for this track
+    def test_answer_to_the_ultimate_question_of_life(self):
+        self.assertIs(classify(42), "abundant")
+
+    def test_odd_abundant(self):
+        self.assertIs(classify(945), "abundant")
+
+    def test_even_abundant(self):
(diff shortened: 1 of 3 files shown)


exercises/perfect-numbers/perfect_numbers_test.py
```diff
@@ -1,42 +1,78 @@
...
+    def test_large_deficient_number(self):
+        self.assertIs(classify(33550337), "deficient")
+
+    def test_edge_case(self):
+        self.assertIs(classify(1), "deficient")
+
+
+class InvalidInputsTest(unittest.TestCase):
+    def test_zero(self):
+        self.assertRaises(ValueError, classify, 0)
```
cmccandless: By [convention](https://github.com/exercism/python/blob/78dfc6740fb962085d5f215e9b837e4292a20687/README.md#conventions) we use context manager syntax for `assertRaises`. Example:
```Python
with self.assertRaises(ValueError):
    classify(0)
```

Also, if the [canonical data](https://github.com/exercism/problem-specifications/blob/16f6dbcf4239099d25049a33f512ce936a14119c/exercises/perfect-numbers/canonical-data.json) specifies an error string, it is best practice to use [`assertRaisesRegexp`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp) to check that the raised error contains the correct reason.

emerali:
I've made the requested changes, however it seems like `assertRaisesRegexp` is deprecated under Python 3 as it was renamed to `assertRaisesRegex` in 3.2. 

Is this alright?

cmccandless:
@emerali I was not aware of the deprecation. In order to maintain backwards compatibility with Python 2.7, we have the following options:
- Revert to `assertRaises`
- Keep `assertRaisesRegexp`, but add a comment acknowledging deprecation and explaining why it was used anyway

@N-Parsons @m-a-ge thoughts?

m-a-ge:
@cmccandless checking the Python version looks reasonable to me

N-Parsons:
@cmccandless, I would prefer that we just use `assertRaises` - do we really care about the exact wording that someone is raising within their `ValueError`?

If we do want to continue with checking the message, then I would suggest defining a new class method that will call either `self.assertRaisesRegexp` or `self.assertRaisesRegex` dependent on version.

```python
try:
    self.assertRaisesMessage =  self.assertRaisesRegex  # Python3
except AttributeError:
    self.assertRaisesMessage = self.assertRaisesRegexp  # Python2
```

cmccandless:
@N-Parsons the only reason I advised checking the error message is that in this case the spec for this exercise actually contains specific error messages. Normally they do not. 

However, I think you raise a good point on exact wording.... Perhaps it would be best to ignore error messages entirely as a track.

emerali:
Just to add my 2 cents:
I think it'd be better to have there be an exact error message provided in the tests, as this'll give users hints on edge cases to look out for.
Secondly, I think the test cases should be as easy to read as possible, and while I love functional shenanigans, since many users are likely not super familiar with either programming or python, I think the `try-except` solution might be a bit hard to understand (though I can't really think of a decent alternative).

N-Parsons:
@cmccandless, while they do appear in the problem spec, I don't know how important the exact wording actually is (and there's some discussion about how to represent errors in exercism/problem-specifications#905 and exercism/problem-specifications#902). I'm happy with either checking or not checking error messages, but if we are going to expect a particular error message, then it should go into the README (via `.meta/hints.md`). I think it might be worth raising an issue to discuss error message checking in general, so that we can come to a track-wide decision and keep track of whatever changes might be necessary to keep exercises consistent.

@emerali, expecting exact error messages is ok, but you will need to put a note about it in `.meta/hints.md` (you'll probably need to create it) and then regenerate/manually update the README to include it. I always find it really frustrating when I have to look in the test-suite to find something like this that is required despite it not making any functional difference.

With regard to using a `try .. except` block, I think it's the cleanest solution, and it avoids redundant code. I also don't think that we should shy away from using fundamental concepts - people can learn from seeing them, and they aren't required to understand it anyway because it's part of the test-suite and not the exercise.

cmccandless:
@N-Parsons I've created a new issue to discuss convention on handling exception messages.

cmccandless:
Pending discussion in #1080

cmccandless:
@emerali Thank you for your patience on this. I'd say for now remove all checks for exception messages. If the conclusion in #1080 calls for checking them, a new pull request will be made to add them (other exercises will likely require similar changes anyway).

emerali:
@cmccandless No worries, I got a free t-shirt out of this, so I'm happy :smile:

I've removed the exception message checks (and also updated the placeholder file to remove the unnecessary skeleton function), please double check and let me know if I've missed anything.

Once the discussion in #1080 concludes, do let me know if you need some help updating some exercises, I'd be glad to help!

cmccandless:
@emerali Merging. Congrats, and thanks for working on this!

PR #1053, opened 2017-10-26
word-search: Update tests to version 1.1.0(#1012)

wilson8507:
resolve issue #1012


exercises/word-search/word_search_test.py
```diff
@@ -36,48 +137,179 @@ def test_vertical_words_top_to_bottom(self):
             self.example.search('ecmascript'),
             (Point(9, 0), Point(9, 9))
         )
+        self.assertEqual(
```
cmccandless: There are a lot of redundant and misplaced assertions. Please check that each assertion actually belongs under its test method (Example: checks for "clojure" in `self.example` should only appear in tests for horizontal words).

wilson8507:
@cmccandless I've fixed it, please check if there is any problem. Thanks!

exercises/word-search/word_search_test.py
```diff
@@ -2,9 +2,112 @@
 
 from word_search import WordSearch, Point
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.1.0
+
```
N-Parsons: Minor formatting point, but can you move the version comment down a line (ie. two blank lines before, one after).

wilson8507:
@N-Parsons I've resolved it. Thank you for pointing out my carelessness!

N-Parsons:
Perfect! Thanks, @wilson8507! I'm a bit picky about formatting and consistency :stuck_out_tongue:

PR #1061, opened 2017-10-27
kindergarten-garden: update tests to version 1.0.0

### LINKED ISSUE #997: kindergarten-garden: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `kindergarten-garden` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/kindergarten-garden/kindergarten_garden_test.py](https://github.com/exercism/python/blob/master/exercises/kindergarten-garden/kindergarten_garden_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/kindergarten-garden/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/kindergarten-garden/canonical-data.json)

smalley:
Renamed existing tests to match test case descriptions.
Deleted and added test cases to bring test list into agreement with data.
Added missing test version comment. Closes #997

### DIFF
--- a/exercises/kindergarten-garden/kindergarten_garden_test.py
+++ b/exercises/kindergarten-garden/kindergarten_garden_test.py
@@ -2,14 +2,26 @@ import unittest
 
 from kindergarten_garden import Garden
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
+
 
 class KindergartenGardenTests(unittest.TestCase):
-    def test_alices_garden(self):
+    def test_garden_with_single_student(self):
         self.assertEqual(
             Garden("RC\nGG").plants("Alice"),
             "Radishes Clover Grass Grass".split())
 
-    def test_bob_and_charlies_gardens(self):
+    def test_different_garden_with_single_student(self):
+        self.assertEqual(
+            Garden("VC\nRC").plants("Alice"),
+            "Violets Clover Radishes Clover".split())
+
+    def test_garden_with_two_students(self):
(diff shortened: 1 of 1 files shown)


exercises/kindergarten-garden/kindergarten_garden_test.py
```diff
@@ -26,16 +38,6 @@ def test_full_garden(self):
...
-    def test_disordered_test(self):
-        garden = Garden(
-            "VCRRGVRG\nRVGCCGCV",
-            students="Samantha Patricia Xander Roger".split())
-        self.assertEqual(
-            garden.plants("Patricia"),
-            "Violets Clover Radishes Violets".split())
-        self.assertEqual(
-            garden.plants("Xander"), "Radishes Grass Clover Violets".split())
-
```
cmccandless: I don't think there's a problem with having track-specific tests (as long as they do not conflict with canonical tests); there are other exercises in which we do this. I think we ought to keep this test, but add a comment that this is a track-specific test and is not canonical.

smalley:
I think I may have messed with history on that commit a little trying to clean up that branch, but the requested change should now be included. The previous test was added back with a comment indicating it is an additional test for that track.

cmccandless:
@smalley Merged. Congrats and thanks for working on this!

PR #1062, opened 2017-10-27
minesweeper: update tests to version 1.1.0

### LINKED ISSUE #998: minesweeper: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `minesweeper` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/minesweeper/minesweeper_test.py](https://github.com/exercism/python/blob/master/exercises/minesweeper/minesweeper_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/minesweeper/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/minesweeper/canonical-data.json)

smalley:
Add missing test version comment.
Add ordered tests with names matching canonical data descriptions.
Disable empty board tests which are incompatible with board class.
Move selected old tests to additional tests section. Closes #998

### DIFF
--- a/exercises/minesweeper/example.py
+++ b/exercises/minesweeper/example.py
@@ -1,4 +1,6 @@
 def board(inp):
+    if(inp == []):
+        return []
     verify_board(inp)
     rowlen = len(inp[0])
     collen = len(inp)
@@ -7,9 +9,13 @@ def board(inp):
         for i2 in range(rowlen):
             if b[i1][i2] != ' ':
                 continue
-            cnt = inp[i1 - 1][i2 - 1:i2 + 2].count('*') + \
-                inp[i1][i2 - 1:i2 + 2].count('*') + \
-                inp[i1 + 1][i2 - 1:i2 + 2].count('*')
+            low = max(i2 - 1, 0)
+            high = min(i2 + 2, rowlen + 2)
+            cnt = inp[i1][low:high].count('*')
+            if(i1 > 0):
+                cnt += inp[i1 - 1][low:high].count('*')
+            if(i1 < collen - 1):
+                cnt += inp[i1 + 1][low:high].count('*')
             if cnt == 0:
                 continue
             b[i1][i2] = str(cnt)
@@ -17,23 +23,13 @@ def board(inp):
 
 
 def verify_board(inp):
-    # Null board or a null row
-    if not inp or not all(r for r in inp):
-        raise ValueError("Invalid board")
     # Rows with different lengths
     rowlen = len(inp[0])
-    collen = len(inp)
     if not all(len(r) == rowlen for r in inp):
         raise ValueError("Invalid board")
     # Unknown character in board
     cset = set()
     for r in inp:
         cset.update(r)
-    if cset - set('+- *|'):
-        raise ValueError("Invalid board")
-    # Borders not as expected
-    if any(inp[i1] != '+' + '-' * (rowlen - 2) + '+'
-           for i1 in [0, -1]) or any(inp[i1][i2] != '|'
-                                     for i1 in range(1, collen - 1)
-                                     for i2 in [0, -1]):
+    if cset - set(' *'):
         raise ValueError("Invalid board")
--- a/exercises/minesweeper/minesweeper_test.py
+++ b/exercises/minesweeper/minesweeper_test.py
@@ -14,193 +14,136 @@ from minesweeper import board
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
 
 class MinesweeperTest(unittest.TestCase):
-    # Tests in canonical data incompatible with board class
+
     def test_no_rows(self):
-        with self.assertRaises(ValueError):
-            board([])
+        self.assertEqual(board([]), [])
 
     def test_no_columns(self):
-        with self.assertRaises(ValueError):
-            board([""])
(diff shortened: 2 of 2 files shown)


exercises/minesweeper/minesweeper_test.py
```diff
@@ -10,114 +10,158 @@
...
-               "|112*4*|",
-               "|1*22*2|",
-               "|111111|",
-               "+------+"]
+    # Tests in canonical data incompatible with board class
+    # def test_no_rows(self):
+    #     self.assertEqual(board([]), [])
+
+    # def test_no_columns(self):
+    #     self.assertEqual(board([""]), [""])
```
cmccandless: Can you please give a more detailed explanation for why these tests are disabled? I don't like setting a precedent for skipping/disabling canonical tests.

smalley: Originally this was because the implementation of the board class uses the verify_board function to reject boards without the border or no rows/columns but the canonical data looks like it expects a return of the board as is for both of these cases.

If we'd be okay with it, I can reenable both of these cases but convert both into checks that the board class raises a ValueError in response to the board type instead.

cmccandless: I would prefer that you re-enable the disabled test cases, but instead of converting the tests, can we simply do away with the `verify_board` method? There is no requirement in either the [canonical data](https://github.com/exercism/problem-specifications/blob/7735764b0882360563772aa66a2b1fd77c5bb568/exercises/minesweeper/canonical-data.json) or the [description](https://github.com/exercism/problem-specifications/blob/7735764b0882360563772aa66a2b1fd77c5bb568/exercises/minesweeper/description.md) that the solution needs to validate the board; it is merely meant to convert it.

smalley: We could remove the verify_board method from the example implementation, but I think it would necessitate removing a few of the additional test cases as well (the final 3 that test malformed boards). 

I think it may be desirable to have a verify_board method though. If there's no restriction on the input being a well formed board especially cases with rows/columns of differing sizes it may be hard to decide what's truly adjacent to a mine thus complicating generating output.

It also seems like the no row/no column board  with no border might not be valid given that all  the boards in the problem description and old test set always include a drawn border.

N-Parsons: Personally, these don't look like valid boards to me, given the borders present in all other examples, so I would expect these to raise `ValueError`s. Ultimately, I don't mind whether these tests are modified or removed - I don't think that they add anything to what we already have.

@cmccandless, I like the inclusion of board verification, and I don't think that it makes sense to remove it in order to better match the canonical data for two trivial cases.

exercises/minesweeper/minesweeper_test.py
```diff
@@ -10,114 +10,158 @@
 
 from minesweeper import board
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
+
```
N-Parsons: @smalley, can you move the version comment down by a line for consistency? (ie. two blank lines before, one after)

N-Parsons:
Having looked at this again, I noticed that the canonical data doesn't have the borders. @cmccandless, what are your thoughts on removing the borders?

exercises/minesweeper/minesweeper_test.py
smalley: Sure, no problem at all @N-Parsons, I can make that change and push an update.

smalley:
I updated this pull request to convert the first two tests to expect ValueError to be raised and uncommented them. If we want to remove the border or convert the no row/ no column tests to have borders we can definitely do those as well.

cmccandless:
@N-Parsons if the canonical data does not have borders, then I don't think the tests should have them. If this change were made, do we have any reason to keep the board validation?

N-Parsons:
Ok, I agree that we should remove the borders. I think that without the borders, the validation becomes too trivial to really worry about making people implement, so I think that it can be removed as well.

@smalley, are you happy to make these changes?

smalley:
@N-Parsons sure, I can work on making those changes to the boards and remove the validation function

smalley:
@N-Parsons 
I made an update that removes the borders but I left the verification function in to catch the few remaining totally invalid board cases (e.g. row lengths varying between rows).  I also tweaked the example.py case to properly return the totally empty case and account for the fact that indexes in some operations can't count on the border being around anymore.

exercises/minesweeper/example.py
```diff
@@ -31,9 +33,3 @@ def verify_board(inp):
         cset.update(r)
     if cset - set('+- *|'):
```
N-Parsons: The second set should be `set(" *")` now, since the other elements were for the border.

N-Parsons:
Thanks @smalley, I've left a comment about a very minor issue the example solution (the set of valid characters needs updating). Once that's fixed, I think this will be ready to merge.

@cmccandless, can you review this again?

N-Parsons:
Looks good. Thanks, @smalley!

PR #1063, opened 2017-10-27
forth: Fix errors in tests

N-Parsons:
I noticed that there was an error in one of the tests, and that the simple parsing/stack tests from the `canonical-data` were not implemented. This PR fixes these issues, as well as applying the convention of `assertEqual(actual, expected)`.

### DIFF
--- a/exercises/forth/example.py
+++ b/exercises/forth/example.py
@@ -11,6 +11,8 @@ def is_integer(string):
 
 
 def evaluate(input_data):
+    if not input_data:
+        return []
     defines = {}
     while input_data[0][:1] == ':':
         values = input_data.pop(0).split()
--- a/exercises/forth/example.py
+++ b/exercises/forth/example.py
@@ -12,7 +12,7 @@ def is_integer(string):
 
 def evaluate(input_data):
     defines = {}
-    while input_data[0][0] == ':':
+    while input_data[0][:1] == ':':
         values = input_data.pop(0).split()
         values.pop()
         values.pop(0)
--- a/exercises/forth/forth_test.py
+++ b/exercises/forth/forth_test.py
@@ -8,6 +8,11 @@ from forth import evaluate, StackUnderflowError
 
 class ForthParsingTest(unittest.TestCase):
     def test_empty_input_empty_stack(self):
+        input_data = []
+        expected = []
+        self.assertEqual(evaluate(input_data), expected)
+
+    def test_empty_line_empty_stack(self):
         input_data = [""]
         expected = []
         self.assertEqual(evaluate(input_data), expected)
--- a/exercises/forth/forth_test.py
+++ b/exercises/forth/forth_test.py
@@ -4,13 +4,25 @@ from forth import evaluate, StackUnderflowError
 
 
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.2.0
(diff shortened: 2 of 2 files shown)


exercises/forth/example.py
```diff
@@ -12,7 +12,7 @@ def is_integer(string):
 
 def evaluate(input_data):
     defines = {}
-    while input_data[0][0] == ':':
+    while input_data[0][:1] == ':':
```
cmccandless: Took me a moment to understand how this changed anything... good catch! I'm not sure I would've thought of that, probably would've done something clunky, such as:
```Python
while input_data[0] != '' and input_data[0][0] == ':':
...
```

exercises/forth/forth_test.py
```diff
@@ -4,13 +4,25 @@
 
 
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.2.0
+# Tests for case-insensitivity are track-specific
+
+class ForthParsingTest(unittest.TestCase):
+    def test_empty_input_empty_stack(self):
+        input_data = [""]
```
cmccandless: According to [problem spec](https://github.com/exercism/problem-specifications/blob/84ce49e4002468aa374c5a62a24116650fa9ce6c/exercises/forth/canonical-data.json), this should be
```Python
input_data = []
```

exercises/forth/example.py
N-Parsons: Yeah, it's a bit of a weird string-slicing trick :stuck_out_tongue:

exercises/forth/forth_test.py
N-Parsons: Thanks for pointing this out, I've added the correct test now, but I left this particular test because I think that solutions should be able to handle empty lines.

cmccandless: I agree. Perhaps this merits a PR to [problem spec](https://github.com/exercism/problem-specifications/blob/84ce49e4002468aa374c5a62a24116650fa9ce6c/exercises/forth/canonical-data.json)?

N-Parsons: Just about to make one :smile:

exercises/forth/forth_test.py
```diff
@@ -4,13 +4,30 @@
 
 
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.2.0
+# Tests for case-insensitivity are track-specific
```
cmccandless: RE: [problem description](https://github.com/exercism/problem-specifications/blob/7735764b0882360563772aa66a2b1fd77c5bb568/exercises/forth/description.md):

> Words are case-insensitive.

N-Parsons: @cmccandless, the reason for the comment was that tests for case insensitivity were not in the canonical data, even if they were in the problem description. Perhaps I could have worded it better/differently, but I'm not sure that it is worth changing now?

cmccandless:
@N-Parsons Fair point. What might be a better course of action is to add the case-insensitivity tests to the problem specification via PR, and create a new PR here to remove comment when/if that's merged.

⁣


Corey McCandless​

N-Parsons:
@cmcandless, case-insensitivity has now been proposed in exercism/problem-specifications#979. I will update the Python tests once this is merged.

PR #1065, opened 2017-10-27
Triangle: update tests to version 1.1.0

### LINKED ISSUE #1008: triangle: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `triangle` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/triangle/triangle_test.py](https://github.com/exercism/python/blob/master/exercises/triangle/triangle_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/triangle/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/triangle/canonical-data.json)

yo-ga:
Renamed existing tests to match test case descriptions.
Deleted and added test cases to bring test list into agreement with data.
Added missing test version comment. 
Closes #1008

### DIFF
--- a/exercises/triangle/triangle_test.py
+++ b/exercises/triangle/triangle_test.py
@@ -3,7 +3,8 @@ import unittest
 from triangle import Triangle, TriangleError
 
 
-# Tests adapted from `problem-specifications/exercises/triangle/canonical-data.json` @ v1.0.0
+# Tests adapted from `problem-specifications/exercises/\
+# triangle/canonical-data.json` @ v1.0.0
 
 class TriangleTests(unittest.TestCase):
     def test_equilateral_triangles(self):
@@ -32,11 +33,11 @@ class TriangleTests(unittest.TestCase):
     # Additional tests for this track
 
     def test_illegal_triangles(self):
-        self.assertEqual(Triangle(0, 0, 0).kind(), "error")
-        self.assertEqual(Triangle(3, 4, -5).kind(), "error")
-        self.assertEqual(Triangle(1, 1, 3).kind(), "error")
-        self.assertEqual(Triangle(2, 4, 2).kind(), "error")
-        self.assertEqual(Triangle(7, 3, 2).kind(), "error")
+        self.assertRaise(TriangleError, Triangle(0, 0, 0).kind())
+        self.assertRaise(TriangleError, Triangle(3, 4, -5).kind())
+        self.assertRaise(TriangleError, Triangle(1, 1, 3).kind())
+        self.assertRaise(TriangleError, Triangle(2, 4, 2).kind())
+        self.assertRaise(TriangleError, Triangle(7, 3, 2).kind())
 
     def test_larger_equilateral_triangles_also_have_equal_sides(self):
         self.assertEqual(Triangle(10, 10, 10).kind(), "equilateral")
--- a/exercises/triangle/triangle_test.py
+++ b/exercises/triangle/triangle_test.py
@@ -33,11 +33,12 @@ class TriangleTests(unittest.TestCase):
     # Additional tests for this track
 
     def test_illegal_triangles(self):
-        self.assertRaises(TriangleError, Triangle(0, 0, 0).kind())
-        self.assertRaises(TriangleError, Triangle(3, 4, -5).kind())
-        self.assertRaises(TriangleError, Triangle(1, 1, 3).kind())
-        self.assertRaises(TriangleError, Triangle(2, 4, 2).kind())
-        self.assertRaises(TriangleError, Triangle(7, 3, 2).kind())
+        with self.assertRaises(TriangleError):
+            Triangle(0, 0, 0).kind()
+            Triangle(3, 4, -5).kind()
+            Triangle(1, 1, 3).kind()
+            Triangle(2, 4, 2).kind()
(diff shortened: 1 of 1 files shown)


cmccandless:
Travis-CI build failed due to flake8 violations:
```Bash
$ flake8
./exercises/triangle/triangle_test.py:3:1: F401 'triangle.TriangleError' imported but unused
./exercises/triangle/triangle_test.py:6:80: E501 line too long (93 > 79 characters)
```

exercises/triangle/triangle_test.py
```diff
@@ -3,55 +3,43 @@
...
+        self.assertEqual(Triangle(5, 4, 6).kind(), "scalene")
+        self.assertNotEqual(Triangle(4, 4, 4).kind(), "scalene")
+        self.assertNotEqual(Triangle(4, 4, 3).kind(), "scalene")
+        self.assertNotEqual(Triangle(7, 3, 2).kind(), "scalene")
+        self.assertEqual(Triangle(0.5, 0.4, 0.6).kind(), "scalene")
+
+    # Additional tests for this track
+
+    def test_illegal_triangles(self):
+        self.assertEqual(Triangle(0, 0, 0).kind(), "error")
```
cmccandless: For illegal triangles, an error should actually be raised. I do agree that removing the need for `TriangleError` is a good idea, but I think it should just be replaced with `ValueError`.

cmccandless: I guess I can live with `TriangleError`

exercises/triangle/triangle_test.py
```diff
@@ -3,55 +3,45 @@
...
+# Tests adapted from `problem-specifications/exercises/\
+# triangle/canonical-data.json` @ v1.0.0
+
 class TriangleTests(unittest.TestCase):
-    def test_equilateral_triangles_have_equal_sides(self):
+    def test_equilateral_triangles(self):
         self.assertEqual(Triangle(2, 2, 2).kind(), "equilateral")
+        self.assertNotEqual(Triangle(2, 3, 2).kind(), "equilateral")
+        self.assertNotEqual(Triangle(5, 4, 6).kind(), "equilateral")
+        # self.assertNotEqual(Triangle(0, 0, 0).kind(), "equilateral")
```
m-a-ge: Why is this line commented out?

exercises/triangle/triangle_test.py
```diff
@@ -3,55 +3,45 @@
...
+        self.assertNotEqual(Triangle(4, 4, 3).kind(), "scalene")
+        # self.assertNotEqual(Triangle(7, 3, 2).kind(), "scalene")
+        self.assertEqual(Triangle(0.5, 0.4, 0.6).kind(), "scalene")
 
-    def test_very_small_triangles_are_legal(self):
-        self.assertEqual(Triangle(0.4, 0.6, 0.3).kind(), "scalene")
+    # Additional tests for this track
 
-    def test_triangles_with_no_size_are_illegal(self):
+    def test_illegal_triangles(self):
```
m-a-ge: Maybe we should skip track specific tests by default?

exercises/triangle/triangle_test.py
```diff
@@ -3,55 +3,45 @@
...
         with self.assertRaises(TriangleError):
-            Triangle(0, 0, 0)
+            Triangle(0, 0, 0).kind()
+            Triangle(3, 4, -5).kind()
+            Triangle(1, 1, 3).kind()
+            Triangle(2, 4, 2).kind()
+            Triangle(7, 3, 2).kind()
```
N-Parsons: This block will only test that one of these method calls raises a `TriangleError`. In order to test all of these, each should be within its own `assertRaises` block - the previous format with separate tests would be preferable.

I also think that it would be better for invalid triangles to raise an error on creation, rather than being able to exist and only cause an error when used, so calls should be to `Triangle`, not `Triangle.kind()`.

exercises/triangle/triangle_test.py
```diff
@@ -3,55 +3,45 @@
 from triangle import Triangle, TriangleError
 
 
+# Tests adapted from `problem-specifications/exercises/\
+# triangle/canonical-data.json` @ v1.0.0
```
N-Parsons: This version string should be: ```# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0```

We leave the intervening paths as being implied by the `//` to save splitting the line as you had to do.

N-Parsons:
@m-a-ge, the current implementation of this exercise uses `Triangle.kind` to return a string indicating what type of triangle it is. However, the canonical data tests whether a triangle is `isosceles`, `scalene` or `equilateral`, with each returning `True`/`False`. Do you think it would be better to update our implementation here to better match the format of the canonical data?

m-a-ge:
@N-Parsons yes, I think it'll be better to implement it as it's outlined in `canonical-data`

N-Parsons:
@yo-ga, if you're still interested in working on this, could you update the exercise implementation to better match the canonical data? (as described above)

PR #1067, opened 2017-10-27
rna-transcription: Use an exception to handle input errors

N-Parsons:
Based on discussion at exercism/problem-specifications#956, this PR changes the `rna-transcription` exercise to raise a `ValueError` instead of returning an empty string.


m-a-ge:
👍

PR #1068, opened 2017-10-27
alphametics: Update example solution to be more performance efficient 

### LINKED ISSUE #524: Alphametics reference solution
I noticed, when running test/check-exercises.py that test for alphametics take almost 220 seconds, give or take. I am not sure, maybe there is more optimal solution out there somewhere?

### LINKED ISSUE #1066: alphametics: Tests are taking a long time to complete
I was having a look at the `travis-ci` build checks, and it looks like `alphametics` is taking up a large chunk of time when running the test-suite. I feel like this is an issue.

## Run times for an [example](https://travis-ci.org/exercism/python/builds/293841177?utm_source=github_status&utm_medium=notification):
* Python 2.7: **1 min 19 s** / 1 min 59 s
* Python 3.3: **1 min 45 s** / 2 min 39 s
* Python 3.4: **1 min 15 s** / 2 min 2 s
* Python 3.5: **34 s** / 1 min 11s
* Python 3.6: **2 min 46 s** / 3 min 47 s

There's obviously quite a bit of variation in the run-time of the `alphametics` test and in the overall run-time of the checks, but it's fairly consistently taking about half of the total check time. _(Variation is presumably due to varying loads on the service)._

## Possible solutions:
1. Leave it as is - accept the additional run time for tests.
2. Add `@unittest.skip` to last test and note it in the README.
3. Refactor the example for `alphametics` to improve performance (if possible).

For me, the ideal solution would be to refactor the example code to improve the performance, but I'm not sure how feasible that actually is.

Alternatively, we could skip the test for now. On my system, skipping the last test cuts the run time down from ~35 seconds to ~5 seconds, which would seem to be worth doing. This skip could be temporary and we could leave an open issue for refactoring the example in case anyone does find a way.

What do you think?

nikamirrr:
Faster example...

Resolves #1066 // m-a-ge
Resolves #524 // m-a-ge

### DIFF
--- a/exercises/alphametics/example.py
+++ b/exercises/alphametics/example.py
@@ -32,8 +32,8 @@ def solve(an):
             for otest in permutations(dset - set(digtest), olen):
                 if any(map(lambda x: x == "0", otest[:leadotherlen])):
                     continue
-                tr = str.maketrans(combostg, "".join(digtest + otest))
-                fulleval = [[int(w.translate(tr)) for w in s] for s in fullexp]
+                alldict = dict(zip(combostg, "".join(digtest + otest)))
+                fulleval = [[int("".join([alldict[c] for c in w])) for w in s] for s in fullexp]
                 if sum(fulleval[0]) == sum(fulleval[1]):
                     return dict(zip(combostg, map(int, digtest + otest)))
     return {}
--- a/exercises/alphametics/example.py
+++ b/exercises/alphametics/example.py
@@ -32,8 +32,11 @@ def solve(an):
             for otest in permutations(dset - set(digtest), olen):
                 if any(map(lambda x: x == "0", otest[:leadotherlen])):
                     continue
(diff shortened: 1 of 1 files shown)


N-Parsons:
[alphametics: Refactor example.py](https://trello.com/c/PTOg2IDB/1-alphametics-refactor-examplepy)

nikamirrr:
Completely agree,
And I do not mind at all if someone replaced it with whatever fits.

It is on my list to fix the code.
 I am in the different timed coding challenge at the moment. But I do see few ways to improve the code. ETA next weekend.
Particularly I was quite frustrated that "maketrans" works differently in 2.7 and 3.6.

Sorry, I wish I could do it full time.

N-Parsons:
I'm going to go ahead and merge this as is.

@nikamirrr, thanks for your work. If you want to refactor this at some point feel free (just create a new PR), but don't worry if you don't have time.

PR #1071, opened 2017-10-29
ocr-numbers: Update ocr-numbers to reflect changes to canonical data

### LINKED ISSUE #999: ocr-numbers: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `ocr-numbers` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/ocr-numbers/ocr_numbers_test.py](https://github.com/exercism/python/blob/master/exercises/ocr-numbers/ocr_numbers_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/ocr-numbers/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/ocr-numbers/canonical-data.json)

emerali:
Resolves #999

### DIFF
--- a/exercises/ocr-numbers/example.py
+++ b/exercises/ocr-numbers/example.py
@@ -1,10 +1,10 @@
-ROW = 4
-COL = 3
+NUM_ROWS = 4
+NUM_COLS = 3
 
 
 def split_ocr(ocr):
-    return [[ocr[i][COL * j:COL * (j + 1)] for i in range(ROW)]
-            for j in range(len(ocr[0]) // COL)]
+    return [[ocr[i][NUM_COLS * j:NUM_COLS * (j + 1)] for i in range(NUM_ROWS)]
+            for j in range(len(ocr[0]) // NUM_COLS)]
 
 
 ALL = ['    _  _     _  _  _  _  _  _ ',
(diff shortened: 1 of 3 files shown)


emerali:
Accidentally committed changes for another PR as well, will fix in a sec

exercises/ocr-numbers/example.py
```diff
@@ -16,28 +16,25 @@ def split_ocr(ocr):
 OCR_LIST = [OCR_LIST[-1]] + OCR_LIST[:9]
 
 
-def number(ocr):
-    if (len(ocr) != ROW or len(ocr[0]) % COL or
-            any(len(r) != len(ocr[0]) for r in ocr)):
+def convert(input_grid):
```
m-a-ge: Exercise placeholder should be also updated in this case:
https://github.com/exercism/python/blob/master/exercises/ocr-numbers/ocr_numbers.py

emerali:
@m-a-ge I've made the requested change.

m-a-ge:
Merged, thanks @emerali!!

PR #1077, opened 2017-10-30
update pig-latin test case

capacitate:
Issue #1019 

follow the order from canonical data


N-Parsons:
Looks like we already have an active PR for this: #1034.

N-Parsons:
Closing since this issue was resolved in #1034

PR #1079, opened 2017-10-30
Add topics for difference-of-squares

### LINKED ISSUE #814: difference-of-squares: add topics

rahulagrawal97:
fixes #814 
This is my first PR at exercism!
please add the hactoberfest label too.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1117,6 +1117,19 @@
       ]
     },
     {
+        "uuid": "55c818d6-04a8-2480-de83-9f85fd0d2eafa71935f",
+        "slug": "pov",
+        "core": false,
+        "unlocked_by": null,
+        "difficulty": 9,
+        "topics": [
+            "graphs",
+            "recursion",
+            "searching",
+            "trees"
+        ]
+    },
+    {
       "uuid": "e7351e8e-d3ff-4621-b818-cd55cf05bffd",
       "slug": "accumulate",
       "deprecated": true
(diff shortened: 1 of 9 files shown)


config.json
```diff
@@ -168,7 +168,8 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-
+        "logic",
```
cmccandless: What is the reasoning behind adding the `logic` topic?

rahulagrawal97: In the begining one will have to apply a little logic to see that it can be solved by using the mathematical formulas of sum of squares of first n natural numbers and square of sum of first n natural numbers

cmccandless: The `logic` topic is more applicable to exercises such as the [`zebra puzzle`](https://github.com/exercism/problem-specifications/blob/ae1cdbd67df28f0c4b2219237e8202f280dc8f19/exercises/zebra-puzzle/description.md). I don't think that it applies to this exercise.

rahulagrawal97: Can I make the change you suggested or the one who merges the request will ave to do so?

cmccandless: @rahulagrawal97 You should make the change. Just remove the `logic` topic here.

cmccandless:
@rahulagrawal97 There is only a very minor fix required for this to merge. As you have not granted maintainers permissions to push commits in your fork, I cannot make the change myself in this PR. Please make the change discussed above so that we can merge.

N-Parsons:
@cmccandless, it looks like they have deleted their copy of this repository and therefore cannot make the necessary change. Unfortunately this means that this PR can't be merged.

cmccandless:
@N-Parsons ah, in that case I will close this PR.

PR #1081, opened 2017-10-30
update check-exercises.py to run skipped tests

### LINKED ISSUE #391: Run skipped tests

cmccandless:
Fixes #391

### DIFF
--- a/test/check-exercises.py
+++ b/test/check-exercises.py
@@ -30,7 +30,7 @@ def check_assignment(name, test_file):
         else:
             with open(test_file, 'r') as src_file:
                 lines = [line for line in src_file.readlines()
-                         if not line.strip().startswith('@unittest.skip(')]
+                         if not line.strip().startswith('@unittest.skip')]
             with open(test_file_out, 'w') as dst_file:
                 dst_file.writelines(lines)
         shutil.copyfile(os.path.join(os.path.dirname(test_file), 'example.py'),
--- a/test/check-exercises.py
+++ b/test/check-exercises.py
@@ -11,6 +11,9 @@ import sys
 import tempfile
 import json
 
+# Allow high-performance tests to be skipped
+ALLOW_SKIP = ['largest-series-product']
+
 
 def python_executable_name():
     return 'python{}.{}'.format(sys.version_info.major, sys.version_info.minor)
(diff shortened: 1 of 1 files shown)


N-Parsons:
This doesn't seem to be working as anticipated - there are still tests being skipped. The approach seems sensible, but somewhere along the line it's not working as anticipated.

As a general point, there's also a good reason why some tests might need to be skipped by the validation script - some tests actually take quite a long time to run!

N-Parsons:
Looking a bit closer at it, I think we can categorise them and then skip some and not others. For example, [linked-list](https://github.com/exercism/python/blob/master/exercises/linked-list/linked_list_test.py#L46) is using `@unittest.skip("extra-credit")`, and I think that this should run. But there are some, for example in [largest-series-product](https://github.com/exercism/python/blob/master/exercises/largest-series-product/largest_series_product_test.py#L69) that are computationally intensive and are unlikely to require a different implementation, and I think these could safely be skipped by a script (and maybe we insist that people check them manually when tests change).

I'm not sure what other category name we should use alongside `"extra-credit"`.

cmccandless:
@N-Parsons Yeah first two commits I really expected them to work and did the changes in-browser. I'm now working on this locally until I see expected results.

I like the idea of categories... maybe the computationally intensive ones could be tagged something like:
- `high-performance`
- `user-only`
- `run-manually`
- `no-CI`

cmccandless:
@N-Parsons oh; of course. I'm filtering the `skip` lines from the *example solution* and not the *test file* :face_with_head_bandage:. Fix forthcoming...

cmccandless:
@N-Parsons I've got a working solution. My current answer for handling computationally intensive tests is to allow `unittest.skip()` for a specific set of exercises listed in `ALLOW_SKIP`. This can easily be exchanged for a category-based solution (in this PR or later in another).

However, the more I think about this, the more I think that we need to re-evaluate our usage of `unittest.skip`. My suggestion would be the following:
- Remove all `extra-credit` annotations
- For computationally intensive tests (i.e. `longest-series-product`, etc), re-annotate with a different label (see above for suggestions).

*Note: while the `alphametics` tests currently appear to be computationally intensive, a better solution has been proposed in #1068 and will likely be merged soon, so these tests do not need marked.*

test/check-exercises.py
```diff
@@ -22,7 +25,14 @@ def check_assignment(name, test_file):
     example_name = modname_heuristic(test_file)
     try:
         test_file_out = os.path.join(workdir, os.path.basename(test_file))
-        shutil.copyfile(test_file, test_file_out)
+        if name in ALLOW_SKIP:
+            shutil.copyfile(test_file, test_file_out)
+        else:
+            with open(test_file, 'r') as src_file:
+                lines = [line for line in src_file.readlines()
+                         if not line.strip().startswith('@unittest.skip(')]
```
N-Parsons: I would suggest removing the opening bracket from the `startswith` argument (ie. make it `'@unittest.skip'`) - `@unittest.skip` can be used without arguments, so this risks not removing those skips and therefore causing problems in future.

test/check-exercises.py
```diff
@@ -11,6 +11,9 @@
 import tempfile
 import json
 
+# Allow high-performance tests to be skipped
+ALLOW_SKIP = ['largest-series-product']
```
N-Parsons: I'm happy with this solution - if we decide to do something different in the future, this list will make it easy to know where the changes need to be made anyway.

PR #1083, opened 2017-10-30
poker - Update tests to match canonical-data.json @v1.0.0

### LINKED ISSUE #1002: poker: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `poker` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/poker/poker_test.py](https://github.com/exercism/python/blob/master/exercises/poker/poker_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/poker/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/poker/canonical-data.json)

njgingrich:
Resolves #1002.
Updated tests to match canonical data.

I renamed some tests to better match up with the json file, and some of the implementations had slightly different cards so those were matched.

### DIFF
--- a/exercises/poker/poker_test.py
+++ b/exercises/poker/poker_test.py
@@ -2,82 +2,120 @@ import unittest
 
 from poker import poker
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
 
 class PokerTest(unittest.TestCase):
-    def test_one_hand(self):
+    def test_single_hand_wins(self):
         hand = '4S 5S 7H 8D JC'.split()
         self.assertEqual(poker([hand]), [hand])
 
-    def test_nothing_vs_one_pair(self):
-        nothing = '4S 5H 6S 8D JH'.split()
+    def highest_card_wins(self):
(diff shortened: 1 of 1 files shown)


exercises/poker/poker_test.py
```diff
@@ -2,118 +2,159 @@
 
 from poker import poker
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
```
cmccandless: Can you please add an additional line above this (2 above, 1 below) for consistency with the track?

njgingrich:
Sure thing! Sorry about the delay.

cmccandless:
@njgingrich Merged. Congrats! Thanks for working on this!

njgingrich:
Thanks!

PR #1085, opened 2017-10-31
secret-handshake: update tests to canonical version v1.1.0

### LINKED ISSUE #1006: secret-handshake: check test version and update if necessary
As identified in #784, some of the tests currently don't reference the version number for the `canonical-data.json` that they are based on, `secret-handshake` is one of them. In order to help with the ongoing maintenance of this track, we're in the process of checking the tests against the current version of the `canonical-data.json` - if they are up-to-date, then you can just add the version string, but if they are out-of-date, then the tests will need updating.

```
```

**Test suite:** [exercism/python/exercises/secret-handshake/secret_handshake_test.py](https://github.com/exercism/python/blob/master/exercises/secret-handshake/secret_handshake_test.py)
**Canonical data:** [exercism/problem-specifications/exercises/secret-handshake/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/secret-handshake/canonical-data.json)

cmccandless:
- updated tests
- removed redundant/conflicting tests
- updated example solution

Resolves #1006


exercises/secret-handshake/secret_handshake_test.py
```diff
@@ -1,47 +1,71 @@
 import unittest
 
-from secret_handshake import handshake, code
+from example import handshake, code
```
m-a-ge: Don't forget to revert this

exercises/secret-handshake/example.py
```diff
@@ -1,50 +1,17 @@
...
-    s = []
-    for g in seq:
-        if g not in gestures[idx:]:
-            return []
-        newidx = gestures.index(g, idx) + 1
-        s.extend(['0'] * (newidx - idx - 1) + ['1'])
-        idx = newidx
-    s.reverse()
-    return s
+def handshake(code):
```
m-a-ge: I like your names for parameters, what do you think of updating https://github.com/exercism/python/blob/master/exercises/secret-handshake/secret_handshake.py for consistency?

exercises/secret-handshake/example.py
```diff
@@ -1,50 +1,17 @@
...
+    actions = [gestures[i] for i in range(len(gestures))
+               if (code >> i) % 2 == 1]
+    return actions if code < 16 else list(reversed(actions))
+
+
+def code(actions):
```
m-a-ge: `secret_code` looks a bit better in this case I think

exercises/secret-handshake/secret_handshake_test.py
```diff
@@ -1,47 +1,71 @@
...
-    def test_shake_bin_invalid(self):
-        self.assertEqual(handshake('121'), [])
+    def test_do_nothing_for_zero(self):
+        self.assertEqual(handshake(0), [])
 
-    def test_unknown_action(self):
-        self.assertEqual(code(['wink', 'sneeze']), '0')
+    # Track-specific tests
 
     def test_code1(self):
```
m-a-ge: Maybe we should skip tack-specific tests and let the user decide if he/she wants to write the code which passes them?

exercises/secret-handshake/secret_handshake_test.py
cmccandless: Thanks.

exercises/secret-handshake/example.py
cmccandless: Done.

exercises/secret-handshake/example.py
cmccandless: Done.

exercises/secret-handshake/secret_handshake_test.py
cmccandless: I like that idea, since in this case the track-specific tests involve a method that isn't tested for in the canonical tests.

PR #1086, opened 2017-10-31
Make probot/stale more useful

cmccandless:
Propose the following changes:
- `daysUntilStale: 60 -> 21`: With 5 track maintainers (3+ active), issues are being manually closed after ~20 days frequently. 60 days is too long.
- Add `epic` to `exemptLabels`: `epic` denotes issues that are usually super-issues, and closing every sub-issue takes a lot of effort and time.
- `staleLabel: on hiatus -> abandoned`: `abandoned` is typically used by the maintainers here, so change for consistency

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -7,6 +7,8 @@ exemptLabels:
   - discussion
   - pinned
   - epic
+  - enhancement
+  - beginner friendly
 # Label to use when marking an issue as stale
 staleLabel: abandoned
 # Comment to post when marking an issue as stale. Set to `false` to disable
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,16 +1,17 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: 60
+daysUntilStale: 21
 # Number of days of inactivity before a stale issue is closed
 daysUntilClose: 7
 # Issues with these labels will never be considered stale
 exemptLabels:
   - discussion
   - pinned
+  - epic
 # Label to use when marking an issue as stale
-staleLabel: on hiatus
+staleLabel: abandoned
 # Comment to post when marking an issue as stale. Set to `false` to disable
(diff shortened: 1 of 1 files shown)


cmccandless:
@behrtam @N-Parsons any objections?

N-Parsons:
My only concern would be that some sub-issues of large ones could end up getting closed. The `epic` label would protect the main one, but the exercise-specific issues would still be vulnerable.

Perhaps we should also add the `beginner friendly` label to the exemption? These are good issues for new people, and I don't think it matters if they stay around longer than 21 days. I think `enhancement` might also be a good one to exclude? If we've tagged it as an enhancement, we've recognised it's value as something that should be changed, but sometimes that doesn't mean that it will be easily resolvable or resolved quickly.

Alternatively, we need another label (could be one that we have already) that we can add to things that just need doing, regardless of how long it sits there?

cmccandless:
@N-Parsons Apologies for the late response on this; somehow the notification must've slipped between the cracks. 

>Perhaps we should also add the `beginner friendly` label to the exemption? These are good issues for new people, and I don't think it matters if they stay around longer than 21 days. 

I agree with your reasoning.

>I think enhancement might also be a good one to exclude? If we've tagged it as an enhancement, we've recognised it's value as something that should be changed, but sometimes that doesn't mean that it will be easily resolvable or resolved quickly.

Fair enough.

>Alternatively, we need another label (could be one that we have already) that we can add to things that just need doing, regardless of how long it sits there?

We already have `pinned` for this.

N-Parsons:
@cmccandless, no worries - I've been busy myself recently and lost track of a lot of stuff.

If you're happy to add those things, then I guess I'm happy to approve the changes.

cmccandless:
@N-Parsons done.

N-Parsons:
I feel like there's a consensus here, so I've merged it. Thanks @cmccandless!

PR #1089, opened 2017-10-31
Create config.json linter

### LINKED ISSUE #1084: Create config.json linter
It would be nice to have Travis-CI check the formatting of `config.json` for pull requests so that maintainers don't have to micromanage edits to that file. This could be accomplished by adding a script `config.py` (happy to take suggestions on the name of this file) that supports the following sub-commands:
- add <exercise>
    - Creates a new entry in `config.json` for the specified exercise
- edit <exercise>
    - Makes changes to existing exercise entry in config.json
- remove <exercise>
    - Remove specified entry from `config.json`
- deprecate <exercise>
    - Marks the exercise as deprecated in `config.json`
- lint
    - Checks the exercises in `config.json` for style violations, using either a hard-coded format, auto-detecting format using the first exercise as a template (more trivial than it sounds), or some combination of the two. This would be called by Travis-CI.

I actually already have a [script](https://gist.github.com/cmccandless/cdd8741fe258ea3d3f3ccc5b97ef538f) that does most of the above, and could easily be modified to support the `lint` subcommand.

cmccandless:
Resolves #1084 

Decided to put what I have in a pull request for reference in discussion in #1084. If it is decided that #1084 is unnecessary, this PR can be closed.

# Test Cases

### slug

- [x] `slug` is provided by user for all relevant commands 
- [x] Type: r"^[a-z\-]+$"

### uuid

- [x] `uuid` is generated internally via `configlet uuid`
- [x] Type: r"^[a-f\d]{8}(-[a-f\d]{4}){3}-[a-f\d]{12}$" (RE: [#1096](/exercism/python/pull/1096#issuecomment-343169511)) (**technically redundant, since these are auto-generated)

### core

- [x] default: false
- [x] CLI format: --core (`true` if present)
- [x] Type: boolean

### unlocked_by

- [x] default: null
- [x] CLI format: --unlocked_by <exercise> (default if not present) (Ex: `config_helper.py add linked-list --unlocked-by simple-linked-list`)
- [x] Type: r"^[a-z\-]+$" (same as `slug`)
- [x] Validation: <exercise> is already defined in `config.json`

### difficulty

- [x] default: 1
- [x] CLI format: --difficulty <[1..10]> (default if not present) (Ex: `config_helper.py add clock --difficulty 2`)
- [x] Type: integer [1..10]

### topics

- [x] default: [ ] (empty list)
- [x] CLI format: --topics <topic1> [<topic2> [<topic3> ...]] (default if not present) (Ex: `config_helper.py add linked-list --topics lists collections`)
- [x] Type: r"^[a-zA-Z]+$" (each element)
- [x] Validation: topic exists in [topics list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) (each element)

### deprecated

- [x] Not included by default
- [x] Type: boolean

### config_helper.py add <exercise> [options]

- [x] Command supported
- [x] Key values supplied via flags (See above)
- [x] Key values not provided are set to default values
- [x] Complete entry data is returned (not just supplied values)
- [x] Cannot add existing exercise
- [x] `config.json` contains new entry after running

### config_helper.py edit <exercise> [options]

- [x] Command supported
- [x] Key values supplied via flags (See above)
- [x] Key values not provided are not modified
- [x] Complete entry data is returned (not just supplied values)
- [x] Cannot edit non-existent exercise
- [x] `config.json` contains changes to entry data after running

### config_helper.py remove <exercise> [options]

- [x] Command supported
- [x] Cannot remove non-existent exercise
- [x] `config.json` does not contain entry after running

### config_helper.py deprecate <exercise>

- [x] Command supported
- [x] Cannot deprecate non-existent exercise
- [x] `config.json` entry contains only the following after running: 

```json
{ 
"uuid": "<previous uuid>", 
"slug": "<exercise>",
"deprecated": true
}
```

#### Style codes

| Code | Description |
| --- | --- |
| CH10 | Missing root key |
| CH20 | Invalid root key value |
| CH100 | Invalid key value |
| CH110 | Incorrect key order |
| CH120 | Missing key |
| CH201 | Non-existent exercise |
| CH202 | Non-existent topic |
| CH300 | Invalid foregone exercise |

- [x] Command supported
- [x] Style violation codes are return with violation details

#### [CH10] Missing root key

- [x] Detects missing root keys
- [x] Details format: config.py:\<line_no\>: [CH10] root key "\<key\>" not found

#### [CH20] Invalid root key value

- [x] Validates root key values
- [x] Details format (key=language): config.py:\<line_no\>: [CH20] language must be "PYTHON"
- [x] Details format (key=active): config.py:\<line_no\>: [CH20] active must be either true or false

#### [CH110] Incorrect Key Order

- [x] Detects out of order keys
- [x] Details format: config.py:\<line_no\>: [CH110] \<exercise\> keys out of order; expected "<expected key>", found "<actual key>"

#### [CH120] Missing key

- [x] Detects missing keys
- [x] Details format: config.py:\<line_no\>: [CH120] \<exercise\>.\<key\> not found

##### uuid

- [x] Is Validated
- [x] Details format: config.py:\<line_no\>: [CH100] \<exercise\>.uuid: invalid; please generate a new uuid with configlet

##### slug

- [x] Is Validated
- [x] Details format: config.py:\<line_no\>: [CH100] \<exercise\>.slug: expected a string of the form "exercise-slug", found "\<slug\>"

##### core

- [x] Is Validated
- [x] Details format: config.py:\<line_no\>: [CH100] \<exercise\>.core: expected true or false, found "\<core\>"

##### unlocked_by

- [x] Is Validated
- [x] Details format: config.py:\<line_no\>: [CH100] \<exercise\>.unlocked_by: expected a string of the form "exercise-slug" or null, found "\<unlocked_by\>"
- [x] Details format: config.py:\<line_no\>: [CH201] \<exercise\>.unlocked_by: exercise "\<unlocked_by\>" not found

##### difficulty

- [x] Is Validated
- [x] Details format: config.py:\<line_no\>: [CH100] \<exercise\>.difficulty: expected an integer in range [1..10], found \<difficulty\>

##### topics

- [x] Is Validated
- [x] Details format: config.py:\<line_no\>: [CH100] \<exercise\>.topics: expected string of the form "insert_topic_here", found "\<topic\>"
- [x] Details format: config.py:\<line_no\>: [CH202] \<exercise\>.topics: topic "\<topic\>" not found


m-a-ge:
As for the name, maybe something like `config_formatter.py` or `config_helper.py`?

cmccandless:
@m-a-ge 
>Should we reformat the whole config.json? What do you think?

If we are going to use a linter, then it should be applied to the entirety of `config.json`. In the case of the linter I've implemented, this would be mostly whitespace changes since it detects key order automatically.

>As for the name, maybe something like config_formatter.py or config_helper.py?

I like the intent that those names convey, but would prefer if we could come up with something shorter... maybe `configfmt.py` or `confhelp.py`?

m-a-ge:
@cmccandless I believe underscore is still a preferable option in comparison to the length (because it improves readability) - https://www.python.org/dev/peps/pep-0008/#package-and-module-names

cmccandless:
@m-a-ge As a Python module, you would be right. I was approaching this as a stand-alone script that would normally be run directly, but you're approach is probably the correct one.

All things considered, my vote is for `config_helper.py`.

N-Parsons:
I like the idea, but the current implementation seems to have a few issues, and I think the prompts could be more informative of what is expected as an input; for example, giving the current value in brackets may be useful when editing, but when adding a new exercise, it doesn't tell me what the value should look like. For `topics`, I honestly don't know what format these should be in when I look at the command line interface! I also seem to be having several issues running this script.

In Python 3.6 I get a `FileNotFoundError` for `configlet uuid` when using the `config.py add` command. Python 3 has a slightly different input format for the arguments for `subprocess` - each word should be an element of a list (eg. `subprocess.check_output(["bin/configlet", "uuid"])`), and this seems to work in Python 2 as well. I'd also use `bin/configlet`, since I would expect that people would run this from the root directory and probably won't have `configlet` in their path.

And in Python 2.7, the call to `configlet uuid` works, but I can't seem to specify a difficulty as a number (eg. `1`) - it seems to need to be a string containing a number (eg. `"1"`). I think this is because you are using `eval`... It also doesn't handle empty input gracefully - I get a `SyntaxError` for an `EOF`. I don't understand why you are using `eval` here - it seems to cause me loads of problems, and it seems a strange choice regardless.

The below solution (from another project of mine), would probably be safer and more reliable than using `eval`.

```python
def from_repr(s):
    """Get an int or float from its representation as a string"""
    # Strip any outside whitespace
    s = s.strip()
    # "NaN" and "inf" can be converted to floats, but we don't want this
    # because it breaks in Mathematica!
    if s[1:].isalpha():  # [1:] removes any sign
        rep = s
    else:
        try:
            rep = int(s)
        except ValueError:
            try:
                rep = float(s)
            except ValueError:
                rep = s
    return rep
```

I think it would be useful to add some validation checks to make sure that difficulty is between 1 and 10, that the specified topics are in the [topics list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt), and that the `unlocked-by` exercise exists.

cmccandless:
@N-Parsons I'm not surprised that you find a lot of issues with this implementation; it was just something I quickly hacked together. Certainly some more work would be needed before deploying such a solution.

cmccandless:
After some thinking, here are a coupleconclusions I've come to:

- Current implementation of `config.py` needs a large amount of refactoring
- Exercism is all about TDD, so why not TDD this too?

So I've replaced `config.py` with `config_helper.py`, which is currently nothing more than a template. I suggest that we compile a list of test descriptions here (see below for starting point), implement the tests, then create a tool that conforms to those. This will have the added bonus of getting verified across several versions of Python by Travis-CI.

TEST CASES
- Will support `add` command for adding new entries to `config.json`
- Will support `edit` command for editing existing entries in `config.json`
- Will support `remove` command for removing existing entries from `config.json`
- Will support `deprecate` command for marking existing entries in `config.json` as `deprecated`
- Will support `lint` command for `config.json` style checking

cmccandless:
Also, for parsing args passed to commands, I'm thinking something along the lines of [this](http://chase-seibert.github.io/blog/2014/03/21/python-multilevel-argparse.html).

cmccandless:
I've started a more formalized test case list in the description for this PR. If anyone has something to add or notices something wrong with what is there so far, please comment.

cmccandless:
I've determined that I can safely use context managers to revert any changes I make to `config.json`, so the `--config` flag is not necessary for unit tests. Can anyone come up with another practical use case for the flag, or can I simply remove it from requirements?

cmccandless:
Also, while auto key order detection is a cool feature, is there a good reason not to hard-code a key order as a list in the beginning of the script? This would make it easily modifiable at a later date while simplifying development.

cmccandless:
@N-Parsons @m-a-ge I'd appreciate any input you can provide on the following:

>I've determined that I can safely use context managers to revert any changes I make to config.json, so the --config flag is not necessary for unit tests. Can anyone come up with another practical use case for the flag, or can I simply remove it from requirements?

>Also, while auto key order detection is a cool feature, is there a good reason not to hard-code a key order as a list in the beginning of the script? This would make it easily modifiable at a later date while simplifying development.

Additionally, input on test cases is still welcome.

config.json
```diff
@@ -1065,6 +1067,8 @@
     {
       "uuid": "1818f134-0ed9-e680-9b29-45ffd2b3344b0f841c7",
       "slug": "markdown",
+      "core": false,
+      "unlocked_by": null,
```
N-Parsons: Can you make a PR for these changes? I know these keys only matter for Nextercism, but I'd rather have them fixed sooner rather than later.

bin/fetch-topics
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+URL=https://raw.githubusercontent.com/exercism/problem-specifications/master/TOPICS.txt
+curl -s --location $URL > TOPICS.txt
```
N-Parsons: This will work on Linux/BSD (and maybe Mac?), but we'll need a platform-independent solution.

I got curious while writing this comment, and ended up writing a solution before even post it :stuck_out_tongue:. The code below saves the file, but actually, this code could (and maybe should) just be built into the script so that it doesn't need to save and then read the file.

```python
try:
    # Python 3
    from urllib.request import urlopen
except ImportError:
    # Python 2
    from urllib2 import urlopen


URL = 'https://raw.githubusercontent.com/exercism/problem-specifications/master/TOPICS.txt'

topics_file = urlopen(URL).read()

with open("TOPICS.txt", "wb") as f:
    f.write(topics_file)
```

It might be worth making the script fail gracefully if there's an error fetching the topics list and make it just not check the topics list? I feel like it would be a useful thing from a user point of view (the script would work when offline), but I'm not sure whether we be risking Travis-CI not checking topics if there's a (presumably unlikely) network error?

N-Parsons:
@cmccandless 

We can fairly safely assume that the `config.json` exists and is where we expect it to be since it's part of the repo, so I don't think we need a flag for specifying a location for it. The only benefit of being able to specify a path would be the ability to run the script from a location other than the base of the repo, but I can't think of a reason why someone would **need** to be able to do this. Also, since we're using `git`, changes are trivial to revert, so I don't really have a concern with overwriting the file and potentially breaking it (although avoiding breakage is desirable) - if you can do this with context managers then do, since it will make things easier for users.

I'd hard-code the key order just to keep it simple - I can't see a reason to add extra complexity that could cause easily errors. Besides, if the key order changes, we wouldn't update the order in the config by reordering the keys ourselves, we'd make the change in this script and run the script to regenerate the config.

In general, I would suggest keeping things simple (although I regularly fail to do this myself).

bin/fetch-topics
cmccandless: @N-Parsons 

I have no argument against a platform-independent solution, but if curl is not good enough for this, we should change it for `fetch-configlet` too.

My argument for writing topics to a file is for use cases where `config_helper.py` is run multiple times (probably just the `edit` and `lint` commands); performing a network action is must more time consuming than simply reading from a file, if the file is already available.

Regarding offline availability, I'm not entirely sure that's a practical use case. Sure, you can write code offline, but you'll need a network connection eventually to push commits.

There is another option regarding `TOPICS.txt`. We create our own curated track TOPICS.txt that is strictly a subset of the global topics list. We could (optionally) remove topics that are not applicable in Python (such as `lenses`). I'm not particularly in favor of this option, but it would eliminate another network dependency.

config.json
cmccandless: Done. #1110

cmccandless:
I'm investigating why JSON key order is preserved on my machine and not in the Travis-CI environment.

cmccandless:
All implemented test cases are passing. Failure due to invalid UUID for `binary-search-tree`, fix in #1111

bin/fetch-topics
N-Parsons: @cmccandless, I hadn't seen that `fetch-configlet` was doing that. It seems to be an exercism-wide script, so I'm a little confused. I'm away at the moment and so don't currently have access to Windows to check, but I will take a look at it early next week. I'll create an issue for this later though so that it's not forgotten.

I'm not sure how often the `TOPICS.txt` file will be updated now that it's mostly settled in. My only concern about storing a file is that it could become out-of-date. There are a few different ways to handle this, but I'm not sure what would be best. I guess that Travis-CI would always download a fresh copy, so errors shouldn't get through even without an up-to-date copy.

I wouldn't be completely opposed to keeping our own subset `TOPICS.txt` and then considering upstream additions and optionally adding them. It would add a little more work for us, but I could easily write a scripted check that I can leave running on my server to give me a heads up about changes (perhaps even changes to `problem-specifications` in general). I could make them go to a public Trello board, or I could even make it raise issues here automatically.

TL;DR:
* I'll take a look at `fetch-configlet` to see if it needs some work.
* I think it's ok to store the file using the Python script.
* I'll write something to watch `problem-specifications` for us when I get chance.

cmccandless: It most definitely doesn't work on Windows. Command Prompt can't run bash script files, and running it under MinGW-bash (git-bash) results in the Linux executable being downloaded.

cmccandless: @N-Parsons I've implemented platform-independent file downloading (with zip/tar support, as appropriate) in `fetch.py`. I also ported `fetch-configlet` to Python and confirmed correct functionality on Windows. Travis passed, so Linux is also functioning correctly.

bin/fetch-configlet.py
```diff
@@ -0,0 +1,50 @@
...
+def get_arch():
+    arch = platform.machine()
+    if '64' in arch:
+        return '64bit'
+    elif '686' in arch:
+        return '32bit'
+    elif '386' in arch:
+        return '32bit'
+    else:
+        return '64bit'
```
N-Parsons: `configlet` also supports ARM. On ARMv7 `platform.machine() == "armv7l"` (at least on my Raspberry Pi 3) - the ARMv6 build seems to work find on this anyway. I'm not sure about ARMv8, since that's 64-bit but can apparently run the 32-bit code from earlier versions in some cases - I don't have a spare SD card at the moment to test it with my Pi, so I would suggest 

Feel free to tidy it up however you want, but this section should work:

```python
...
elif arch.startswith("armv5"):
    return "arm-v5"
elif arch.startswith("armv6") or arch.startswith("armv7"):
    return "arm-v6"
...
```

bin/fetch.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+import sys
+
+PLATFORM = os.name
+if PLATFORM == 'nt':
+    import zipfile
+
+    def extractall(archive):
+        with zipfile.ZipFile(archive, 'r') as zip:
+            zip.extractall()
```
N-Parsons: You should probably change `zip` to something else to avoid future confusion with the built-in `zip` function.

bin/fetch-topics
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+URL=https://raw.githubusercontent.com/exercism/problem-specifications/master/TOPICS.txt
+./bin/fetch.py $URL
```
N-Parsons: I think this script needs to be made platform-independent.

bin/fetch-configlet.py
cmccandless: I guess I shouldn't have assumed that the existing `fetch-configlet` script was covering all supported architectures, since it clearly wasn't tested on all of them. I'll add this section.

cmccandless:
As #1084 has been closed in favor of exercism/configlet#117, I am going to close this PR.

PR #1090, opened 2017-11-01
simple-linked-list: implement exercise

### LINKED ISSUE #738: simple-linked-list: implement exercise

cmccandless:
Resolves #738 

TODO:
- [x] create `README.md`
- [x] update `config.json`
- [x] write test cases
- [x] create solution template
- [x] write example solution


exercises/simple-linked-list/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+The simplest kind of linked list is a singly linked list. Each element in the
+list contains data and a "next" field pointing to the next element in the list
+of elements.
+
+This variant of linked lists is often used to represent sequences or
+push-down stacks (also called a LIFO stack; Last In, First Out).
+
+As a first take, lets create a singly linked list to contain the range (1..10),
+and provide functions to reverse a linked list and convert to and from arrays.
```
N-Parsons: This suggests that there should be a few more methods.

exercises/simple-linked-list/README.md
```diff
@@ -0,0 +1,38 @@
...
+pervasive in functional programming languages, such as Clojure, Erlang,
+or Haskell, but far less common in imperative languages such as Ruby or
+Python.
+
+The simplest kind of linked list is a singly linked list. Each element in the
+list contains data and a "next" field pointing to the next element in the list
+of elements.
+
+This variant of linked lists is often used to represent sequences or
+push-down stacks (also called a LIFO stack; Last In, First Out).
```
N-Parsons: The current implementation appears to be FIFO, so I think it should be reworked to match this description.

exercises/simple-linked-list/example.py
```diff
@@ -0,0 +1,48 @@
+class Node(object):
+    def __init__(self, value):
+        self._value = value
+        self._next = None
+
+    def value(self):
+        return self._value
+
+    def next(self):
+        return self._next
```
N-Parsons: Personally, I would make `value` and `next` into properties/attributes rather than methods.

exercises/simple-linked-list/simple_linked_list_test.py
```diff
@@ -0,0 +1,88 @@
+import unittest
+
+from example import LinkedList, EmptyListException
+
+
+# No canonical data available for this exercise
+
+class LinkedListTest(unittest.TestCase):
+    def test_empty_list_has_len_zero(self):
+        sut = LinkedList()
```
N-Parsons: What does `sut` mean? I would prefer a variable name with a clear meaning - `linked_list` would work.

cmccandless: `sut` is System Under Test.

exercises/simple-linked-list/example.py
cmccandless: The intent of making `value` and `next` methods was that they could be included as stubs in the solution template but their actual implementation (simple as they might be) could be left to the user.

exercises/simple-linked-list/README.md
cmccandless: Correct; I misread the README as FIFO.

exercises/simple-linked-list/README.md
cmccandless: Testing for `reverse` and `to_list` is reasonable. `from_list` is arguably implemented in the constructor already.

cmccandless:
@N-Parsons I've addressed your comments. Please see my responses as I do not necessarily agree with all suggested changes.

exercises/simple-linked-list/hints.md
```diff
@@ -0,0 +1,3 @@
+## Hints
+
+To support `list()`, see [implementing an iterator for a class.](https://docs.python.org/2/tutorial/classes.html#iterators)
```
N-Parsons: There's an implementation difference between Python 2 and Python 3 that I think needs to be made clear here to avoid confusion (`next()` vs `__next__()`).

I would suggest linking to the Python 3 documentation (I suspect that more people will be using Python 3) and then mentioning that `__next__` should be replaced by `next` if using Python 2, or that `next` can be implemented for dual compatibility as:

```python
def next(self):
    return self.__next__()
```

exercises/simple-linked-list/simple_linked_list.py
```diff
@@ -0,0 +1,27 @@
+class Node(object):
+    def value(self):
+        pass
+
+    def next(self):
+        pass
+
+
+class LinkedList(object):
+    def __len__(self):
```
N-Parsons: I think we should include `__init__` here (and above) - they are necessary methods and are expected to take an argument.

exercises/simple-linked-list/hints.md
cmccandless: That's exactly what I used in the example solution; I'll add the suggested info to `hints.md`

exercises/simple-linked-list/simple_linked_list.py
cmccandless: Good catch; I meant to add that.

N-Parsons:
Thanks @cmccandless!

PR #1093, opened 2017-11-02
Add test generator

cmccandless:
The purpose of this test generator is to create a starting point for contributors when implementing/updating test cases for an exercise. Most likely, all but the simplest exercises will require at least a few minor changes to the generated code, but using a generator will save time for reviewers by creating some consistency between exercises and save time for contributors by reducing the amount of code they must write from scratch.

Other features that could also be added to this generator:
- Generate README using `configlet` for new exercises
- Generate solution template for new exercises
- Copy solution template to `example.py` for new exercises (contributor must still implement actual example solution)
- Update `config.json` with entry for new exercises

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,16 @@ Please see the [contributing guide](https://github.com/exercism/docs/blob/master
 We welcome both improvements to the existing exercises and new exercises.
 A list of missing exercise can be found here: http://exercism.io/languages/python/todo
 
+If you are implementing a new exercise or updating test cases:
+
+```Bash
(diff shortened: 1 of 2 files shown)


cmccandless:
Once exercism/problem-specifications#996 has been fully implemented, this script can be greatly improved, so let's put a pin in it for now.

cmccandless:
Updated to detect new schema.

cmccandless:
Closing, as discussion in #1513 has led in a different direction.

PR #1097, opened 2017-11-09
alphametics: Revised, increased speed

nikamirrr:
Reworked the solution, eliminated few inefficiencies

### DIFF
--- a/exercises/alphametics/example.py
+++ b/exercises/alphametics/example.py
@@ -68,7 +68,6 @@ def solve(an):
                            for let, dig in zip(t_lowdigs, t_digvals)])
             if testsum % 10 == 0:
                 # Low digit test passed, check the main digits
-                # print(lowdigdict)
                 # if there are no other digits that low digits,
                 # test the whole expression and return if OK
                 if len(maindigits) == 0:
--- a/exercises/alphametics/example.py
+++ b/exercises/alphametics/example.py
@@ -27,7 +27,7 @@ def solve(an):
                 expdict[letter] += esign * (10 ** p)
 
     # Extract all letters and check if they are really letters
-    alldigits = set("".join([w for s in alphaexp for w in s]))
+    alldigits = set(expdict.keys())
     if not alldigits <= acuset:
         raise ValueError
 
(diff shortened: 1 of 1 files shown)


exercises/alphametics/example.py
```diff
@@ -1,42 +1,99 @@
...
+        # Generate addtional non-zero digit permutations
+        for lonz in permutations(remnzdigs, len(lonzdigits)):
+            # Build a dictionary for to test the expression
+            t_digvals = lorest + lonz
+            # Evaluate the expression sides
+            testsum = sum([dig * loexpdict[let]
+                           for let, dig in zip(t_lowdigs, t_digvals)])
+            if testsum % 10 == 0:
+                # Low digit test passed, check the main digits
+                # print(lowdigdict)
```
cmccandless: Please remove `print()` statements (even commented out).

nikamirrr:
I think we can do even better, like a recursive algorithm to check the lowest digits,
move to next digits, generate them and check them.
A carryover will be accumulated for the next digits.
I am going to try to implement it over the weekend. Yesterday attempt was not very successful.
I may also try to design a special test case, which will cripple down the inefficient code.

nikamirrr:
Since the pull request is approved. Is there anything else I need to do on my side? Please, let me know

cmccandless:
@nikamirrr Regarding a recursive solution, it may not be as fast as you think. From [this article by Lars Kruse (found on the Python Wiki)](https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Data_Aggregation):
> Function call overhead in Python is relatively high, especially compared with the execution speed of a builtin function.

nikamirrr:
I understand the general impact of the recursion. But if it allows to drop bad paths early enough it will be beneficial

cmccandless:
@nikamirrr Fair enough. I'm curious to see how it turns out.

N-Parsons:
Looks like run time is down to 0.5 seconds now compared to ~5 seconds with the previous iteration.

nikamirrr:
Please, merge as is. Any further improvement is experimental and a bit less likely.

cmccandless:
@nikamirrr Merged. Thanks!

nikamirrr:
@cmccandless 
Looks like the recursive solution is faster.
The function call overhead is compensated by the fact that a good found candidate is traced through the next levels immediately and bad ones are rejected faster and do not build up.

Issue #1095, opened 2017-11-07
Create maintainer-specific docs


m-a-ge:
As mentioned in #1086 I think we should create some sort of maintainer-specific docs and document thing related to the best practices we use. It will help us stay on the same page etc

What do you think?

TODO:
- [ ] Add docs on how to choose Exceptions correctly per https://github.com/exercism/python/issues/1095#issuecomment-345554041

cmccandless:
Some of the information I would imagine the proposed docs would contain (Conventions, Code Style) is already in the [README](https://github.com/exercism/python/blob/4f083db4d3e55feffccbbf743c5eaa20cfb6fb4e/README.md). I don't see why we couldn't put label documentation there as well?

m-a-ge:
I thought that README is targeting contributors more than maintainers. For that reason I've created this issue

N-Parsons:
@m-a-ge, I'm not sure what documentation we'd need that is specific to maintainers?

However, I think it might be worth putting together a `CONTRIBUTING.md` (or equivalent) that would give more detail than the README currently does about how to contribute, what conventions we follow, etc. This would be useful both for new people to find and read, and for us to be able to point to when something isn't quite right with a PR. We could probably achieve a lot of the kinds of things you are thinking about with such a document?

m-a-ge:
@N-Parsons yes, that's a good idea

N-Parsons:
Coming from #1080, part of this documentation could be a list of exception types and when to use them (something like the [Choosing Exceptions to Raise](https://github.com/pandas-dev/pandas/wiki/Choosing-Exceptions-to-Raise) Pandas has).

cmccandless:
For now, [the README](https://github.com/exercism/python/blob/master/README.md) for this track is adequate as a contributing guide. If someone submits a PR for a more comprehensive CONTRIBUTING.md, it would be welcome, but this is not an urgent need at this time.

PR #1100, opened 2017-11-10
all exercises: Add __init__.py files to exercise directories

wsgac:
Some environments (e.g. Elpy under Emacs) find it difficult to
properly import solution files to test files (at least when running
tests interactively) and to jump between definitions, unless exercise
directories are treated as modules.


cmccandless:
I'm not sure how you're running the tests, but `pytest` does not require exercises to be modules and is the official way of running tests for this track. The issue you're describing sounds like it is specific to your individual development environment, and I don't think that environment-specific fixes necessarily belong here.

wsgac:
Yes, it's kind of specific to Emacs/Elpy. It might (though I'm not sure) apply to other environments as well. I know that you can run individual test cases like this: `pytest <test_file>::<test_class>::<test_method>` which gives you some granularity, but sometimes I wanted to run tests from the Python REPL and having the `__init__.py` files there helped. I won't push though :).

cmccandless:
Maintainers' Note: I'm marking as invalid for future reference; however, I don't feel that this label is as accurate as it could be (maybe something like `environment-specific` would be better?). Tagging #1095 for discussion on labels.

cmccandless:
@wsgac I'm going to close this, since it won't be merged. Thank you for putting in the PR anyway; if someone raises this issue again, we'll reference this PR for a user-specific solution.

wsgac:
Ok, thanks.

PR #1101, opened 2017-11-12
Ignore pytest cache directory

N-Parsons:
If people use `pytest-cache` when working on exercises, it creates a `.cache` directory that is currently not ignored - I've experienced this myself, and I just saw a PR with this issue.

This PR updates `.gitignore` to ignore the cache directories of `py-test` and Python modules (`.cache` and `__pycache__`, respectively).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ bin/configlet
 bin/configlet.exe
 .idea/
 .cache
+__pycache__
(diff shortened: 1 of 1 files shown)


N-Parsons:
@cmccandless, I've added `__pycache__` - I at the very least it can't hurt to ignore it.

cmccandless:
@N-Parsons merge at will.

PR #1104, opened 2017-11-13
go-counting: fix typo in import

cmccandless:
There is no `gocounting` module; it should be `go_counting`


N-Parsons:
My only question is: how did this pass the tests with the wrong module name?

cmccandless:
@N-Parsons I was wondering that myself... I have no answer.

PR #1105, opened 2017-11-13
check-exercises.py: Enforce strict module naming

N-Parsons:
This PR simplifies the `check-exercises.py` script by enforcing strict module naming.

The previous version used a heuristic to determine the correct name for the example to be renamed to for the tests to run. This meant that:
* The test-suite could import an arbitrary module name as long as it didn't exist.
* If the script was run locally, it would fail to determine the module name if the name clashed with an installed module (eg. `markdown`).
* Incorrect module names were accepted by Travis-CI checks, and thus could be displayed to learners and prevent their tests from passing.

~~DO NOT MERGE THIS!!!~~
~~This PR is designed to test Travis-CI for a possible issue I noticed due to #1104.~~

### DIFF
--- a/test/check-exercises.py
+++ b/test/check-exercises.py
@@ -2,8 +2,6 @@
 from __future__ import print_function
 
 import os
-import ast
-import imp
 import glob
 import shutil
 import subprocess
@@ -22,7 +20,7 @@ def python_executable_name():
 def check_assignment(name, test_file):
     # Returns the exit code of the tests
     workdir = tempfile.mkdtemp(name)
-    example_name = modname_heuristic(test_file)
+    example_name = name.replace("-", "_")
     try:
         test_file_out = os.path.join(workdir, os.path.basename(test_file))
         if name in ALLOW_SKIP:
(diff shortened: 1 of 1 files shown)


N-Parsons:
@cmccandless, it looks like we might need to look into this issue a bit more? If the import is broken, it still passes the tests. If we then break the test cases though, the tests fail as we would expect. This raises the question, where is Travis-CI importing modules from? And, can we make it check that the import name is correct?

cmccandless:
Travis imports from a temporary file that contains the contents of the exercise's `example.py`. I suspect that at the time of implementing `check_exercises.py`, there were not strict rules about module naming. At this point, we could probably decide switch to converting the exercise directory name to snake_case to obtain the expected module name.

⁣


Corey McCandless​

N-Parsons:
I also found an issue running this script locally because I have the `markdown` package installed, and therefore the script finds the `markdown` module and thus fails to correctly rename `example.py`.

My solution, as you suggested, was to enforce strict naming for the modules, since this is what I would expect it to do anyway.

PR #1107, opened 2017-11-14
parallel-letter-frequency: Mark exercise as foregone

### LINKED ISSUE #1106: parallel-letter-frequency: Needs major rework
As @AtelesPaniscus raised in [#1099 (comment)](https://github.com/exercism/python/issues/1099#issuecomment-344240302) there are a number of issues with the current implementation of `parallel-letter-frequency`.

**These issues are:**
* It currently appears early in the `config.json`, meaning that in the current version of exercism.io, it's an early exercise.
* The current tests don't check for parallel execution, and it's therefore possible to pass all tests with non-parallel code, and several submitted solutions are doing this.

I feel that the last point could be resolved by adding a hint about the use of `multiprocessing` and then trusting the learner to implement it faithfully (the new mentorship model will also help here). I think that will only work though, if we move the exercise to near the end of the list, so that people are confident enough to do it properly.

Having looked at the current implementation more closely, I think that there is a further issue in that it isn't truly using parallel execution, since it is using `threading` rather than `multiprocessing`. ([An article discussing parallelisation](https://www.quantstart.com/articles/parallelising-python-with-threading-and-multiprocessing)) suggests that `threading` doesn't give simultaneous computation, but rather allows one thread to run while another is waiting for something (eg. I/O).

We do also have the option to forego an exercise if we don't think that it makes sense for this track.

I welcome opinions on how we proceed. My current suggestion would be to remove the exercise until we have resolved these issues and know what we want from this exercise.

N-Parsons:
Resolves #1106 by marking `parallel-letter-frequency` as foregone and removing the exercise directory.

I'm open to discussion on whether the directory and (incorrect) implementation should remain in the repo, but I have removed it because the current implementation is wrong (and not parallel) and so I feel that it holds no value to retain it.


N-Parsons:
I'm going to go ahead and merge this now, since there are issues with this exercise that could put learners off.

cg-cnu:
Thanks @N-Parsons  for removing it. Just working on it and the title and instructions are not clear and seems to be a bottleneck for many people to move forward.
I guess the numbers says it all

image

N-Parsons:
@cg-cnu, it was a fairly recent addition to the track (9 days ago), which is most of the reason why the numbers are so low. Hopefully people weren't too put off during that time - we'll be more careful in future with where in the config each exercise is being placed, although the upcoming revision to the way Exercism works should also stop this kind of thing from happening.

cg-cnu:
Hey @N-Parsons Sorry, its a misunderstanding. Didn't know that its a recent addition. You guys are doing awesome work ❤️  Thanks for your time 🙇

PR #1096, opened 2017-11-09
config.json: Fix invalid UUIDs

nywilken:
This change pertains to https://github.com/exercism/configlet/issues/99

PR #1111, opened 2017-11-16
binary-search-tree: replace invalid uuid

cmccandless:
Re #1096, valid uuid is X(8)-X(4)-X(4)-X(4)-X(12) where X is [\da-f]. This exercise must've been merged after #1096 or was simply missed.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1187,7 +1187,7 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


cmccandless:
@N-Parsons I actually caught this testing #1084. If #1084 is merged the check will be automatic.

PR #1113, opened 2017-11-18
Implement checks for raising messages with exceptions

### LINKED ISSUE #1080: Testing exception messages
Re discussion in #1052, what is the general consensus on testing for exception messages? The way I see it, there are a few ways we can handle this (in order of work required):
1. Ignore exception messages
2. Test only for exception messages that are explicitly included in the canonical data for the exercise
3. When exception messages are not explicitly stated in the canonical data, require that the exercise contributor invent messages for all expected exceptions

Additionally, if we are to test for exception messages at all, we need to have a conventional method of doing so in both Python2 and Python3. `assertRaisesRegexp` is replaced by `assertRaisesRegex` in Python3, so as mentioned in #1052 there are a few ways of handling this difference.

N-Parsons:
Resolves #1080 by adding checks that all exceptions are raised with messages. The aim is to encourage learners to raise meaningful messages, but testing for meaningfulness isn't feasible, so this instead just tests that a message of non-zero length is passed when raising an exception.

### TODO:
- [X] Update any failing examples
- [X] Add section to the README insert to explain the need for meaningful exception messages
- [X] Regenerate READMEs to include the new insert


N-Parsons:
I've held back the change to the `proverb` exercise description due to significant changes and put these in a separate PR (#1114) with the associated changes to the test-suite. This means that there will be a merge conflict, but I should be able to quickly resolve this once one is merged.

N-Parsons:
@cmccandless, all changes reflect changes to the upstream problem descriptions. I had a look through and nothing significant appeared to be changing (except in `proverb`, hence #1114), so the changes shouldn't cause issues.

PR #1116, opened 2017-11-20
say: change to more appropriate exception type

cmccandless:
From the [Python documentation](https://docs.python.org/2/library/exceptions.html#exceptions.AttributeError), `AttributeError` is "raised when an attribute reference or assignment fails." This situation is more appropriate for [`ValueError`](https://docs.python.org/2/library/exceptions.html#exceptions.ValueError): "Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value..."


m-a-ge:
Guys, sorry for my disappearance, I'll try to catch up

PR #1117, opened 2017-11-21
series: Update exercise to canonical-data v1.0.0

### LINKED ISSUE #1109: Exercise named series:  Poor Test Cases ?
The README for this simple exercise reads:

> Given a string of digits, output all the contiguous substrings of length n in that string.

For "49142" you might reason the substrings of length n are ["491", "914", "142"].  You would be wrong.

If you just look at the README instead of actually reading it, you might be forgiven for thinking the answer that is wanted is:  [491, 914, 142].  You would be wrong again.

Someone please explain how [[4, 9, 1], [9, 1, 4], [1, 4, 2]] is an appropriate answer in any programming language let alone Python.

Note:  this is not a new exercise and there appears to be no 'canonical' data to blame.

This exercise illustrates a more general point.

In the real world, people code to pass the tests because they don't have access to the requirements (because if they did it wouldn't help).  Shame.  For Exercism exercises would expect (and I hope you would agree) the student should be able to 'reason about' the problem and design a solution without having to study the test cases.  (Studying the test case may be useful to debug a solution but should not be necessary to design it.)

This clearly isn't the case for this exercise but there are others with incomplete problem statements that more or less oblige the student to study the test cases to figure out what is wanted.  I fear this is encouraging students to glance at the read me and then study the test cases.  They are learning to reason about the solutions, not the problem (arghhh the engineering mentality).

Consider this exercise's problem statement again.  Would the output for ("123123", 3) contain "123" twice or only once ?

I infer twice and, if I were wrong, I would expect at least one test case to fail as a consequence.  There is no test case for this so my solution would pass anyway and so would those of all the students who never reasoned about this.

What about the order in which substrings are output ?  I infer that since it is not mentioned it is unimportant.  Wrong.

I would like to suggest that the order in which substrings are output is not an implementation detail but more likely a side-effect of how the student reasoned about the problem.  Perfectly good reasoning could lead to a solution that fails all tests and forces a redesign.  For simple Exercism exercises like this, this should just not be possible because it is possible to write a complete enough README.

N-Parsons:
**Note:** This PR is dependent on exercism/problem-specifications#1020 and should only be merged after that PR is finalised and merged.

Resolves #1109 by clarifying the description and reworking the example and tests to expect a list of strings.

### DIFF
--- a/exercises/series/README.md
+++ b/exercises/series/README.md
@@ -1,18 +1,18 @@
 # Series
 
 Given a string of digits, output all the contiguous substrings of length `n` in
-that string.
+that string in the order that they appear.
 
 For example, the string "49142" has the following 3-digit series:
 
-- 491
-- 914
-- 142
+- "491"
+- "914"
+- "142"
 
 And the following 4-digit series:
 
-- 4914
-- 9142
+- "4914"
+- "9142"
 
 And if you ask for a 6-digit series from a 5-digit string, you deserve
 whatever you get.
--- a/exercises/series/example.py
+++ b/exercises/series/example.py
@@ -1,6 +1,4 @@
 def slices(series, length):
-    numbers = [int(digit) for digit in series]
-    if not 1 <= length <= len(numbers):
-        raise ValueError("Invalid slice length for this series: " + str(
-            length))
-    return [numbers[i:i + length] for i in range(len(numbers) - length + 1)]
+    if not 1 <= length <= len(series):
+        raise ValueError("Invalid slice length: {}" + str(length))
+    return [series[i:i + length] for i in range(len(series) - length + 1)]
--- a/exercises/series/series_test.py
+++ b/exercises/series/series_test.py
@@ -9,36 +9,54 @@ import unittest
 from series import slices
 
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
+
 class SeriesTest(unittest.TestCase):
     def test_slices_of_one(self):
         self.assertEqual(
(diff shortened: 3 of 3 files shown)


AtelesPaniscus:
@N-Parsons  The changes look fine to me.

I have one remark of general pertinence to the Python track.  The construction of test suites favours (usually requires) solutions that use lists and discourages those that use some other iterable (e.g. generators and generator expressions).  I assume this is not deliberate.  In the revised readme for this exercise I saw no mention of lists.  So why should an iterable that, when evaluated, yields all the right substrings in the right order be a fail and not a pass ?

The issue, to me, is about the students learning experience.  I've plenty of practice now with list comprehensions and I'd like to practice generators and explore how this interesting alternative affects how I think about problems.

OK, I can (and do) make the effort and my solutions have a couple of extra lines of code compared with others.  I don't think many others will.  Some students' work shows them to be sufficiently unsure of themselves that I expect they will take the track not accepting iterables as suggesting generators are bad.  Most students seem to believe that any line of code may contains a bug so best the way to avoid bugs is to write fewer, not more, lines of code.

I don't think this `series` exercise is well suited to generators (`anagram` is) but here is an example:

```
def slices(series, length):
    if len(series) < length or length < 1:
        raise ValueError

    def slice_generator(series):
        while len(series) >= length:
            yield [int(x) for x in list(series[:length])]
            series = series[1:]
                                                                                                                                                                                                                                             
    return list(slice_generator(series))
```
Your example solution could return a generator just by changing the [...] to (...) in the return statement.

PR #1118, opened 2017-11-21
Migrate string formatting operations to `str.format`

N-Parsons:
I noticed that some of the exercises were still using the old style string formatting (eg.`"%s" % value`) rather than using the newer `str.format` method - this PR moves everything to the newer style.

### DIFF
--- a/exercises/beer-song/example.py
+++ b/exercises/beer-song/example.py
@@ -8,8 +8,8 @@ def song(first, last=0):
 
 def verse(number):
     return ''.join([
-        "%s of beer on the wall, " % _bottles(number).capitalize(),
-        "%s of beer.\n" % _bottles(number),
+        "{} of beer on the wall, ".format(_bottles(number).capitalize()),
+        "{} of beer.\n".format(_bottles(number)),
         _action(number),
         _next_bottle(number),
     ])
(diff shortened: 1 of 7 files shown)


exercises/beer-song/example.py
```diff
@@ -19,13 +19,15 @@ def _action(current_verse):
     if current_verse == 0:
         return "Go to the store and buy some more, "
     else:
-        return "Take %s down and pass it around, " % (
-            "one" if current_verse > 1 else "it"
+        return "Take {} down and pass it around, ".format(
+            "one" if current_verse > 1 else "it",
```
cmccandless: Was surprised that the Python interpreter doesn't complain about the trailing comma here; was there a particular reason for adding it?

N-Parsons: It's a stylistic thing that helps to avoid bugs and make diffs cleaner if someone adds an entry in future. It probably won't ever make a difference here, so I'm happy to remove them if you want - I mostly put them in to satisfy a `flake8` extension that I have installed for other projects. They shouldn't cause issues though, since it's perfectly allowable in Python.

cmccandless: Now you've got me curious: what's the extension? I've never seen an implementation of flake8/pep8/pyflakes that complains about those commas not being there.

Regarding this PR, I've no issue with leaving them. Was just curious.

N-Parsons: It's `flake8-commas` ([PyPI](https://pypi.python.org/pypi/flake8-commas)). There are quite a few different extensions if you check `pip search flake8`.

PR #1119, opened 2017-11-23
Alphametics exercise updated 

nikamirrr:
Let's cripple the brute force solutions. :)


cmccandless:
@nikamirrr I'm hesitant to add such a test on a whim. Currently this exercise is listed as a difficulty 6 in `config.json`; this new test would certainly raise that, as brute-force solutions could easily take minutes to solve the 10-letter test (satisfying your goal of removing them as viable solutions), and non-brute-force solutions take much more thought. I would be curious to know the opinions of those more active in the problem-specifications repo on this... would you be willing to open a PR over [there](https://github.com/exercism/problem-specifications/blob/master/exercises/alphametics/canonical-data.json) to add this test? If they think adding the test is a good idea, I would be happy to merge the test here.

nikamirrr:
Will do, sure.

Though, I would question the real difficulty. The most difficult part is the recursive mod-10 login.
But beyond that all 10-letter problems are the same. 10! (3628800) brute force complexity.
All you need to do in order to make all 10-letter the same is to group and count all letters on both sides in the same decimal rank.
With little thought one can change a scary looking expression into the no more that 10-pair dictionary: letter -> multiplier to check the digit permutations.

Since you already have 10-digit problems among the tests, I don't think the one I add really increases complexity. But it will be very pedagogical stimulating the thought how one can more efficiently count the repeating letters.

nikamirrr:
My point is that if the current solutions for 10-letter problems are acceptable, it does not take much to make them handle the test I add. All one need to do is to group the letters on each side.

cmccandless:
> But beyond that all 10-letter problems are the same.

Allow me to suggest that you are *almost* correct; if `N` is the number of unique letters in the problem, all problems will have `O(N!)` time complexity, so problems with equal `N` will be the same in that regard. 

However, consider if `M` is the *total* number of letters in the problem. Because of the need to check each candidate solution against the actual problem (requiring `M` substitutions each time), our actual time complexity is `O(N! x M)`, which will be vastly different between a problem with 31 total letters and a problem with 10000 total letters.

nikamirrr:
You do not need to count every letter individually. No matter how many letters you use, there are no more than 10 unique letters (unknowns) with 10 non-repeating options for each unknown
For example one can notice:
NO + NO + TOO == LATE
NO + NO + TOO - LATE == 0

10*N + O  + 10*N + O + 100*T+10*O+O == 1000*L + 100*A+10*T+E

20*N + 13*O + 90*T - 1000*L - 100*A - E == 0

So the complexity is O((10 + 1)!)

nikamirrr:
Anyway, I am going to pull the test out of the PR and suggest a solution only. The test case will go through the JSON

cmccandless:
Ah, I see. I misunderstood you in grouping the letters; did not think of that. You are completely correct then. My apologies.

⁣


Corey McCandless​

nikamirrr:
NP at all :) we are all having fun doing this.

exercises/alphametics/example.py
```diff
@@ -1,98 +1,94 @@
...
-                # if there are no other digits that low digits,
-                # test the whole expression and return if OK
-                if len(maindigits) == 0:
-                    testsum = sum([dig * expdict[let]
-                                   for let, dig in zip(t_lowdigs, t_digvals)])
-                    if testsum == 0:
-                        return dict(zip(t_lowdigs, t_digvals))
+    totchars = set()
+    for p, chardict in enumerate(tchars):
+        for c, cnt in tuple(chardict.items()):
```
cmccandless: This should run fine without wrapping `chardict.items()` in `tuple()`.

cmccandless:
@nikamirrr My mistake; when I remarked on the `tuple()` wrap, I was only doing a cursory line-by-line reading, and I didn't connect that you were deleting items from the dictionary while iterating through it. I've corrected my comment.

nikamirrr:
NP, we have time, I will add comments over the weekend and update PR for the review.
I should have commented the code, you are exactly right. I am not a fulltime pro developer, so documenting the code was never a requirement for me, which I know is not right )

nikamirrr:
Comments added

nikamirrr:
Hello, would anybody pull this, please?

cmccandless:
@nikamirrr apologies, I was waiting for exercism/problem-specifications#1024 to be merged first, then I forgot about it. Merged.

PR #1124, opened 2017-12-01
Add link to the Python 2.7 EOL clock for context

jflory7:
This PR simply adds a link to the Python 2.7 EOL clock when it's mentioned on the Python "About" page. I thought this would be a good page to link for context, if someone wants to know more about why Python 2.7 will lose active support.

Also, a while back, @kytrinyx asked some of us from Mozilla Open Source Student Network to say hello if we ever contributed to Exercism… hi @kytrinyx! 😄


docs/ABOUT.md
```diff
@@ -9,7 +9,7 @@ This is often invisible to the beginning programmer, but if there are uses for w
 
 Python was started by Guido van Rossum in 1989.
 Its name is an homage to the comedy troupe Monty Python. 
-Python 2 is used widely but support may end by 2020.
+Python 2 is used widely but support [may end by 2020](https://pythonclock.org/).
```
cmccandless: I like the idea of adding a link to validate this statement, but would be more in favor of linking to [PEP-373](https://www.python.org/dev/peps/pep-0373/#id2) since it comes from an official source (pythonclock.org is not owned or maintained by the Python Foundation).

@N-Parsons thoughts?

kytrinyx:
@jflory7 hiiii! It's great to see you here. I was traveling and missed the original ping, but *welcome* ✨ 🌻 ✨

docs/ABOUT.md
jflory7: @cmccandless Sorry for the late response. I can update this with a new commit soon, if the PEP-373 link is preferred!

kytrinyx: Yeah, an official source seems like a better choice here. Thanks!

cmccandless: @jflory7 were you still going to change the link to an official source?

jflory7:
@cmccandless Thanks for your patience! I updated and rebased my commit to use PEP-373. Should be good to merge now. 😄

cmccandless:
@jflory7 Merged; thanks!

PR #1125, opened 2017-12-01
BUG: fix tests for transpose (moved join outside of call to transpose)

mileskrains:
The tests were comparing 
`transpose("\n".join(input_line))` with `"\n".join(expected)`, 
and would all fail. 

Moving the `join` from inside to outside of the call to `transpose` has them to compare
`"\n".join(transpose(input_line))` with `"\n".join(expected)`,
which allows them to pass.

### DIFF
--- a/exercises/transpose/transpose_test.py
+++ b/exercises/transpose/transpose_test.py
@@ -15,13 +15,13 @@ class TransposeTests(unittest.TestCase):
 
     def test_two_characters_in_a_row(self):
         self.assertEqual(
-            transpose("A1"),
+            "\n".join(transpose("A1")),
             "\n".join(["A", "1"])
         )
 
     def test_two_characters_in_a_column(self):
         self.assertEqual(
-            transpose("\n".join(["A", "1"])),
+            transpose(["A", "1"]),
             "A1"
         )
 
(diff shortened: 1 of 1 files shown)


cmccandless:
@mileskrains I'm not sure I understand: how were the tests failing before? The expected behavior of `transpose()` is "AB\nCD" -> "AC\nBD". In fact, your changes actually break passing tests (see the Travis-CI build results).

mileskrains:
In the tests, I was assuming `input_line` as actual intended input, as the provided stub uses that as the argument name for `transpose`, and I was reading `expected` as the actual expected output.

While I now see where I went wrong, it seems to me that the test code could be modified such that they do not cause similar confusion for future readers.  Do you think my confusion is due to lack of familiarity with test code construction, or do you also see it as potentially confusing?

I apologize for the bad pull request, I'm new at this and I was over-eager to get my first bit of practice with the whole fork/branch/submit process.

cmccandless:
@mileskrains I understand the confusion, and would welcome a PR that moves the `'\n'.join()` call from inside the `transpose()` call to the `input_line = ` assignment to avoid future confusion. Just be sure to maintain the current order of execution.

mileskrains:
Thanks @cmccandless, I will submit one along those lines.

PR #1133, opened 2017-12-08
secret-handshake: fix broken link in README

sjwarner-bp:
Fixes link in README


cmccandless:
In accordance with exercism/problem-specifications#1029

exercises/secret-handshake/README.md
```diff
@@ -39,7 +39,7 @@ please see the [help page](http://exercism.io/languages/python).
 
 ## Source
 
-Bert, in Mary Poppins [http://www.imdb.com/character/ch0011238/quotes](http://www.imdb.com/character/ch0011238/quotes)
+Bert, in Mary Poppins [http://www.imdb.com/title/tt0058331/quotes/qt0437047](http://www.imdb.com/title/tt0058331/quotes/qt0437047)
```
cmccandless: It's redundant to use link syntax here, since the text itself is a link.
In other words:
```markdown
http://www.imdb.com/title/tt0058331/quotes/qt0437047
```
would be better than
```markdown
[http://www.imdb.com/title/tt0058331/quotes/qt0437047](http://www.imdb.com/title/tt0058331/quotes/qt0437047)
```

cmccandless:
Again, I failed to give this my full attention. The change I requested was probably unnecessary, but you don't have to revert it. These READMEs are typically generated by configlet, so *if* there's an issue with link formatting here that would be a configlet issue. Perhaps the link syntax is used to show that a given link was not created unintentionally.

As for this PR, I think we're good to merge. In the future, updating exercise README files can be done with `configlet generate <path/to/track> -p <path/to/problem/specifications> -o <exercise-slug>`. You would have to clone exercism/problem-specifications as well.

PR #1140, opened 2017-12-19
custom-set: implement exercise

### LINKED ISSUE #746: custom-set: implement exercise

cmccandless:
Resolves #746


cmccandless:
[Canonical data](https://github.com/exercism/problem-specifications/blob/dd3c07edfe6c15e31cb7c86a4d1e8ceab24cb879/exercises/custom-set/canonical-data.json)

exercises/custom-set/example.py
```diff
@@ -0,0 +1,42 @@
+class CustomSet(object):
+    def __init__(self, elements=[]):
+        self.elements = list(elements)
+
+    def empty(self):
+        return not any(self.elements)
```
N-Parsons: This should be `return not self.elements` - the current version would wrongly return `True` if `self.elements == [0]`

exercises/custom-set/custom_set.py
```diff
@@ -0,0 +1,30 @@
...
+    def empty(self):
+        pass
+
+    def __iter__(self):
+        pass
+
+    def subset(self, other):
+        pass
+
+    def disjoint(self, other):
```
N-Parsons: I think these should be `isdisjoint` and `issubset` to match with Python's implementation of `set`. I would also then change `empty` to `isempty` for consistency.

exercises/custom-set/example.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __iter__(self):
+        return iter(self.elements)
```
N-Parsons: `__iter__` is a kind of useful method to implement here, but I think that `__contains__` would be a more logical for testing whether a set contains a particular value. Presumably `__iter__` is providing that functionality here, but from a learner point of view, I don't think that this would be clear enough.

cmccandless: I somehow was not aware of the `__contains__` method previously, and though that implementing `__iter__` was the correct way of supporting the `in` operator... I will fix this, thanks!

exercises/custom-set/example.py
cmccandless: Good catch... fixing.

exercises/custom-set/custom_set.py
cmccandless: Agreed.

PR #1147, opened 2017-12-27
word-search: adapt testdata-format from canonical-testdata-format

pheanex:
I felt the "\n" are superfluous and therefore removed them (I've made this suggestion before in another Issue/PR).
Also this proposed format (with a list of strings) matches the format from [canonical-data of problem-specifications  ](https://github.com/exercism/problem-specifications/blob/master/exercises/word-search/canonical-data.json)
@N-Parsons and @behrtam could you review this please?


cmccandless:
@pheanex Merged; thanks!

PR #1149, opened 2018-01-01
Rename test cases in reverse_string_test.py so that they will run

flores-jacob:
Hi,
I fixed the naming of the test cases in reverse_string_test.py by adding the prefix `test_` in front of their names.  I noticed that they were not running and I made the small adjustments.

Thanks and regards,

J F


cmccandless:
@flores-jacob Not sure how I missed that in the initial PR to add the exercise: merged. Thanks!

PR #1152, opened 2018-01-03
rename unittests in reverse_string_test.py

kcover:
renamed the unittests to start with "test", otherwise they would not run.


cmccandless:
Fixed in #1149 ; closing.

PR #1153, opened 2018-01-03
Fixed class name

vlmutolo:
No tests would run because the class name was `ReverseStringTests` instead of `ReverseStringTest`, which matches the filename.


vlmutolo:
Issue was persistent. Filed prematurely.

PR #1156, opened 2018-01-06
phone-number: rename 2 tests

pheanex:
I added the reason for why the tests should fail to give users a better understanding and also to reflect the the naming from canonical


cmccandless:
@pheanex Merged; thanks!

PR #1159, opened 2018-01-09
grep: replace usage of real files

### LINKED ISSUE #534: grep: why is a unit test messing with files??
This is what mocking and monkey patching are for.

cmccandless:
Fixes #534

### DIFF
--- a/exercises/grep/grep_test.py
+++ b/exercises/grep/grep_test.py
@@ -1,7 +1,6 @@
-import os
 import unittest
 
-from grep import grep
+from example import grep
 
 
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
@@ -15,8 +14,7 @@ And Heroes gave (so stood the will of Jove)
 To dogs and to all ravening fowls a prey,
 When fierce dispute had separated once
 The noble Chief Achilles from the son
-Of Atreus, Agamemnon, King of men.
-'''
+Of Atreus, Agamemnon, King of men.'''
 
 MIDSUMMERNIGHTFILENAME = 'midsummer-night.txt'
 MIDSUMMERNIGHTCONTENTS = '''I do entreat your grace to pardon me.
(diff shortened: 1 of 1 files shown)


exercises/grep/grep_test.py
```diff
@@ -36,17 +37,55 @@
...
 Of Oreb, or of Sinai, didst inspire
-That Shepherd, who first taught the chosen Seed
-'''
+That Shepherd, who first taught the chosen Seed'''
+FILENAMES = [
+    ILIADFILENAME,
+    MIDSUMMERNIGHTFILENAME,
+    PARADISELOSTFILENAME,
+]
+builtins.FILES = {}
```
N-Parsons: Does this need to be `builtins.FILES` rather than just `FILES`?

exercises/grep/grep_test.py
```diff
@@ -36,17 +37,55 @@
...
+        return builtins.oldopen(name, mode, *args, **kwargs)
+
+
+builtins.open = open
+
+
+def remove_file(file_name):
+    del builtins.FILES[file_name]
+
+
```
N-Parsons: I think it would be useful for these lines (70-88) to have comments explaining what it's doing at each step. It'll just make it easier to deal with if something breaks in future.

exercises/grep/grep_test.py
```diff
@@ -36,17 +37,55 @@
...
+class File(object):
+    def __init__(self, name=''):
+        self.name = name
+        self.contents = ''
+
+    def read(self):
+        return self.contents
+
+    def readlines(self):
+        return [l + '\n' for l in self.read().split('\n') if l]
```
N-Parsons: Can you please swap `l` for a more meaningful (and less easily confused) variable name? When I first looked at this line, I thought they were `1`s.

exercises/grep/grep_test.py
cmccandless: When I tested with just `FILES`, an empty dictionary was being checked for file contents when the new `open()` was called from the solution file instead of the populated one.

exercises/grep/grep_test.py
cmccandless: Fair enough.

exercises/grep/grep_test.py
cmccandless: I'm not sure what other issue was at play when I tested it then. I will try making the suggested change and see how Travis handles it. If it passes I have no argument against it.

cmccandless: It passed.

PR #1161, opened 2018-01-10
fix deprecation warnings

cmccandless:
In changes made in #1113, a convention was added to ensure that all exceptions raises by exercise solutions contained messages. The existing Python2/3 compatible solution was actually calling `assertRaisesRegexp` in Python3, raising `DeprecationWarning`s in the Travis-CI output for all Python3 builds. This PR fixes this, accomplishing the original intent to call `assertRaisesRegex` in Python3 and `assertRaisesRegexp` in Python2.

### DIFF
--- a/exercises/all-your-base/all_your_base_test.py
+++ b/exercises/all-your-base/all_your_base_test.py
@@ -82,9 +82,10 @@ class AllYourBaseTests(unittest.TestCase):
     # Utility functions
     def setUp(self):
         try:
-            self.assertRaisesRegex = self.assertRaisesRegexp
+            with self.assertRaisesRegex(AttributeError, r".+"):
+                raise AttributeError('x')
         except AttributeError:
-            pass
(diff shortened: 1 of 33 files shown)


exercises/all-your-base/all_your_base_test.py
```diff
@@ -82,9 +82,10 @@ def test_both_bases_are_negative(self):
     # Utility functions
     def setUp(self):
         try:
-            self.assertRaisesRegex = self.assertRaisesRegexp
+            with self.assertRaisesRegex(AttributeError, r".+"):
+                raise AttributeError('x')
```
N-Parsons: A cleaner solution is to just call `self.assertRaisesRegex` without actually using it.

```python
try:
    self.assertRaisesRegex
except AttributeError:
```

N-Parsons:
Thanks @cmccandless!

PR #1163, opened 2018-01-10
triangle: update tests to canonical-data v1.0.0

### LINKED ISSUE #1008: triangle: check test version and update if necessary

cmccandless:
Fixes #1008 

[Canonical data](https://github.com/exercism/problem-specifications/blob/506de961e642c1cd4c7489d394b6dd148410b7fb/exercises/triangle/canonical-data.json)

### DIFF
--- a/exercises/triangle/triangle_test.py
+++ b/exercises/triangle/triangle_test.py
@@ -1,65 +1,65 @@
 import unittest
 
-from triangle import equilateral, isosceles, scalene
+from triangle import is_equilateral, is_isosceles, is_scalene
 
 
-# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.1.0
 
-class EquilateralTests(unittest.TestCase):
+class is_equilateralTests(unittest.TestCase):
     def test_true_if_all_sides_are_equal(self):
-        self.assertIs(equilateral([2, 2, 2]), True)
+        self.assertIs(is_equilateral([2, 2, 2]), True)
 
     def test_false_if_any_side_is_unequal(self):
-        self.assertIs(equilateral([2, 3, 2]), False)
+        self.assertIs(is_equilateral([2, 3, 2]), False)
 
     def test_false_if_no_sides_are_equal(self):
-        self.assertIs(equilateral([5, 4, 6]), False)
+        self.assertIs(is_equilateral([5, 4, 6]), False)
 
     def test_false_if_all_sides_are_zero(self):
-        self.assertIs(equilateral([0, 0, 0]), False)
+        self.assertIs(is_equilateral([0, 0, 0]), False)
 
     def test_sides_may_be_floats(self):
-        self.assertIs(equilateral([0.5, 0.5, 0.5]), True)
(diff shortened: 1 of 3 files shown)


exercises/triangle/triangle_test.py
```diff
@@ -1,67 +1,65 @@
 import unittest
 
-from triangle import Triangle, TriangleError
+from triangle import equilateral, isosceles, scalene
 
 
-class TriangleTests(unittest.TestCase):
-    def test_equilateral_triangles_have_equal_sides(self):
-        self.assertEqual(Triangle(2, 2, 2).kind(), "equilateral")
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
```
N-Parsons: Looks like these are actually the tests from version 1.1.0 now - there was an update in `problem-specifications` that moved inputs into an `input` key.

exercises/triangle/triangle_test.py
```diff
@@ -1,67 +1,65 @@
 import unittest
 
-from triangle import Triangle, TriangleError
+from triangle import equilateral, isosceles, scalene
```
N-Parsons: I would prefer these to be `is_equilateral`, `is_isosceles`, and `is_scalene` - it just makes it clearer what these functions do.

exercises/triangle/triangle_test.py
```diff
@@ -1,67 +1,65 @@
...
 
-    def test_isosceles_triangles_have_first_and_last_sides_equal(self):
-        self.assertEqual(Triangle(4, 3, 4).kind(), "isosceles")
+    def test_false_if_no_sides_are_equal(self):
+        self.assertIs(equilateral([5, 4, 6]), False)
 
-    def test_isosceles_triangles_have_two_first_sides_equal(self):
-        self.assertEqual(Triangle(4, 4, 3).kind(), "isosceles")
+    def test_false_if_all_sides_are_zero(self):
+        self.assertIs(equilateral([0, 0, 0]), False)
```
N-Parsons: I feel like this should be raising an exception, rather than just returning `False` - I know this is in the canonical data, but we're asking whether a point is an equilateral triangle and just ignoring the fact that they are different things.

cmccandless: I understand what you're saying, and under the old format of `kind() = 'equilateral'|'isosceles'|'scalene'`, I would've preferred raising an exception, because the usage of `kind()` assumed that the input was indeed a triangle. However, the question of "Is a point an equilateral triangle?" is a valid one, even if it isn't one that would normally need asking.

PR #1165, opened 2018-01-10
beer-song: update tests to canonical data v2.1.0

### LINKED ISSUE #1128: beer-song: update tests to canonical-data v2.0.0
Currently, the `beer-song` exercise does not reference the [canonical data](https://github.com/exercism/problem-specifications/blob/5c3baae204802cf1c0e73493a671b1de5da10f0f/exercises/beer-song/canonical-data.json). The current tests must be checked and, if necessary, updated to match the canonical data. A comment line referencing the canonical data version must also be added (see #784 for the format of this comment).

cmccandless:
Fixes #1128 

[Canonical data](https://github.com/exercism/problem-specifications/blob/506de961e642c1cd4c7489d394b6dd148410b7fb/exercises/beer-song/canonical-data.json)

### DIFF
--- a/exercises/beer-song/beer_song_test.py
+++ b/exercises/beer-song/beer_song_test.py
@@ -5,6 +5,87 @@ from beer_song import recite
 
 # Tests adapted from `problem-specifications//canonical-data.json` @ v2.1.0
 
+class BeerTest(unittest.TestCase):
+    def test_first_generic_verse(self):
+        expected = [
+            "99 bottles of beer on the wall, 99 bottles of beer.",
+            "Take one down and pass it around, 98 bottles of beer on the wall."
+        ]
+        self.assertEqual(recite(start=99), expected)
+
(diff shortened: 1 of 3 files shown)


exercises/beer-song/beer_song_test.py
```diff
@@ -1,70 +1,389 @@
 import unittest
 
-from beer_song import song, verse
+from beer_song import recite
+
+
+# Tests adapted from `problem-specifications//canonical-data.json` @ v2.1.0
+
+SONG = [
```
N-Parsons: Minor point, but I think it might be better to move this massive chunk of text to after the test class. Normally I would always expect constants to be at the start, but when completing exercises, I regularly look at the test file to see what the tests look like and having to scroll past ~300 lines would be a bit annoying.

cmccandless: Good point, I'll move it.

PR #1169, opened 2018-01-18
check-exercises: Skip deprecated exercises

N-Parsons:
This PR edits the `check-exercises.py` script to skip testing of deprecated ~~and foregone~~ exercises.

The logic used assumes that `"deprecated"` ~~and `"foregone"`~~ are not keys in active exercises, as is currently the case.

### DIFF
--- a/test/check-exercises.py
+++ b/test/check-exercises.py
@@ -47,7 +47,8 @@ def load_config():
         raise SystemExit(1)
 
     try:
-        problems = [entry['slug'] for entry in data['exercises']]
+        problems = [entry['slug'] for entry in data['exercises']
+                    if "deprecated" not in entry and "foregone" not in entry]
     except KeyError:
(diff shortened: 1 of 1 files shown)


test/check-exercises.py
```diff
@@ -47,7 +47,8 @@ def load_config():
         raise SystemExit(1)
 
     try:
-        problems = [entry['slug'] for entry in data['exercises']]
+        problems = [entry['slug'] for entry in data['exercises']
+                    if "deprecated" not in entry and "foregone" not in entry]
```
cmccandless: Foregone exercises are actually listed under a separate list in `config.json`, and do not use the `foregone` field. It won't hurt to leave this, but foregone exercises are already not run.

N-Parsons: Ah, I presumed that they were now listed in the same way as `deprecated` exercises - I've removed the conditional.

PR #1173, opened 2018-01-22
bob: Supplement problem specification in README

mRcfps:
Two improvements are proposed:

- Add the missing case for replying "Calm down, I know what I'm doing!".
- Clearer definitions of questions and yelling phrases.

### DIFF
--- a/exercises/bob/README.md
+++ b/exercises/bob/README.md
@@ -6,6 +6,8 @@ Bob answers 'Sure.' if you ask him a question.
 
 He answers 'Whoa, chill out!' if you yell at him.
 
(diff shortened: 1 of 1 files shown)


cmccandless:
Adding the new test case is good, but i think that working out the clear definitions is part of the exercise. The best thing would be to adhere to the [canonical exercise description](https://github.com/exercism/problem-specifications/blob/master/exercises/bob/description.md). If you feel information needs to be added that isn't there already, please feel free to submit a PR, over there. Changes there will eventually cascade down to this track.

mRcfps:
😁Thanks for getting me on the right path!

PR #1182, opened 2018-02-12
bowling: use context manager form for assertRaises (RE: #477)

cmccandless:


cmccandless:
This work is now included in #1272; closing.

PR #1263, opened 2018-02-14
rail-fence-cipher: improve stub file

mRcfps:
Modifications in two places:

- Adjust the order of function parameters to be consistent with [test file](https://github.com/exercism/python/blob/master/exercises/rail-fence-cipher/rail_fence_cipher_test.py).
- Remove `fence_pattern` function to give learners more freedom.


cmccandless:
@mRcfps Merged; thanks for working on this!

PR #1265, opened 2018-02-15
leap: update tests to v1.3.0

### LINKED ISSUE #1184: leap: update tests to v1.3.0
[Canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/leap/canonical-data.json)

mRcfps:
Closes #1184. 

Since canonical data was only updated for new input policy, which has nothing to do with test file in Python, so update of version number suffices.

### DIFF
--- a/exercises/leap/leap_test.py
+++ b/exercises/leap/leap_test.py
@@ -3,7 +3,7 @@ import unittest
 from leap import is_leap_year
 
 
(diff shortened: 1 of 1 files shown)


cmccandless:
@mRcfps Note that "Solves" is not a GitHub keyword for closing issues. Use Resolves, Fixes, or Closes instead.

PR #1271, opened 2018-02-15
Readme template/running tests

cmccandless:
Add "Running the tests" section to README template and regenerate README files.

Closes #533 (redundant)

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -20,16 +20,33 @@ raise Exception("Meaningful message indicating the source of the error")
 
 ## Running the tests
 
-To run the tests run the command `python -m pytest {{ .Spec.SnakeCaseName }}_test.py`
-(`py.test {{ .Spec.SnakeCaseName }}_test.py` for Python 2; `pytest <exercise>_test.py` for Python 3). Or run it with
-`-v` to be more verbose like this: `pytest {{ .Spec.SnakeCaseName }}_test.py`. Using the
-`-x` flag will cause the test to fail at the first failure instead of it
-trying all cases before exiting.
+To run the tests, run the appropriate command below:
 
-To run this exercises's tests for example, run this command:
-```
-python -m pytest {{ .Spec.SnakeCaseName }}_test.py -v
-```
+| Python Version | Command |
+| --- | --- |
+| 2.7 | `py.test {{ .Spec.SnakeCaseName }}_test.py` |
+| 3.3+ | `pytest {{ .Spec.SnakeCaseName }}_test.py` |
+| 2.7, 3.3+ | `python -m pytest {{ .Spec.SnakeCaseName }}_test.py` |
+
+<table>
+    <tr>
+        <td colspan="2"><strong>Common pytest options</strong></td>
+    </tr>
+    <tr>
+        <td>-v</td>
+        <td>enable verbose output</td>
+    </tr>
+    <tr>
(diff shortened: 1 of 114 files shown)


config/exercise_readme.go.tmpl
```diff
@@ -18,6 +18,19 @@ To raise a message with an exception, just write it as an argument to the except
 raise Exception("Meaningful message indicating the source of the error")
 ```
 
+## Running the tests
+
+To run the tests run the command `python -m pytest {{ .Spec.SnakeCaseName }}_test.py`
+(`py.test {{ .Spec.SnakeCaseName }}_test.py` for Python 2; `pytest <exercise>_test.py` for Python 3). Or run it with
```
kytrinyx: This should be `{{ .Spec.SnakeCaseName }}` instead of `<exercise>_test.py` for Python 3 as well.

config/exercise_readme.go.tmpl
```diff
@@ -18,6 +18,19 @@ To raise a message with an exception, just write it as an argument to the except
...
+`-v` to be more verbose like this: `pytest {{ .Spec.SnakeCaseName }}_test.py`. Using the
```
kytrinyx: This line doesn't make sense. It seems to be explaining the verbose flag, but doesn't use it in the example.

config/exercise_readme.go.tmpl
```diff
@@ -18,6 +18,19 @@ To raise a message with an exception, just write it as an argument to the except
...
+`-x` flag will cause the test to fail at the first failure instead of it
+trying all cases before exiting.
+
+To run this exercises's tests for example, run this command:
+```
+python -m pytest {{ .Spec.SnakeCaseName }}_test.py -v
+```
```
kytrinyx: I find this section a little bit hard to scan. How about something like:

------

The command to run the tests is:

- (Python 2) `python -m py.test {{ .Spec.SnakeCaseName }}_test.py`
- (Python 3) `python -m pytest {{ .Spec.SnakeCaseName }}_test.py`

Additional flags are available for pytest:

- `-v` to get more verbose failure messages
- `-x` to fail at the first failure instead of running all the tests and showing all the failures at once

For example:

```
python -m pytest -x {{ .Spec.SnakeCaseName }}_test.py
```

cmccandless: I've rewritten this section for clarity; however, it just occurred to me that while tables are valid markdown, they might be discouraged in this instance to make it easier for users to view the README file in a text editor. If that is the case, I am happy to convert my changes to a similar format to what you have written.

kytrinyx: Yeah, this needs to be readable in plain text.

exercises/accumulate/README.md
```diff
@@ -39,6 +39,36 @@ To raise a message with an exception, just write it as an argument to the except
...
 
+## Running the tests
+
+To run the tests, run the appropriate command below:
+
+| Python Version | Command |
+| --- | --- |
+| 2.7 | `py.test accumulate_test.py` |
+| 3.3+ | `pytest accumulate_test.py` |
+| 2.7, 3.3+ | `python -m pytest accumulate_test.py` |
```
kytrinyx: This might be obvious to people who use python, but I have no idea from looking at this why 2.7 and 3.3+ both have two different commands. When would I use one or the other? What would I actually type on my command-line (given that I have 2.7) to run the tests?

PR #1298, opened 2018-02-16
sum-of-multiples: update tests to v1.2.0

codePrady:
All test cases are covered even now and they all pass in both python 2.7 and python3. Hence just changing the version alone.


cmccandless:
Duplicate of #1283. Please review open PRs to make sure the issue you are attempting to fix is not already being worked on before creating a new one.

PR #1306, opened 2018-02-17
twelve-days: re-implement according to canonical data

### LINKED ISSUE #1221: twelve-days: update tests to v1.1.0
[Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/canonical-data.json)

mRcfps:
Closes #1221. Re-implement this exercise according to [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/canonical-data.json).

### DIFF
--- a/exercises/twelve-days/twelve_days_test.py
+++ b/exercises/twelve-days/twelve_days_test.py
@@ -1,67 +1,69 @@
 import unittest
 
-from twelve_days import sing, verse, verses
+from twelve_days import recite
 
 
-class TwelveDaysTests(unittest.TestCase):
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.1.0
+
+class VerseTests(unittest.TestCase):
     def test_verse1(self):
-        expected = ("On the first day of Christmas my true love gave to me, "
-                    "a Partridge in a Pear Tree.\n")
-        self.assertEqual(verse(1), expected)
+        expected = ["On the first day of Christmas my true love gave to me, "
+                    "a Partridge in a Pear Tree."]
+        self.assertEqual(recite(1, 1), expected)
 
     def test_verse2(self):
-        expected = ("On the second day of Christmas my true love gave to me, "
+        expected = ["On the second day of Christmas my true love gave to me, "
                     "two Turtle Doves, "
-                    "and a Partridge in a Pear Tree.\n")
-        self.assertEqual(verse(2), expected)
+                    "and a Partridge in a Pear Tree."]
+        self.assertEqual(recite(2, 2), expected)
 
     def test_verse3(self):
-        expected = ("On the third day of Christmas my true love gave to me, "
+        expected = ["On the third day of Christmas my true love gave to me, "
                     "three French Hens, "
                     "two Turtle Doves, "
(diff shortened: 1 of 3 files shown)


exercises/twelve-days/twelve_days_test.py
```diff
@@ -128,23 +130,137 @@ def test_verse12(self):
...
+                    "and a Partridge in a Pear Tree."]
+        self.assertEqual(recite(12, 12), expected)
+
 
-    def test_multiple_verses(self):
-        expected = ("On the first day of Christmas my true love gave to me, "
-                    "a Partridge in a Pear Tree.\n\n"
+class LyricsTests(unittest.TestCase):
+    def test_first_three_verses_of_the_song(self):
+        expected = ["On the first day of Christmas my true love gave to me, "
```
cmccandless: Since there are other tests verifying correct output from `recite(n, n)`, I think the following would be adequate:

```Python
expected = [recite(n, n) for n in range(1, 4)]
```

This also removes the need for large redundant text blocks.

exercises/twelve-days/twelve_days_test.py
```diff
@@ -128,23 +130,137 @@ def test_verse12(self):
...
                     "On the third day of Christmas my true love gave to me, "
                     "three French Hens, "
                     "two Turtle Doves, "
-                    "and a Partridge in a Pear Tree.\n\n")
-        self.assertEqual(verses(1, 3), expected)
+                    "and a Partridge in a Pear Tree."]
+        self.assertEqual(recite(1, 3), expected)
+
+    def test_three_verses_from_the_middle_of_the_song(self):
+        expected = ["On the fourth day of Christmas my true love gave to me, "
```
cmccandless: Again, `expected = [recite(n, n) for n in range(4, 7)]`

exercises/twelve-days/twelve_days_test.py
```diff
@@ -128,23 +130,137 @@ def test_verse12(self):
...
+                    "five Gold Rings, "
+                    "four Calling Birds, "
+                    "three French Hens, "
+                    "two Turtle Doves, "
+                    "and a Partridge in a Pear Tree."]
+        self.assertEqual(recite(4, 6), expected)
 
     def test_the_whole_song(self):
-        self.assertEqual(verses(1, 12), sing())
+        expected = ["On the first day of Christmas my true love gave to me, "
```
cmccandless: Again, `expected = [recite(n, n) for n in range(1, 13)]`

exercises/twelve-days/twelve_days_test.py
mRcfps: @cmccandless Pretty nice trick! But since `recite` returns a list of one string, we should capture its first element:

```python
expected = [recite(n, n)[0] for n in range(1, 4)]
```

cmccandless: On the idea of nice tricks...

```Python
expected = sum([recite(n, n) for n in range(1, 4)], [])
```

Not really any neater than what you're already doing, and it's not as readable, so please don't use that for this exercise.

PR #1309, opened 2018-02-17
bracket-push: update tests to v1.2.0

### LINKED ISSUE #1223: bracket-push: update tests to v1.2.0
[Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/bracket-push/canonical-data.json)

mRcfps:
Closes #1223. [Canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/bracket-push/canonical-data.json).

Besides updating version number, I have renamed the function `check_bracket` and its parameter `input_string` to `is_paired` with parameter `value` to be consistent with canonical data.

### DIFF
--- a/exercises/bracket-push/bracket_push.py
+++ b/exercises/bracket-push/bracket_push.py
@@ -1,2 +1,2 @@
-def is_paired(value):
+def is_paired(input_string):
     pass
--- a/exercises/bracket-push/bracket_push.py
+++ b/exercises/bracket-push/bracket_push.py
@@ -1,2 +1,2 @@
-def check_brackets(input_string):
+def is_paired(value):
     pass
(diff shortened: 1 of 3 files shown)


exercises/bracket-push/example.py
```diff
@@ -1,4 +1,4 @@
-def check_brackets(string):
```
cmccandless: I would like to see the parameter name match the one from the stub.

exercises/bracket-push/bracket_push.py
```diff
@@ -1,2 +1,2 @@
-def check_brackets(input_string):
```
cmccandless: I think that `input_string` is more descriptive. To me, `value` is too generic. I would prefer that we leave this as-is.

PR #1325, opened 2018-02-18
word-search: update tests to v1.2.0

### LINKED ISSUE #1241: word-search: update tests to v1.2.0
[Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/word-search/canonical-data.json)

mRcfps:
Closes #1241.

Since [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/word-search/canonical-data.json) was only updated for new input policy, which has nothing to do with test file in Python, so update of version number suffices.

Besides I have removed unnecessary code in the stub to ensure more freedom of implementing this exercise.

### DIFF
--- a/exercises/word-search/word_search.py
+++ b/exercises/word-search/word_search.py
@@ -1,22 +1,6 @@
 class Point(object):
     def __init__(self, x, y):
-        self.x = x
-        self.y = y
-
-    def __repr__(self):
-        return 'Point({}:{})'.format(self.x, self.y)
-
-    def __add__(self, other):
-        return Point(self.x + other.x, self.y + other.y)
-
-    def __sub__(self, other):
-        return Point(self.x - other.x, self.y - other.y)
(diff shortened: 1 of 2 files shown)


exercises/word-search/word_search.py
```diff
@@ -1,27 +1,11 @@
...
-    def __repr__(self):
-        return 'Point({}:{})'.format(self.x, self.y)
-
-    def __add__(self, other):
-        return Point(self.x + other.x, self.y + other.y)
-
-    def __sub__(self, other):
-        return Point(self.x - other.x, self.y - other.y)
-
-    def __eq__(self, other):
```
cmccandless: Typically, `__eq__` is implemented for the user to ensure that tests work correctly. Removing the other methods is fine, but please keep `__eq__` and the body of `__init__`.

PR #1334, opened 2018-02-18
two-bucket: update tests to v1.4.0

### LINKED ISSUE #1250: two-bucket: update tests to v1.4.0
[Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/two-bucket/canonical-data.json)

mRcfps:
Closes #1250. [Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/two-bucket/canonical-data.json).

### DIFF
--- a/exercises/two-bucket/example.py
+++ b/exercises/two-bucket/example.py
@@ -1,14 +1,13 @@
 '''
-    This solution implements a breadth-first search of the graph
+    This solution implements a breadth-start_bucket search of the graph
     of possible valid states for the two buckets until it reaches a state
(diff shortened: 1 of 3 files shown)


exercises/two-bucket/example.py
```diff
@@ -1,14 +1,13 @@
 '''
-    This solution implements a breadth-first search of the graph
+    This solution implements a breadth-start_bucket search of the graph
```
cmccandless: This appears to be an unintentional change from a Replace-All; please revert it.

PR #1343, opened 2018-02-20
Convert tests to ddt (where appropriate)

cmccandless:
Much of the current test suites is identical code repeated for each test case. Converting this repetitive code to using ddt can reduce the size of the test suites, in some cases quite significantly.

I've started this PR by converting a few exercises as an example. If the consensus is that this is a good change, I will identify all exercises that can benefit from using ddt and perform the necessary changes to those as well.

I am particularly interested in the opinions of the other maintainers (@Dog @behrtam @N-Parsons @m-a-ge), but welcome opinions from others as well.


exercises/acronym/acronym_test.py
```diff
@@ -1,26 +1,40 @@
...
+        ),
+        AnnotatedCase(
+            'punctuation without whitespace',
+            'Complementary metal-oxide semiconductor',
+            'CMOS'
+        ),
+    )
+    @unpack
+    def test_abbreviate(self, phrase, expected):
+        self.assertEqual(abbreviate(phrase), expected)
```
cmccandless: Interesting note: this test file actually is actually longer than the original. Clearly test suite size will only reduce where repetitive test code is either several lines long or repeated in more than just a few instances.

mRcfps:
In the README of this repo:

> We use `unittest` (Python Standard Library) and no 3rd-party-framework.

This convention does make sense because it greatly reduces the setup work for a student new to Exercism. With `ddt` involved, we now have to add `pip install ddt` to all introductory guides, and some of our users even don't even know what `pip` is (remember Python has become a popular choice for teaching programming). Moreover, it can be quite confusing for beginners when trying to figure out what test cases are`.

Anyway, I don't think it worthwhile to trade effortlessness of getting started for reduction of the size of test suites.

cmccandless:
Aside from going against current convention (conventions can be changed if the change is unanimously agreed to be for the good of the track), @mRcfps raises a good point regarding simplicity for newcomers to the track. I no longer (as contributor or maintainer) believe that this PR is worth pursuing, and am now closing it.

PR #1347, opened 2018-02-22
food-chain: update tests to v2.1.0

### LINKED ISSUE #1253: food-chain: update tests to v2.1.0
[Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json)

mRcfps:
Closes #1253. [Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json). 

Taken the same testing approach as [twelve-days](https://github.com/exercism/python/blob/master/exercises/twelve-days/twelve_days_test.py).

### DIFF
--- a/exercises/food-chain/food_chain_test.py
+++ b/exercises/food-chain/food_chain_test.py
@@ -1,108 +1,107 @@
 import unittest
 
-from food_chain import chain
-
-# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
-
-song = """I know an old lady who swallowed a fly.
-I don't know why she swallowed the fly. Perhaps she'll die.
-
-I know an old lady who swallowed a spider.
-It wriggled and jiggled and tickled inside her.
-She swallowed the spider to catch the fly.
-I don't know why she swallowed the fly. Perhaps she'll die.
-
-I know an old lady who swallowed a bird.
-How absurd to swallow a bird!
-She swallowed the bird to catch the spider that wriggled
-and jiggled and tickled inside her.
-She swallowed the spider to catch the fly.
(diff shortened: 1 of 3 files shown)


exercises/food-chain/food_chain_test.py
```diff
@@ -1,108 +1,107 @@
...
 
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v2.1.0
+
 class FoodChainTest(unittest.TestCase):
 
     def test_fly(self):
-        self.assertEqual(original[0], generated[0])
+        expected = [
+            "I know an old lady who swallowed a fly."
```
cmccandless: I think within each verse, we ought to keep the newlines (`\n`).

mRcfps: But we seem to have reached a consensus in [beer-song](https://github.com/exercism/python/blob/master/exercises/beer-song/beer_song_test.py), [house](https://github.com/exercism/python/blob/master/exercises/house/house_test.py) and [twelve-days](https://github.com/exercism/python/blob/master/exercises/twelve-days/twelve_days_test.py), where there is no newline at the end of each verse.

cmccandless: Hmm... I obviously didn't pay as much attention to `beer-song` and `house` as I thought. In the case of `twelve-days`, each verse is a single line, is this was not a relevant point.

After considering the format used in the canonical data, I have changed my mind and agree that it should remain this way to be consistent both with the other exercises and with the canonical data. However, this means that `house` is not actually consistent after all, as each verse contains multiple element in the returned list. This should be corrected.

PR #1354, opened 2018-02-26
house: return singleton list for single verse

cmccandless:
RE: #1347 ([comment](https://github.com/exercism/python/pull/1347#discussion_r170626365))


mRcfps:
@cmccandless I've raised exercism/problem-specifications#1200 to address this problem in canonical data.

PR #1361, opened 2018-03-12
Add ValueError test to rna_transcription

mogranjm:
Add a test to assert that a ValueError is raised when an invalid string is submitted for transcription (the invalid string that is tested in this case is 'ABCD', which could conceivably be transcribed to 'U_G_' or similar)


cmccandless:
This case actually used to exist. The decision was made in exercism/problem-specifications#1027 to remove tests for invalid input from this exercise. As such, we will not be restoring it here. I'm going to close this PR.

Feel free to check out our [list of open issues](https://github.com/exercism/python/issues), especially those with the `help wanted` label.

PR #1365, opened 2018-03-20
Verify the ValueError is raised when an improper strand is passed for rna-transcription

wrightling:
The rna-transcription exercise requests that a ValueError be raised if an invalid strand is provided.


cmccandless:
This case actually used to exist. The decision was made in exercism/problem-specifications#1027 to remove tests for invalid input from this exercise. As such, we will not be restoring it here. I'm going to close this PR.

Feel free to check out our list of open issues, especially those with the help wanted label.

PR #1372, opened 2018-04-10
bowling: negative test uses positive value

yawpitch:
Fixes /exercism/python/#1371.


cmccandless:
@yawpitch merged; thanks for working on this!

PR #1379, opened 2018-04-23
series: re-implement according to canonical data

### LINKED ISSUE #1109: Exercise named series:  Poor Test Cases ?

mRcfps:
Close #1109. 
Close #1117. 

[Canonical Data](https://github.com/exercism/problem-specifications/blob/master/exercises/series/canonical-data.json). 
[Description](https://github.com/exercism/problem-specifications/blob/master/exercises/series/description.md).

### DIFF
--- a/exercises/series/README.md
+++ b/exercises/series/README.md
@@ -1,18 +1,18 @@
 # Series
 
 Given a string of digits, output all the contiguous substrings of length `n` in
-that string.
+that string in the order that they appear.
 
 For example, the string "49142" has the following 3-digit series:
 
-- 491
-- 914
-- 142
+- "491"
+- "914"
+- "142"
 
 And the following 4-digit series:
 
-- 4914
-- 9142
+- "4914"
+- "9142"
 
 And if you ask for a 6-digit series from a 5-digit string, you deserve
 whatever you get.
--- a/exercises/series/example.py
+++ b/exercises/series/example.py
@@ -1,6 +1,5 @@
 def slices(series, length):
-    numbers = [int(digit) for digit in series]
-    if not 1 <= length <= len(numbers):
+    if not 1 <= length <= len(series):
         raise ValueError("Invalid slice length for this series: " + str(
             length))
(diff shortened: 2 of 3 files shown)


cmccandless:
@mRcfps merged; thanks for working on this!

PR #1382, opened 2018-04-26
food-chain: Enforce the same output format as canonical data

### LINKED ISSUE #1381: food-chain: tests do not match canonical data
The [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json) for this exercise expects a list of individual lines for each verse, but the tests are using implicit string concatenation so that the individual lines are joined into a single line. The end result is some non-sensical punctuation where the end of any given line is immediately followed by the first character of the next without whitespace.

mRcfps:
Closes #1381. Use a list of individual lines as expected output instead of implicit string concatenation.

### DIFF
--- a/exercises/food-chain/example.py
+++ b/exercises/food-chain/example.py
@@ -50,5 +50,14 @@ def verses(letter):
 
 
 def recite(start_verse, end_verse):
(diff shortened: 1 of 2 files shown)


cmccandless:
@mRcfps merged; thanks for working on this!

PR #1388, opened 2018-05-03
markdown: unordered lists test of inital implementation fails when it should pass

### LINKED ISSUE #1380: markdown: unordered lists test of inital implementation fails when it should pass.
The test for unordered lists was updated in #975 so that the expected output of the implementation **should not** include \<p> tags, which reflects the current canonical data.

While the tests were updated, the default implementation provided in markdown.py was not also updated to reflect that change, and so the unordered lists test fails after a clean fetch.

This is confusing, since the description of this particular exercise indicates that all tests should be passing on the provided implementation, the intent of the exercise being that the user refactors that "working" code into an actually readable version.

cmccandless:
Fixes #1380

### DIFF
--- a/exercises/markdown/markdown.py
+++ b/exercises/markdown/markdown.py
@@ -29,28 +29,24 @@ def parse_markdown(markdown):
                     curr = m1.group(1) + '<em>' + m1.group(2) + \
                         '</em>' + m1.group(3)
                     is_italic = True
-                if is_italic or is_bold:
-                    i = '<ul><li>' + curr + '</li>'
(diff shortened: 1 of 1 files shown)


cmccandless:
@yawpitch, as you were the one who raised the initial issue and were concerned with retaining the impracticality of the initial implementation, I welcome your thoughts on this solution.

yawpitch:
Away from my machine for approximately 24 more hours, so cannot make a meaningful comment right now. Will replay ASAP.

M

PR #1402, opened 2018-06-12
rectangles: fix TestCase name

cmccandless:


cmccandless:
Overlapped by #1403; closing.

PR #1410, opened 2018-06-25
Make example and stub implementation line up with tests

NobbZ:
The old version of those files still expected an argument, which was used in very early versions of the exercise to create dynamic output.
But as we are only testing for a single static return value, namely `"Hello, World!"`, it is safe to remove that argument from example and stub.


NobbZ:
Feel free to squash-merge. I do not need any evidence in the history that I'm doing to many languages with implicit return ;)

cmccandless:

PR #1412, opened 2018-06-28
reverse-string: Don't redefine builting function 'input'

frerich:
It hurts readability to redefine the meaning of well-known builtins such
as 'input', and pylint rightfully warns about this, saying

  W:  1,12: Redefining built-in 'input' (redefined-builtin)

It's a little unfortunate that Python grabs this part of the namespace,
but let's fix the issue by using a different word for 'the input text'.


frerich:
@cmccandless I just amended the commit in this PR as well as PR #1413 such that the `example.py` files are updated as well. I also edited the PR title to include the name of the exercise being modified.

cmccandless:

PR #1413, opened 2018-06-28
linked-list: Don't redefine builting function 'next'

frerich:
It hurts readability to redefine the meaning of well-known builtins such
as 'next', and pylint rightfully warns about this, saying

  W:  2,30: Redefining built-in 'next' (redefined-builtin)

It's a little unfortunate that Python grabs this part of the namespace,
but let's fix the issue by using slightly different words for 'the item
before' and the 'the item next'.


cmccandless:

PR #1396, opened 2018-05-30
rna-transcription: Removed HINTS.md ValueError reference

davepeake:
The README for the python "rna_transcription" exercise reads:

Your function will need to be able to handle invalid inputs by raising a `ValueError` with a meaningful message.

The test suite doesn't check for this currently, so I've added a test case where it checks for bad bases which are any capital letters that are not G,C,T,A. There is probably room for another 3 - 4 tests; testing lower case, testing numbers, testing punctuation.

PR #1423, opened 2018-07-19
added tests that method should raise ValueError

ydawant:
In the problem description I was seeing on exercism.io, there was a blurb about needing to make sure the function can handle invalid inputs, so I added some tests for that.


ydawant:
closing this, seeing that the removal of test cases against this is was part of: https://github.com/exercism/python/pull/1396, so it's intended :)

The solution that I looked at is from 5 days, so before these changes went in.

PR #1424, opened 2018-07-19
reverse-string: do not use default mutable argument

dmerejkowsky:
This can leads to confusing behavior, especially for beginners.

More info here:
https://docs.python-guide.org/writing/gotchas/

Note: there may be other instances of this anti-pattern in the exercises


dmerejkowsky:
> strings are not mutable objects in Python.

Woops, my bad :)

PR #1430, opened 2018-07-26
added testcase to allergies checking is_allergic_to

rachelhornung:
One of the solutions I mentored returned True for all allergies that had a score lower than the total test score. This test should catch this


rachelhornung:
Ah, ok. Still learning my way around here. I will go ahead and suggest the update for the canonical data then.

PR #1435, opened 2018-08-03
complex-numbers: don't require user implementation to round values

### LINKED ISSUE #1429: complex-numbers machine-precision friendly unittest
Unittest must be changed in order to take into account machine precision? or there is something wrong with my machine?

- __OS:__ _Ubuntu 16.04 x64_
- __Python:__ _3.5.2_

Two methods fail for me: __exp__ and __truediv__.
Here is my implementation of __Division__ and __Exponentiation__:
```
# Get cartesian representation from polar representation
    @staticmethod
    def cartesian_from_polar( r, theta ):
        """Given z = r * exp( i * theta ), calculates
        z = real + i * imaginary
        """
        # Cartesian Representation
        real = r * math.cos(theta)
        imaginary = r * math.sin(theta)
        # Return
        return( real, imaginary )
# Division
def __truediv__(self, other):
        """Complex Number Division
        z = r * exp( i * theta ), w = r' * exp( i * theta' ) =>
        z / w = ( r/r' ) * exp( i*(theta - theta') )
        """
        # Calculate Components
        r = self.r / other.r
        theta = self.theta - other.theta
        real, imaginary = ComplexNumber.cartesian_from_polar( r, theta )
        # Return new instance
        return( ComplexNumber( real, imaginary ) )
# Exponentiation
def exp(self):
        """Complex Number Exponentiation
        z = a + i * b =>
        exp(z) = exp(a) * exp( i * b )
        """
        # Calculate Components
        r = math.exp(self.real)
        theta = self.imaginary
        real, imaginary = ComplexNumber.cartesian_from_polar( r, theta )
        # Return new instance
        return( ComplexNumber( real, imaginary ) )
```
here is the failed unittest:
```
...........F..F.F..............
FAIL: test_divide_numbers_with_real_and_imaginary_part (__main__.ComplexNumbersTest)
  File "complex_numbers_test.py", line 114, in test_divide_numbers_with_real_and_imaginary_part
    self.assertEqual(input_number / divider, expected)
AssertionError: 0.44 +i*0.07999999999999999 != 0.44 +i*0.08

FAIL: test_eulers_identity_formula (__main__.ComplexNumbersTest)
  File "complex_numbers_test.py", line 156, in test_eulers_identity_formula
    self.assertEqual(input_number.exp().imaginary, expected.imaginary)
AssertionError: 1.2246467991473532e-16 != 0

FAIL: test_exponential_of_a_number_with_real_and_imaginary_part (__main__.ComplexNumbersTest)
  File "complex_numbers_test.py", line 174, in test_exponential_of_a_number_with_real_and_imaginary_part
    self.assertEqual(input_number.exp().imaginary, expected.imaginary)
AssertionError: 2.4492935982947064e-16 != 0
```

cmccandless:
Resolves #1429

### DIFF
--- a/exercises/complex-numbers/complex_numbers_test.py
+++ b/exercises/complex-numbers/complex_numbers_test.py
@@ -152,26 +152,30 @@ class ComplexNumbersTest(unittest.TestCase):
     def test_eulers_identity_formula(self):
         input_number = ComplexNumber(0, math.pi)
         expected = ComplexNumber(-1, 0)
-        self.assertEqual(input_number.exp().real, expected.real)
-        self.assertEqual(input_number.exp().imaginary, expected.imaginary)
+        actual = input_number.exp()
+        self.assertAlmostEqual(actual.real, expected.real)
+        self.assertAlmostEqual(actual.imaginary, expected.imaginary)
 
     def test_exponential_of_0(self):
         input_number = ComplexNumber(0, 0)
         expected = ComplexNumber(1, 0)
-        self.assertEqual(input_number.exp().real, expected.real)
-        self.assertEqual(input_number.exp().imaginary, expected.imaginary)
+        actual = input_number.exp()
+        self.assertAlmostEqual(actual.real, expected.real)
+        self.assertAlmostEqual(actual.imaginary, expected.imaginary)
 
     def test_exponential_of_a_purely_real_number(self):
         input_number = ComplexNumber(1, 0)
         expected = ComplexNumber(math.e, 0)
-        self.assertEqual(input_number.exp().real, expected.real)
-        self.assertEqual(input_number.exp().imaginary, expected.imaginary)
+        actual = input_number.exp()
+        self.assertAlmostEqual(actual.real, expected.real)
+        self.assertAlmostEqual(actual.imaginary, expected.imaginary)
 
     def test_exponential_of_a_number_with_real_and_imaginary_part(self):
         input_number = ComplexNumber(math.log(2), math.pi)
         expected = ComplexNumber(-2, 0)
-        self.assertEqual(input_number.exp().real, expected.real)
-        self.assertEqual(input_number.exp().imaginary, expected.imaginary)
+        actual = input_number.exp()
+        self.assertAlmostEqual(actual.real, expected.real)
+        self.assertAlmostEqual(actual.imaginary, expected.imaginary)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


cmccandless:
@SinRas

SinRas:
@cmccandless great! keep up the good work! 👍 👍 

as i checked the changes in https://github.com/exercism/python/pull/1435/files , only the tests for `exp` are changed to `assertAlmostEqual`.

__P.S:__ https://github.com/exercism/python/issues/1429#issue-344613750 as mentioned in the comment, isn't the same `assertAlmostEqual` necessary for __all other types of operations (e.g. `__truediv__`)__ or am i being too obsessive with math? ❓

__P.P.S:__ i know that introducing more changes is in contrast with the [Three Rules of TDD](http://butunclebob.com/ArticleS.UncleBob.TheThreeRulesOfTdd) since there is no error now, i guess 😄

SinRas:
> @cmccandless I am aware of the fact that this is a programming project and not a math one :smile: 

and also i won't be taking too much time from you after this comment (don't wanna be an annoying user:D)

Here are the reasons that i would like to present to back the idea of _all operations should be checked with `assertAlmostEqual`_:
- i know that machine precision is finite, so each operation introduces some error. this directly depends on the type of operation, e.g. some introduce more error than others. For example addition and multiplication (in general) introduce less error that exponentiation and taking the square root.
- [Complex Numbers][1] can be represented in two common forms [Cartesian Form][2] and [Polar Form][3]
- Some operations are easier to implement in __Cartesian Form__ e.g. [__Conjugate__][4], [__Addition/Subtraction__][5]
- Some operations are easier to implement in __Polar Form__, e.g. [__Multiplication/Division__][7] and [__Absolute Value__][8]
- Some operations bridge between two representations, e.g. [__Exponentiation__][6]
- Suppose that a complex number is represented with two classes:

different operations are shown below:
<ul>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
<li></li>
</ul>

Now in order to be able to switch between representations (to implement all operations for the class `ComplexNumber`), we can use:


Theoretically, these transformations are correct. Yet practically, they introduce errors of magnitude `1e-11` which makes the `assertEqual` fail.


[1]: https://en.wikipedia.org/wiki/Complex_number
[2]: https://en.wikipedia.org/wiki/Complex_number#Cartesian_form_and_definition_via_ordered_pairs
[3]: https://en.wikipedia.org/wiki/Complex_number#Polar_form
[4]: https://en.wikipedia.org/wiki/Complex_number#Conjugate
[5]: https://en.wikipedia.org/wiki/Complex_number#Addition_and_subtraction
[6]: https://en.wikipedia.org/wiki/Complex_number#Exponentiation
[7]: https://en.wikipedia.org/wiki/Complex_number#Multiplication_and_division_in_polar_form
[8]: https://en.wikipedia.org/wiki/Complex_number#Absolute_value_and_argument

PR #1436, opened 2018-08-04
Add 'math' topic to relevant exercises in track config

kytrinyx:
The new site lets people filter optional exercises by topic, and this
makes it easy for people to skip past them.

This is important since the math-y exercises are not key to learning
the language, and a lot of people find them intimidating and demotivating.

This updates any 'mathematics' topics to be 'math' since it's shorter
and won't get cropped in the UI.

It also removes the math topics for any exercises that are not in the
canonical math-y list.

exercism/exercism.io#4110

### DIFF
--- a/config.json
+++ b/config.json
@@ -202,7 +202,7 @@
       "topics": [
         "algorithms",
         "loops",
-        "mathematics"
+        "math"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -640,7 +636,6 @@
         "conditionals",
         "lists",
         "loops",
-        "mathematics",
```
cmccandless: I would argue that this exercise is still relevant to the `math` topic.

cmccandless: After reviewing discussion in exercism/exercism.io#4110, I am OK with removing this topic from this exercise.

PR #1449, opened 2018-08-07
saddle-points: Add tests for non-square matrices

Alexhans:
The saddle-points exercise made no indication that it was just for NxN matrices but tests for NxM matrices were non-existent.

This pull request depends on the successful merge of [exercism/problem-specifications/pull/1288](https://github.com/exercism/problem-specifications/pull/1288)

### DIFF
--- a/exercises/saddle-points/saddle_points_test.py
+++ b/exercises/saddle-points/saddle_points_test.py
@@ -10,7 +10,7 @@ import unittest
 from saddle_points import saddle_points
 
 
-# Tests adapted from `problem-specifications//canonical-data.json` @ v1.2.0
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.3.0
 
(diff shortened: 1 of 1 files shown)


exercises/saddle-points/saddle_points_test.py
```diff
@@ -20,6 +20,10 @@ def test_identify_single_saddle_point(self):
     def test_empty_matrix_has_no_saddle_points(self):
         self.assertEqual(saddle_points([]), set())
 
+    def test_matrix_with_one_elem_has_single_saddle_point(self):
+        matrix = [[1]]
+        self.assertEqual(saddle_points(matrix), set([(0, 0)]))
+
```
cmccandless: This test is not canonical, but I think it's a good one. Might not be a bad idea to create another PR in problem-specifications to add it there too, but I don't think this PR needs to be held up waiting on that.

PR #1457, opened 2018-08-15
Add a test case to prevent favoring the first books

valrus:
The previously existing `test_two_groups_of_4...` had multiple copies of the
first few books, which allowed a solution that greedily added books in order to
pass the tests despite not handling all cases correctly. This is one of those
cases, added to prevent that.

### DIFF
--- a/exercises/book-store/book_store_test.py
+++ b/exercises/book-store/book_store_test.py
@@ -30,6 +30,9 @@ class BookStoreTest(unittest.TestCase):
     def test_two_groups_of_4_is_cheaper_than_group_of_5_plus_group_of_3(self):
         self.assertEqual(calculate_total([1, 1, 2, 2, 3, 3, 4, 5]), 5120)
 
+    def test_two_groups_of_4_is_best_without_favoring_first_books(self):
+        self.assertEqual(calculate_total([1, 1, 2, 3, 4, 4, 5, 5]), 5120)
+
     def test_group_of_4_plus_group_of_2_is_cheaper_than_2_groups_of_3(self):
         self.assertEqual(calculate_total([1, 1, 2, 2, 3, 4]), 4080)
 


valrus:
Ah, I saw the canonical-data thing but it's mentioned only obliquely in [Python's README](https://github.com/exercism/python/blob/master/README.md). I was looking for it under the "Testing" heading.

Thanks for the pointers - I'll follow the procedures you mentioned.

Ornataweaver:
Can you explain an example of the greedy solution?
How would [1, 1, 2, 3, 4, 4, 5, 5] go wrong?

valrus:
Yes, sorry. That was bad wording. I meant "...allowed a solution that greedily added books _in the order in which they appear in the input list_ to pass the tests". Let me give an example.

For the existing test case `[1, 1, 2, 2, 3, 3, 4, 5]`, you can form the correct groups of 4 by going left to right through the list and adding available books as you find them. On the first pass you'd create the group `[1, 2, 3, 4]` and on the second pass you'd make `[1, 2, 3, 5]`. I saw two solutions in a row on the Python track that solved the problem this way.

But for the new test case `[1, 1, 2, 3, 4, 4, 5, 5]`, this doesn't work. On the first pass you'd create the group `[1, 2, 3, 4]` and then have `[1, 4, 5, 5]` left over. There's no way to create another group of 4 from that. But you can create two groups of four if you don't just add books to groups from right to left: `[1, 2, 4, 5]` and `[1, 3, 4, 5]`. Neither of those two solutions mentioned above did this correctly.

It was pointed out in my PR on the problem-specifications repo that there's an open [issue](https://github.com/exercism/problem-specifications/issues/1215) for this, which has more investigation if you want it.

cmccandless:
As https://github.com/exercism/problem-specifications/pull/1297 has been merged, this test case addition is good to go, but the example solution still needs updated.

cmccandless:
Issue resolved in #1478; closing this PR.

PR #1477, opened 2018-08-20
Move explainaition from the test file to the README.md

TiDreyer:
The introduction of the "Pythagorean Triplet" side exercise in the python track does not reflect the requirements of the unit tests. There was a comment in the unit test file explaining the exercise better, I moved it to the introduction.

### DIFF
--- a/exercises/pythagorean-triplet/README.md
+++ b/exercises/pythagorean-triplet/README.md
@@ -13,9 +13,36 @@ For example,
 3**2 + 4**2 = 9 + 16 = 25 = 5**2.
 ```
 
-There exists exactly one Pythagorean triplet for which a + b + c = 1000.
+The test cases assume two functions are defined:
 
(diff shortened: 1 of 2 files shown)


cmccandless:
Hi, thanks for the PR!

The way we have things structured here, exercise READMEs are generated using [configlet](https://github.com/exercism/configlet), from the [global exercise description](https://github.com/exercism/problem-specifications/blob/master/exercises/pythagorean-triplet/description.md) and from the [track README template](https://github.com/exercism/python/blob/master/config/exercise_readme.go.tmpl). In this case, I would recommend checking out https://github.com/exercism/problem-specifications and submitting a PR there. If your suggested changes are merged, we can then merge here.

Also, it appears that https://github.com/exercism/problem-specifications/issues/1211 is related to this; I suggest catching up on the conversation there before creating a PR in that repository.

PR #1516, opened 2018-10-01
forth: Refactor duplicate utility functions

### LINKED ISSUE #1514: Implement utility functions only once in long tests?
Hello, this is my first post :)

In long test files with multiple test classes, like [forth_test.py](https://github.com/exercism/python/blob/master/exercises/forth/forth_test.py), the same utility functions,  `setUp` and `assertRaisesWithMessage`, are redefined for each class (9 times in Forth). Perhaps these tests could have a utility class that inherits from `unittest.TestCase`, where these functions would be defined? The test classes could then inherit from the utility class and use the functions.

In the case of `forth_test.py` this reduced file length from 384 to 307 lines. It also makes it easier for user to comment some of the tests out, without breaking previous tests by mistakenly commenting their utility functions.

Would this complicate anything I have not taken into account? If this is to be implemented, should it be done manually for particularly long tests, or perhaps written as a  template for the [tests generator](https://github.com/cmccandless/python/tree/jinja2-generator/generators)?

I would be happy to hear from you.

Here is an example:

``` forth_test.py
class ForthUtilities(unittest.TestCase):
    # Utility functions
    def setUp(self):
        try:
            self.assertRaisesRegex
        except AttributeError:

    def assertRaisesWithMessage(self, exception):
        return self.assertRaisesRegex(exception, r".+")
    

class ForthParsingTest(ForthUtilities):
    def test_numbers_just_get_pushed_to_stack(self):
        input_data = ["1 2 3 4 5"]
        expected = [1, 2, 3, 4, 5]
        self.assertEqual(evaluate(input_data), expected)


class ForthAdditionTest(ForthUtilities):
    def test_can_add_two_numbers(self):
        input_data = ["1 2 +"]
        expected = [3]
        self.assertEqual(evaluate(input_data), expected)

# and so on...
```

omer-g:
Define utility functions in a utility class that inherits
from 'unittest.TestCase'. The test classes inherit from the utility
class.

Reduces line count from 384 to 307. It should also make it easier
for a user to comment out tests without mistakenly breaking
previous tests by commenting out their utility functions.

Resolves #1514

### DIFF
--- a/exercises/forth/forth_test.py
+++ b/exercises/forth/forth_test.py
@@ -5,14 +5,26 @@ from forth import evaluate, StackUnderflowError
 
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.6.0
 
-class ForthParsingTest(unittest.TestCase):
+class ForthUtilities(unittest.TestCase):
+    # Utility functions
+    def setUp(self):
+        try:
+            self.assertRaisesRegex
+        except AttributeError:
+            self.assertRaisesRegex = self.assertRaisesRegexp
+
+    def assertRaisesWithMessage(self, exception):
+        return self.assertRaisesRegex(exception, r".+")
+
+
+class ForthParsingTest(ForthUtilities):
     def test_numbers_just_get_pushed_to_stack(self):
         input_data = ["1 2 3 4 5"]
         expected = [1, 2, 3, 4, 5]
         self.assertEqual(evaluate(input_data), expected)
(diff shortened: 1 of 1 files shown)


cmccandless:
> Resolves: 1514

Correct formatting for automatic linking of issues is: `Resolves #1514`. Please change accordingly.

cmccandless:

PR #1535, opened 2018-10-01
wordy: update tests to v1.2.0

### LINKED ISSUE #1530: wordy: update tests to v1.2.0
The [test suite](https://github.com/exercism/python/blob/master/exercises/wordy/wordy_test.py) for wordy is out of date and needs updated to conform to the [latest canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/wordy/canonical-data.json).

fortrieb:
Update wordy test to 1.2.0. Fix #1530

### DIFF
--- a/exercises/wordy/wordy_test.py
+++ b/exercises/wordy/wordy_test.py
@@ -2,8 +2,8 @@ import unittest
 
 from wordy import calculate
 
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.2.0
 
-# Tests adapted from `problem-specifications//canonical-data.json` @ v1.1.0
 
 class WordyTest(unittest.TestCase):
     def test_addition(self):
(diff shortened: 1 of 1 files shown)


exercises/wordy/wordy_test.py
```diff
@@ -51,12 +51,14 @@ def test_multiple_division(self):
             calculate("What is -12 divided by 2 divided by -3?"), 2)
 
     def test_unknown_operation(self):
-        with self.assertRaisesWithMessage(ValueError):
+        with self.assertRaisesWithMessage(ValueError) as error:
```
cmccandless: In this track, we don't test for actual exception message content, only that one is present. The reasoning for this is summarized well here: https://github.com/exercism/python/issues/1080#issuecomment-342013223

fortrieb: Ok. Well should i check that the error message is not empty? Or is raising an error enough?

cmccandless: `assertRaisesWithMessage` checks for a non-null, non-empty error message.

cmccandless:

PR #1540, opened 2018-10-01
pythagorean-triplet: Update exercise to v1.0.0

### LINKED ISSUE #1510: pythagorean-triplet: update tests to v1.0.0
The [test suite](https://github.com/exercism/python/blob/master/exercises/pythagorean-triplet/pythagorean_triplet_test.py) for pythagorean-triplet is out of date and needs updated to conform to the [latest canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/pythagorean-triplet/canonical-data.json).

omer-g:
Modify tests to fit latest canonical data. Add additional function to
find triplets with sum. Remove primitive triplets function from
exercise. Replace example.py with @cmccandless solution (with
permission). New solution runs considerably faster.

Brute force solutions that do not use primitive triplets will now
take very long. Perhaps hint should be added. This could be discussed
in problem-specifications.

Resolves #1510

### DIFF
--- a/exercises/pythagorean-triplet/example.py
+++ b/exercises/pythagorean-triplet/example.py
@@ -1,6 +1,3 @@
-from math import gcd
-
-
 def triplets_in_range(min, max):
     result = set([x for b in
                   range(4, max + 1, 4) for x in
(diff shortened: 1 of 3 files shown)


cmccandless:

PR #1547, opened 2018-10-02
[WIP] Add Python test generator

yawpitch:
First pass that actually functions; there's definitely going to be some need for additional passes at the work of converting the canonical input dict to arguments that can be passed to Python functions; the canonical data is all over the place from exercise to exercise, and simply treating the dict as keyword arguments doesn't work well because quite a few exercises use keys that are legal dict keys but illegal keyword arguments (ie numbers, language reserved words).

But I think this is a decent start for Hacktoberfest.


generators/generate.py
```diff
@@ -0,0 +1,206 @@
...
+
+def get_properties(data):
+    """
+    Get the list of the unique property values accessed in all tests.
+    """
+
+    def _get_properties(obj):
+        if "cases" in obj:
+            for case in obj["cases"]:
+                yield from get_properties(case)
```
cmccandless: [Travis-CI build](https://travis-ci.org/exercism/python/jobs/436265795) fails because this syntax is not valid in Python 2.7.

yawpitch:
The `yield from` has been replaced with 2.7 compatible form.

generators/generate.py
```diff
@@ -0,0 +1,207 @@
...
+def format_expect(case):
+    """
+    Format the case's expected values for use in a function call.
+    """
+    if is_error(case):
+        return repr(case["expected"]["error"])
+    return repr(case["expected"])
+
+
+def main():
```
cmccandless: I highly recommend the following modification:

```Python
def main(args=None):  # where args may be a list of str
...
	parser.parse_args(args)  # uses sys.argv if args is None
```

This will make automating testing of generated test suites much easier.

generators/templates/_default.j2
```diff
@@ -0,0 +1,8 @@
+{% from '__macros.j2' import make_import, make_tests %}
+import unittest
+{{ make_import(data) }}
+
+# Tests adapted from `problem-specifications//canonical-data.json` @ {{ data.version }}
```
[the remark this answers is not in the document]
yawpitch: Good catch, fixed and pushed.

generators/generate.py
yawpitch: Interesting, hadn't thought of that use case. Fixed and pushed.

cmccandless:
As is, the following exercises generate passing test suites:
- hello-world,
- reverse-string,
- isogram,
- pangram,
- rna-transcription,
- atbash-cipher,
- acronym,
- flatten-array,
- pig-latin,
- scrabble-score,
- robot-name,
- rail-fence-cipher,
- bracket-push,
- dot-dsl,
- tree-building,
- poker,
- zebra-puzzle,
- simple-linked-list,
- linked-list,
- protein-translation,
- error-handling,
- bank-account,
- ledger

Several others ought to be passing (`leap`, `bob`, etc), but have mis-matching property names (`is_leap_year` vs `leap_year`, etc).

I would say that's a fantastic start!

cmccandless:
> especially since I don't think there's canonical data for, for instance, linked-list or simple-linked-list.

Good point. I ran a pretty naive script to check for passing exercises, so I haven't looked closely at those yet.

Also, I'm for some reason any exercise that has error cases is indenting the test functions an extra level, making them undetectable. I'm not sure where the extra indentation is coming from, but I've confirmed that it doesn't occur if the `if data.has_error` block is not present.

One more thing: the line spacing around the canonical data reference is incorrect: is should be 2 blank lines above, 1 below.

cmccandless:
Here is the updated list of passing exercises, excluding those without canonical data:
- hello-world
- reverse-string
- isogram
- pangram
- rna-transcription
- isbn-verifier
- atbash-cipher
- acronym
- flatten-array
- pig-latin
- scrabble-score
- rail-fence-cipher
- bracket-push
- poker
- zebra-puzzle
- protein-translation

yawpitch:
I'll try and take a look at those two issues tonight; failing that it might have to be a in a day or two.

generators/templates/__macros.j2
```diff
@@ -0,0 +1,53 @@
...
+{% macro add_assert_equal(case) %}
+{%- if not case.input %}
+self.assertEqual({{ case.property | to_snake }}(), {{ case | format_expect }})
+{%- else %}
+self.assertEqual({{ case.property | to_snake }}({{ case | format_input }}), {{ case | format_expect }})
+{% endif %}
+{% endmacro %}
+
+{% macro add_assert_raises(case) %}
+self.assertRaisesRegex({{ case.property | to_snake }}({{ case | format_input }}), {{ case | format_expect }})
```
cmccandless: By [convention](https://github.com/exercism/python/#conventions), this macro should be:
```
with self.assertRaisesRegex({{ case.error_type | to_camel }}):
	{{ case.property | to_snake }}({{ case | format_input }})
```

cmccandless:
> Also, I'm for some reason any exercise that has error cases is indenting the test functions an extra level, making them undetectable. I'm not sure where the extra indentation is coming from, but I've confirmed that it doesn't occur if the if data.has_error block is not present.

See https://github.com/yawpitch/python/pull/1

> One more thing: the line spacing around the canonical data reference is incorrect: is should be 2 blank lines above, 1 below.

I've determined this is caused by the yapf auto-formatting. In particular, the style configuration item `blank_lines_around_top_level_definition=2`. However, other formatting is incorrect if that item is changed to `1`.

generators/templates/__macros.j2
cmccandless: In most cases, `case.error_type` will be `ValueError`. However, there are some exercises that define their own error types. I suspect some sort of file will have to be created that defines exceptions to the "use ValueError" rule (perhaps yaml would be a good choice for this?).

yawpitch: Let me think about that one. Have you got a specific example of an exercise that does this?

cmccandless: https://github.com/exercism/python/blob/master/exercises/hamming/hamming_test.py#L53-L55

cmccandless: I agree that custom error classes are probably not a requirement for this track and should be removed in the process of creating generators. However, I don't think it is correct (or beneficial to those learning to code for the first time) to always require the same error class or, worse, to allow any error to be thrown.

For example, the Pandas library has [an excellent example](https://github.com/pandas-dev/pandas/wiki/Choosing-Exceptions-to-Raise) of the difference between different types of exceptions, and I think we would be doing students a disservice to instill a habit of throwing any old meaningless exception, instead of teaching the habit of putting thought into your exceptions.

yawpitch: I agree, however that might make more sense as an exercise to itself, as the exception hierarchy system in Python is quite dense (and also rather unique) ... like I'd **love** to have something that shows why you *never* `except: do_something()` in production code, but that involves something far more dense than is likely to be able to be defined in  canonical data, much less built via an automated script.

I mean personally I hate what we're doing right now, where we're policing a specific error message (in English), but I cannot see a pathway by which we convert what little we know about the error state being handled in the canonical data to something we could get imported in the template, unless we leave that up to individual, per-exercise template and configuration.

Which I'm not ruling out, it just sounds like a lot of complexity for a relatively small number of known cases.

yawpitch: An exercise that required the student to handle some of the _vast_ range of possible OS error symbols defined in [errno](https://docs.python.org/3.7/library/errno.html) would be brilliant, BTW ... and they're relatively standard, which means there _might_ be both some overlap with other languages and a meaningful way of writing canonical properties that would meaningfully map to meaningful tests that could also be meaningfully generated.

yawpitch: I just said meaningful way too much ... it's quite late here. I'll come back to this tomorrow.

cmccandless: > I mean personally I hate what we're doing right now, where we're policing a specific error message

That's exactly what I'm trying to avoid. All we currently do is ensure that there is indeed a message, and that it is not empty; the Python track does not check for the verbatim error message described in the canonical data.

> Which I'm not ruling out, it just sounds like a lot of complexity for a relatively small number of known cases.

It is; I think it could be done independent of whether the exercise has its own unique template or not. Something like this might be the simplest we could get:

```yml
# generators/errors.yml
forth:
  errors_if_there_is_only_one_value_on_the_stack: IndexError
```

```python
# __macros.j2
{% macro add_assert_raises(case) %}
with self.assertRaisesRegex({{ case | error_type }}):
    {{ case.property | to_snake}}({{ case | format_input }})
{% endmacro %}
```
```python
# generate.py
import yaml
with open('errors.yml') as f:
    errors = yaml.load(f)

def error_type(case):
	case_name = to_snake(case['description'])
	current_exercise = ???
	return errors[current_exercise].get(case_name, 'ValueError')
```

It's still a little messy though, and I agree that it does seem like it might be to much effort for too few use cases.

> An exercise that required the student to handle some of the vast range of possible OS error symbols defined in errno would be brilliant, BTW ... and they're relatively standard, which means there might be both some overlap with other languages and a meaningful way of writing canonical properties that would meaningfully map to meaningful tests that could also be meaningfully generated.

Sounds like a good exercise candidate; you could create an issue in problem-specifications to take the idea further!

cmccandless:
If we're already creating individual directories, then it would be simpler to have just [exercise]/errors.yml and [exercise]/template.j2.

yawpitch:
To clarify, you mean directly in python/exericses/[exercise]?

cmccandless:
Actually, the precedent for track files that aren't user-facing would be python/exercises/[exercise]/.meta/

yawpitch:
Right, seems sensible to expect exercise-specific templates and configuration to be in there. Rather than `errors.yml` I think perhaps a `generate.yml` with errors being a section item within it would be better. That way we can also have a section for import overrides for things like class names that won't be parseable from the canonical data. Could help minimize the need for custom templates.

generators/generate.py
```diff
@@ -177,7 +196,7 @@ def main(args=None):
             LOGGER.warning("No canonical data found for %s: skipping",
                            exercise)
             continue
-        data = json.loads(canonical.read_text(), object_pairs_hook=dict)
+        data = json.loads(canonical.read_text(), object_pairs_hook=OrderedDict)
```
cmccandless: :+1:

yawpitch:
Ok, just did a push that should, in theory, get several more tests working. And there's a *lot* of ones that should be pretty easy fixes, just going to have to figure out which is more important, using the property names from the canonical data or retaining the legacy names that are in the tests.

Then there are a _lot_ of weird edge cases, where inputs and expectations have been put into arbitrary structures. I think we're probably going to have to define a small suite of format functions for those, then use per-exercise configuration to say which one(s) to use. We could also define a name map that says property X == test name Y in there.

generators/generate.py
```diff
@@ -190,14 +209,15 @@ def main(args=None):
...
             out_file.write_text(code)
-            try:
-                FormatFile(
-                    str(out_file),
-                    style_config=args.style,
-                    in_place=True,
-                    verify=True)
-            except SyntaxError:
-                LOGGER.exception("Unable to reformat %s: skipping", exercise)
+            if not args.no_autoformat:
```
cmccandless: :+1: :+1:

cmccandless:
Apologies for the late response. This one slipped between the cracks in my workflow somehow.

> Then there are a lot of weird edge cases, where inputs and expectations have been put into arbitrary structures. I think we're probably going to have to define a small suite of format functions for those, then use per-exercise configuration to say which one(s) to use. We could also define a name map that says property X == test name Y in there.

I think it's perfectly reasonable to have individual templates. How those templates handle the test generation is up to the implementer of that template (and subject to sensible review).

yawpitch:
Agreed, and my turn to apologize, missed this notification in a flood of GitHub emails. Haven't had a lot of time free to work on this one, I'm afraid. Hopefully I can take a longer look at it again soon.

cmccandless:
@yawpitch any movement on this?

yawpitch:
Afraid not. My free time has been limited and what I've had available has been soaked up with mentoring. I've got a few hours today and once I've cleared my backlog of notifications I'll try and take a look.

yawpitch:
Ok I've taken a solid first pass at per-exercise configuration and been able to get a significant chunk of exercises to the point where they're generating what I believe are workable tests. There are in some cases some substantial changes to the format of the test files themselves, as the various implementers took a rather wide range of approaches to the naming of tests and the layout of inputs and outputs ... I've gone for consistency over trying to keep a small diff.

That said the diff will be much smaller if you run `yapf -i` on the existing test file. I just did `find ./exercises -name '*_test.py' | xargs yapf -i` and saved myself a lot of parsing through whitespace.

Things I haven't done: tried to handle any exercise that uses a class instead of functions; these will require their own template, though I've got some ideas about how to handle some of the simpler ones in configuration.

Also haven't tried to deal with any of the many different ways that input arguments have been re-mapped from the canonical form ... several exercises flip positional places, others use keywords derived from the canonical properties, others just make things up ... will need to implement some more flexible ways of handling these as functions that get enabled via configuration.

Overall though should get us a lot closer on a lot more exercises.

.gitignore
```diff
@@ -9,3 +9,6 @@ bin/configlet.exe
 .idea/
 .cache
 __pycache__
+
+# virtual environments
+venv
```
cmccandless: :+1:

exercises/atbash-cipher/.meta/generate.yml
```diff
@@ -0,0 +1,5 @@
+cases:
+  - description: "encode decode"
+    property: "decode"
+    input: 'encode("Testing, 1 2 3, testing.")'
+    expected: "testing123testing"
```
cmccandless: Is this for track-specific cases?

generators/generate.py
```diff
@@ -7,13 +7,19 @@
...
 from string import punctuation, whitespace
 
 # 3rd party imports:
 from jinja2 import Environment, PackageLoader, select_autoescape, TemplateError
 from yapf.yapflib.yapf_api import FormatFile, style
+from yaml import load, dump
+try:
+    from yaml import CLoader as Loader, CDumper as Dumper
+except ImportError:
+    from yaml import Loader, Dumper
```
cmccandless: I'm guessing this is for Python2/3 compatibility. Would you mind adding comments specifying which is which?

cmccandless: Ah, of course. I should've taken that hint from the naming schemes. Still would you mind adding a comment explaining that for easy future reference?

yawpitch: Sure.

cmccandless:
Any updates on this?

This ended up being a much bigger project than I though; thanks for all your work on it so far! And to think I thought we could finished this the last week of September!

yawpitch:
Hi Corey,

Sorry to say I've got no update. I'd like t find time to round off the rough edges, but my sabbatical is coming to a close and I have to concentrate on projects that I need to complete before the work search begins in earnest.

Unfortunately part of that involves getting a new website up and running, and that's soaked up all my time as I had no idea HTML had changed so much since I moved into the back end of VFX. It's still several weeks away from live.

I think the branch is in a useable state for a lot of the simpler exercises, and could at least take the load off a list of white-listed exercises. Would be good to know which ones it's utterly failing on.

Yeah I think we were a bit ambitious on this one. Not if I could dedicate my full time to it, but right now I'm barely able to eke out time for minimal mentoring.

Sorry,

M

cmccandless:
That's quite alright. I don't have a lot of time for it myself, but it would be good if, when I do have some bandwidth, I could contribute directly to the work done so far. If that is fine with you, feel free to add me as a contributor on your fork.

yawpitch:
Happy for you to. I've sent the add request.

yawpitch:
Closing this as it's woefully out of date and we've gone down a different path.

PR #1548, opened 2018-10-03
rational-numbers: fix deprecated gcd warning

omer-g:
Remove deprecated fractions.gcd from example.py
Before, warning was thrown in Travis Python 3.6 tests.

### DIFF
--- a/exercises/rational-numbers/example.py
+++ b/exercises/rational-numbers/example.py
@@ -1,7 +1,5 @@
 from __future__ import division
 
-from fractions import gcd
-
 
 class Rational(object):
     """
@@ -56,3 +54,9 @@ class Rational(object):
 
     def __rpow__(self, base):
         return base ** (self.numer / self.denom)
+
+
+def gcd(a, b):
+    while b:
+        a, b = b, a % b
+    return a
--- a/exercises/rational-numbers/example.py
+++ b/exercises/rational-numbers/example.py
@@ -1,9 +1,9 @@
 from __future__ import division
 
 try:
-	from math import gcd
+    from math import gcd
 except ImportError:
-	from fractions import gcd
+    from fractions import gcd
 
 
 class Rational(object):
--- a/exercises/rational-numbers/example.py
+++ b/exercises/rational-numbers/example.py
@@ -1,23 +1,22 @@
 from __future__ import division
 
+try:
+	from math import gcd
+except ImportError:
+	from fractions import gcd
+
 
 class Rational(object):
-    """
-    Toyish implementation of rational numbers. For production purpose,
-    please use `fractions.Fraction` in standard library instead.
-    """
     def __init__(self, numer, denom):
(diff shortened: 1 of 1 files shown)


exercises/rational-numbers/example.py
```diff
@@ -1,7 +1,5 @@
 from __future__ import division
 
-from fractions import gcd
```
cmccandless: I would recommend not abandoning builtin implementations, since they are known to be efficient (and are occasionally implemented in C, providing better performance potential than any user-defined implementation can provide).

The deprecation warning should able to be resolved with the following snippet:

```Python
try:
	from math import gcd
except ImportError:
	from fractions import gcd
```

omer-g: That seems like a better solution, thanks!
In the current `example.py`,  when math.gcd is used there is a failure in one of the tests - `test_divide_two_negative` (not sure why). My solution passes the tests with `math.gcd`. Rather than trying to modify the current solution, could I suggest the [other solution](https://exercism.io/tracks/python/exercises/rational-numbers/solutions/cb3071d7dbe84854a835639ca996a153) instead?

cmccandless: Am I missing something? Both look the same:

```Python
# Current
def __truediv__(self, other):
     return Rational(self.numer * other.denom, self.denom * other.numer)

# Other
def __truediv__(self, other):
    return Rational(self.numer * other.denom, self.denom * other.numer)

Comment (omer-g) [reply exercises/rational-numbers/example.py]: Yes. Looking into it, I think the problem is with the way the current solution handles negative numbers in "_reduce". The `if` only modifies the case where n is positive and d is negative. The case where both are negative is not treated correctly.

Comment (omer-g) [reply exercises/rational-numbers/example.py]: Ah I figured it out. gcd(-2,-4) works differently. In fractions it returns -2 in math it returns (the expected) 2. Perhaps in the case of `gcd` here, it would be better to go with the user defined implementation, even if it would be slightly slower? It may get confusing to check the different versions when the two gcd functions behave differently

Comment (cmccandless) [reply exercises/rational-numbers/example.py]: > It may get confusing to check the different versions when the two gcd functions behave differently

Or, for safety, just wrap the `gcd` call in `math.abs` (and include a comment saying why).

Comment (omer-g) [reply exercises/rational-numbers/example.py]: Even better, I will try and do that later. Thanks!

File: exercises/rational-numbers/example.py [hunk 2]
```diff
@@ -1,23 +1,22 @@
...
-    def _reduce(self, numer, denom):
-        if numer == 0:
-            n, d = 0, 1
-        else:
-            g = gcd(numer, denom)
-            n, d = int(numer/g), int(denom/g)
-            if n > 0 and d < 0:
-                n, d = -n, -d
-        return n, d
+        # abs used for backward compatibility with fractions.gcd
```
cmccandless: :+1:

cmccandless:

Issue #1550, opened 2018-10-08
sgf-parsing: deprecation warning due to an invalid escape sequence


pranasziaukas:
I am running Python 3.7.0 and `pytest-3.8.2`.

For me, this line
https://github.com/exercism/python/blob/f9029669c93de49bf9f794d674c18e069229b764/exercises/sgf-parsing/sgf_parsing_test.py#L78
returns a warning `DeprecationWarning: invalid escape sequence \]`.

pranasziaukas:
#1551

cmccandless:
Closed by #1551

PR #1551, opened 2018-10-08
Escaped backslash due to `DeprecationWarning`

pranasziaukas:
Added an escaped backslash due to `DeprecationWarning: invalid escape sequence \]`.


pranasziaukas:
This commit fixes #1550

PR #1572, opened 2018-10-28
high-scores: Implement new exercise

behrtam:
Not sure if `@property` is really needed here ... simple getter functions would also do the trick.


config.json
```diff
@@ -121,6 +121,15 @@
 
       ]
     },
+    {
+      "uuid": "574d6323-5ff5-4019-9ebe-0067daafba13",
+      "slug": "high-scores",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
```
cmccandless: Can you please add some topics?

exercises/high-scores/example.py
```diff
@@ -0,0 +1,25 @@
+class HighScores(object):
+    def __init__(self, scores):
+        self.scores = scores
+
+    @property
```
cmccandless: I think traditional methods would be more clear than properties in this instance.

PR #1574, opened 2018-10-30
Flake8 update

behrtam:
Update to the latest flake8 version (close #1571).
- the dependency `pep8` was renamed into `pycodestyle`
- `mccabe` was added as new dependency

The new rule added in flake8 v3.6.0 W605 (https://lintlyci.github.io/Flake8Rules/rules/W605.html) than also had to be fixed.


behrtam:
This also fixes a strange error in the travis builds: https://travis-ci.org/exercism/python/jobs/441000102#L450

requirements-travis.txt
```diff
@@ -1,3 +1,4 @@
-flake8==3.5.0
-pep8>=1.7,<1.7.99
+flake8==3.6.0
+mccabe==0.6.1
+pycodestyle==2.4.0
 pyflakes==2.0.0
```
cmccandless: I believe just having `flake8==3.6.0` should be adequate, as the other three are dependencies of and will be implicitly included with flake8.

behrtam: I remember a time where flake8 was very unspecific about the dependency versions and we had some flaky travis-ci tests. But it looks like they have it pinned in a good way now: https://gitlab.com/pycqa/flake8/blob/master/setup.py#L19-28

So I think you are right and we should just pin the flake8 version.

PR #1590, opened 2018-11-13
custom-set: Corrected Function Names to match Test

### LINKED ISSUE #1589: custom-set: custom_set.py has stubs for wrong methods
In [the solution template](https://github.com/exercism/python/blob/master/exercises/custom-set/custom_set.py), the `union` and `difference` stubs should be `__add__` and `__subtract__` (see [example.py](https://github.com/exercism/python/blob/master/exercises/custom-set/example.py)).

vigneshhari:
Fixes #1589


cmccandless:

PR #1591, opened 2018-11-18
Swap parameter order for Bob unit tests

dstockto:
This makes the error messages more correct and helpful
since the testing framework expects the first parameter
to be "expected" and the second to be actual.

On an initial run, the first test would fail like this:

> Whatever. != None
>
> Expected :None
> Actual   :Whatever.

After this PR, the wording is corrected:

> None != Whatever.
>
> Expected :Whatever.
> Actual   :None

### DIFF
--- a/exercises/bob/bob_test.py
+++ b/exercises/bob/bob_test.py
@@ -7,95 +7,94 @@ from bob import hey
 
 class BobTest(unittest.TestCase):
     def test_stating_something(self):
-        self.assertEqual(hey("Tom-ay-to, tom-aaaah-to."), "Whatever.")
+        self.assertEqual("Whatever.", hey("Tom-ay-to, tom-aaaah-to."))
 
     def test_shouting(self):
-        self.assertEqual(hey("WATCH OUT!"), "Whoa, chill out!")
+        self.assertEqual("Whoa, chill out!", hey("WATCH OUT!"))
 
     def test_shouting_gibberish(self):
-        self.assertEqual(hey("FCECDFCAAB"), "Whoa, chill out!")
+        self.assertEqual("Whoa, chill out!", hey("FCECDFCAAB"))
 
     def test_asking_a_question(self):
         self.assertEqual(
-            hey("Does this cryogenic chamber make me look fat?"), "Sure.")
+            "Sure.", hey("Does this cryogenic chamber make me look fat?"))
 
     def test_asking_a_numeric_question(self):
-        self.assertEqual(hey("You are, what, like 15?"), "Sure.")
+        self.assertEqual("Sure.", hey("You are, what, like 15?"))
 
     def test_asking_gibberish(self):
-        self.assertEqual(hey("fffbbcbeab?"), "Sure.")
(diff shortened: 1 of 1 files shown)


cmccandless:
Hi @dstockto ! Thanks for the PR!

The parameter order used in this track is intentionally `(actual, expected)` in order to match the examples used in the [unittest documentation](https://docs.python.org/3/library/unittest.html#basic-example) (See #440). In particular, this exercise was modified in #452.

What is your use case for getting the above output? When running pytest from command line, I get the following output:

```Bash
$ python3.7 -m pytest -x
============================= test session starts ==============================
platform linux -- Python 3.7.1, pytest-4.0.0, py-1.7.0, pluggy-0.8.0
rootdir: /home/cmccandless/repos/python/exercises/bob, inifile:
collected 25 items                                                             

bob_test.py F

=================================== FAILURES ===================================
________________________ BobTest.test_alternate_silence ________________________

self = <bob_test.BobTest testMethod=test_alternate_silence>

    def test_alternate_silence(self):
>       self.assertEqual(hey("\t\t\t\t\t\t\t\t\t\t"), "Fine. Be that way!")
E       AssertionError: None != 'Fine. Be that way!'

bob_test.py:78: AssertionError
=========================== 1 failed in 0.02 seconds ===========================
```

In my output, there is no mention of "Expected" or "Actual".

dstockto:
Hi @cmccandless - The listing of expected and actual shows up when I run the tests in PyCharm. From further research on with the links you provided and others, it appears that unittest doesn't care about the ordering but perhaps it's the PyCharm runner that cares. In my experience, most other testing frameworks I've run into do put the parameters in expected, actual order.

When I run from the command line I get a result similar to yours. Perhaps it's more of an issue with PyCharm and how they've implemented their test runner.

DrJPizzle:
I think the 'official party-line' on this one is you should be free to use what ever order you want
https://mail.python.org/pipermail/python-dev/2010-December/106954.html
Though where diffs are concerned the report for a failed assertEquals(first, second) reports:
<pre>
- first
+ second
</pre>

I think this fits (expected, actual) better:
<pre>
- expected
+ actual
</pre>

I.e. what you did, not what you have to fix. 
This seems to be the rough consensus too see: https://stackoverflow.com/questions/2404978/why-are-assertequals-parameters-in-the-order-expected-actual

That said, I don't think there is any way in which it's wrong to do it either way around, and if there is already a consensus here...

cmccandless:
I think in this case we are going to continue to match the documentation. As Exercism is often used for learning new languages (or even as a start to programming itself), it is important not to introduce confusion between our content and official documentation.

I do recognize the statement in the mailing (linked above), but as we already have a convention in place (see repository README for the [list of conventions](https://github.com/exercism/python/blob/master/README.md#conventions) we adhere to), I would like to keep this exercise consistent with the others to maintain readability.

cmccandless:
@dstockto Thank you for contributing. If you would like to work on something else, please check our [list of open issues](https://github.com/exercism/python/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc).

PR #1568, opened 2018-10-13
sum-of-multiples: update tests to v1.3.0

patiencedaur:
Fix #1567

PR #1592, opened 2018-11-21
fixes #1567

### LINKED ISSUE #1567: sum-of-multiples: Example does not match canonical data
If we enter 0 as a multiple in the current example of `sum-of-multiples`, a ZeroDivisionError is raised.
[Canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/sum-of-multiples/canonical-data.json) require processing the case when zero is in the list of multiples.
The example file does not take that into account, neither does the tests file.
I came across this issue when fixing the inconsistency of `sum-of-multiples-test.py` with canonical data.

DrJPizzle:


exercises/sum-of-multiples/example.py
```diff
@@ -1,4 +1,4 @@
 def sum_of_multiples(limit, multiples):
     return sum(value for value in range(limit)
                if any(value % multiple == 0
-                      for multiple in multiples))
+                      for multiple in multiples if multiple))
```
cmccandless: While this change technically fixes issue #1567, it would be preferable to also add the new test cases that would cause the old revision to fail. Can you please update `sum_of_multiples_test.py` to the latest canonical data? #1568 is stale, and could be closed if the issue was resolved elsewhere.

exercises/sum-of-multiples/sum_of_multiples_test.py
```diff
@@ -54,6 +54,11 @@ def test_multiples_of_1_up_to_100(self):
     def test_multiples_of_an_empty_list_up_to_10000(self):
         self.assertEqual(sum_of_multiples(10000, []), 0)
 
+    def test_the_only_multiple_of_0_is_0(self):
+        self.assertEqual(sum_of_multiples(1, [0]), 0)
+
+    def test_the_factor_0_does_not_affect_the_sum_of_multiples_of_other_factor(self):
```
cmccandless: You will have to paraphrase this test title.

cmccandless:
This PR is abandoned; closing.

PR #1596, opened 2018-11-26
acronym: update tests to v1.6.0

GascaK:
Updated test case for apostrophes in [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/acronym/canonical-data.json) and updated example.py to correctly solve the new apostrophe test case.

### DIFF
--- a/exercises/acronym/acronym_test.py
+++ b/exercises/acronym/acronym_test.py
@@ -1,6 +1,6 @@
 import unittest
 
-from example import abbreviate
+from acronym import abbreviate
 
 
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.6.0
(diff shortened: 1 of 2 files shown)


exercises/acronym/acronym_test.py
```diff
@@ -31,6 +31,9 @@ def test_consecutive_delimiters(self):
         self.assertEqual(
             abbreviate('Something - I made up from thin air'), 'SIMUFTA')
 
+    def test_apostrophes(self):
+        self.assertEqual(abbreviate('Halley\'s Comet'), 'HC')
```
cmccandless: For readability, can you please use double quotes around the input string to negate the need for escaping the apostrophe?

exercises/acronym/example.py
```diff
@@ -2,5 +2,5 @@
 
 
 def abbreviate(words):
-    regex = '[A-Z]+[a-z]*|[a-z]+'
+    regex = '[A-Z]+[\'a-z]*|[\'a-z]+'
```
cmccandless: Again, switching to double quotes for this string literal removes the need for escaping apostrophes.

GascaK:
Yes, makes sense I will get on that later tonight when I get some free time.

cmccandless:

PR #1599, opened 2018-11-26
pythagorean-triplet: refactor for readability

### LINKED ISSUE #1598: pythagorean-triplet: example.py is almost indecipherable
The example.py for the latest version of `pythagorean-triplet` is laudably fast, but it's an almost unreadable mess, full of deeply nested comprehensions and single letter variables; and frankly it's never a good thing when `yapf` takes several seconds to reformat a file this short.

This needs refactoring and at minimum some comments to  become something more idiomatic.

cmccandless:
...and speed. Turns out, there was a lot of room for improvement. Rough estimate, this solution runs ~15 times faster than the current one.

Resolves #1598

@yawpitch, as you raised #1598, I welcome your input in particular on this one. As always, input from others is also welcome.

### DIFF
--- a/exercises/pythagorean-triplet/example.py
+++ b/exercises/pythagorean-triplet/example.py
@@ -34,11 +34,10 @@ def primitive_triplets(limit):
     """
     for m, n in euclidian_coprimes(limit):
         m2, n2 = m * m, n * n
-        a = m2 - n2
-        b = 2 * m * n
+        a, b, c = m2 - n2, 2 * m * n, m2 + n2
         if a > b:
             a, b = b, a
-        yield a, b, m2 + n2
+        yield a, b, c
 
 
 def is_triplet(x):
--- a/exercises/pythagorean-triplet/example.py
+++ b/exercises/pythagorean-triplet/example.py
@@ -1,37 +1,61 @@
+from math import sqrt, ceil
+
+
 def triplets_in_range(min, max):
-    result = set([x for b in
-                  range(4, max + 1, 4) for x in
-                  primitive_triplets(b)])
-    for x, y, z in set(result):
-        a, b, c = (2 * x, 2 * y, 2 * z)
-        while c <= max:
-            if a >= min:
-                result.add((a, b, c))
-            a, b, c = (a + x, b + y, c + z)
-    return set([(d, e, f) for d, e, f in result if d >= min and f <= max])
-
-
-def maxMin(a, b):
-    return (a, b) if a > b else (b, a)
+    result = []
+    for b in range(4, max + 1, 4):
+        for x, y, z in primitive_triplets(b):
+            a, b, c = (x, y, z)
+            while a < min:
+                a, b, c = (a + x, b + y, c + z)
+            while c <= max:
+                result.append((a, b, c))
+                a, b, c = (a + x, b + y, c + z)
+    return result
 
 
 def gcd(x, y):
+    """gcd(x, 0) = x and gcd(x, y) = gcd(y, x % y).
+    See Euclidean Algorithm
+    (https://en.wikipedia.org/wiki/Greatest_common_divisor#Using_Euclid's_algorithm)
+    for more information
+    """
     while (y != 0):
         x, y = y, x % y
     return x
 
 
 def primitive_triplets(b):
-    if b % 4 != 0:
-        raise ValueError('b must be divisible by 4')
-    b2 = int(b / 2)
-    return set([(x[1], x[0], c) for x, c in
-                [(maxMin(m2 - n2, b), m2 + n2) for m, n, m2, n2 in
-                 [(m, n, int(m * m), int(n * n)) for m, n in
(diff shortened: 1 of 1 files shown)


exercises/pythagorean-triplet/example.py
```diff
@@ -1,37 +1,61 @@
...
+            a, b, c = (x, y, z)
+            while a < min:
+                a, b, c = (a + x, b + y, c + z)
+            while c <= max:
+                result.append((a, b, c))
+                a, b, c = (a + x, b + y, c + z)
+    return result
 
 
 def gcd(x, y):
```
yawpitch: Since you're already using `math.ceil` and `math.sqrt`, why not `math.gcd` as well?

exercises/pythagorean-triplet/example.py
```diff
@@ -1,37 +1,61 @@
...
-            if a >= min:
-                result.add((a, b, c))
-            a, b, c = (a + x, b + y, c + z)
-    return set([(d, e, f) for d, e, f in result if d >= min and f <= max])
+from math import sqrt, ceil
 
 
-def maxMin(a, b):
-    return (a, b) if a > b else (b, a)
+def triplets_in_range(min, max):
```
yawpitch: Ideally you shouldn't shadow the `min` and `max` builtins with these variable names. The empty slug.py file uses **range_start** and **range_end** to avoid that.

exercises/pythagorean-triplet/example.py
cmccandless: Sure, works for me.

exercises/pythagorean-triplet/example.py
cmccandless: Reasonable.

cmccandless:
@yawpitch anything else to add?

yawpitch:
Sorry, I apparently had left a couple of comments in "Pending" and not realized it.

exercises/pythagorean-triplet/example.py
```diff
@@ -1,37 +1,61 @@
...
+    See Euclidean Algorithm
+    (https://en.wikipedia.org/wiki/Greatest_common_divisor#Using_Euclid's_algorithm)
+    for more information
+    """
     while (y != 0):
         x, y = y, x % y
     return x
 
 
 def primitive_triplets(b):
```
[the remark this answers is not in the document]
cmccandless: `m = mn // n` will always guarantee `m > n`, so no sort required there.

The second sort can be replaced with the following:

```Python
a = m2 - n2
b = 2 * m * n
if a > b:
    a, b = b, a
yield a, b, m2 + n2
```

On my machine, this is ~25% faster than the comprehension solution. And yes, I would agree it is more readable.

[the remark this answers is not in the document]
cmccandless: After further testing, a modification of your implementation without the extra function call does not perform noticeably better than the 2-function form, so that's good as-is.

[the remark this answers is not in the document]
yawpitch: Your right, the second sort can be replaced with that ... but honestly I'd just keep the second sort it, it's more readable, and under the hood it's doing essentially the same thing.

exercises/pythagorean-triplet/example.py
```diff
@@ -1,37 +1,44 @@
...
+    (https://en.wikipedia.org/wiki/Pythagorean_triple#Generating_a_triple)
+    for more information
+    """
+    for m, n in euclidian_coprimes(limit):
+        m2, n2 = m * m, n * n
+        a = m2 - n2
+        b = 2 * m * n
+        if a > b:
+            a, b = b, a
+        yield a, b, m2 + n2
```
cmccandless: @yawpitch Picking up where we left off at https://github.com/exercism/python/pull/1599#discussion_r237600191

> keep the second sort it, it's more readable, and under the hood it's doing essentially the same thing.

> ` a, b, c = sorted((m2 - n2, 2 * m * n, m2 + n2))`
> `yield a, b, c`

I disagree. For one, including `c` in the sort is incorrect and redundant. Even excluding `c` from sorting, you end up with:

```Python
a, b = sorted((m2 - n2, 2 * m * n))
yield a, b, m2 + n2
```

...which, while less verbose that the current solution, is IMHO less readable.

yawpitch: It's not really accurate to say that it's incorrect (you end up at a < b < c in all cases) and if I remember Timsort correctly there's no real redundancy (the first two items are swapped if necessary, then the third will be left untouched because it's already in the highest position) ... but I agree yours is clearer than the split sort.

That said, I much prefer to see the **c** actually labelled:

```python
a, b, c = m2 - n2, 2 * m * n, m2 + n2
if a > b:
    a, b = b, a
yield a, b, c
```
... perhaps?

cmccandless: I can live with that. The slight overhead for an additional variable assignment per triplet is negligible, and it does make things much more readable.

PR #1603, opened 2018-11-27
sum-of-multiples: update tests to v1.5.0

GascaK:
Fixes exercism#1580
* Updated 3 test cases to conform to canonical data (reordered list to fit the ascending order convention)
* Updated example to check for ZeroDivisionError (anything <= 0)

### DIFF
--- a/exercises/sum-of-multiples/example.py
+++ b/exercises/sum-of-multiples/example.py
@@ -1,5 +1,4 @@
 def sum_of_multiples(limit, multiples):
-
     return sum(value for value in range(limit)
                if any(value % multiple == 0
                       for multiple in multiples
--- a/exercises/sum-of-multiples/example.py
+++ b/exercises/sum-of-multiples/example.py
@@ -1,10 +1,6 @@
 def sum_of_multiples(limit, multiples):
 
-    # Sanitize in-coming list
-    for i, digits in enumerate(multiples):
-        if digits <= 0:
-            del multiples[i]
-
     return sum(value for value in range(limit)
                if any(value % multiple == 0
-                      for multiple in multiples))
+                      for multiple in multiples
+                      if multiple > 0))
--- a/exercises/sum-of-multiples/example.py
+++ b/exercises/sum-of-multiples/example.py
@@ -1,9 +1,10 @@
 def sum_of_multiples(limit, multiples):
 
-	# Sanitize in-coming list
-	for i, digits in enumerate(multiples):
-		if digits <= 0: del multiples[i]
+    # Sanitize in-coming list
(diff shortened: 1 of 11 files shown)


GascaK:
This should also Fix #1567.
Is there a way to check against the continuous integration (Travis CI) before adding an entire commit?

cmccandless:
Travis runs `flake8` for style checks. It is available through `pip`.

exercises/sum-of-multiples/example.py
```diff
@@ -1,4 +1,10 @@
 def sum_of_multiples(limit, multiples):
+
+    # Sanitize in-coming list
+    for i, digits in enumerate(multiples):
+        if digits <= 0:
+            del multiples[i]
+
     return sum(value for value in range(limit)
                if any(value % multiple == 0
                       for multiple in multiples))
```
cmccandless: Might I suggest the following?

```Python
# Remove added sanitizing code
return sum(value for value in range(limit)
           if any(value % multiple == 0
                  for multiple in multiples
                  if multiple > 0))
```

exercises/sum-of-multiples/sum_of_multiples_test.py
```diff
@@ -54,6 +54,16 @@ def test_multiples_of_1_up_to_100(self):
...
+    def test_multiples_of_0_up_to_1(self):
+        self.assertEqual(sum_of_multiples(1, [0]), 0)
+
+    def test_multiples_of_0_or_3_up_to_4(self):
+        self.assertEqual(sum_of_multiples(4, [0, 3]), 3)
+
+    def test_multiples_of_2_3_5_7_or_11_up_to_10000(self):
+        self.assertEqual(sum_of_multiples(10000,
+                         [2, 3, 5, 7, 11]), 39614537)
+
```
cmccandless: In https://github.com/exercism/problem-specifications/commit/8f8975109653dc798eec74f0f239ddc25af71892#diff-fbfc29c1efc68e4b7c62f38d701c6572, test case names were changed to reflect the significance of each case. Can you please update the test names here as well?

GascaK: No problem.

exercises/sum-of-multiples/example.py
GascaK: I'll change that no problem.

exercises/sum-of-multiples/example.py
```diff
@@ -1,4 +1,6 @@
 def sum_of_multiples(limit, multiples):
+
```
cmccandless: This line may be removed.

cmccandless:

PR #1604, opened 2018-11-28
react: expect_callback_values removed

### LINKED ISSUE #1595: react: expect_callback_values can (should?) be removed
https://github.com/exercism/python/blob/e6ed992777e47da830cba80bbb1a6556c8486fa3/exercises/react/react.py#L16-L17

It appears these are not used by the https://github.com/exercism/python/blob/master/exercises/react/react_test.py . I would thus suggest they be removed.

Nishant23:
Closes #1595


cmccandless:

PR #1615, opened 2018-12-04
sgf-parsing: update tests to 1.2.0

### LINKED ISSUE #1620: sgf-parsing: update tests to v1.1.0
The [test suite](https://github.com/exercism/python/blob/master/exercises/sgf-parsing/sgf_parsing_test.py) for sgf-parsing is out of date and needs updated to conform to the [latest canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/sgf-parsing/canonical-data.json).

elyashiv:
close #1620

### DIFF
--- a/exercises/sgf-parsing/sgf_parsing.py
+++ b/exercises/sgf-parsing/sgf_parsing.py
@@ -21,6 +21,8 @@ class SgfTree(object):
                 return False
         return True
 
+    def __ne__(self, other):
+        return not self == other
 
 def parse(input_string):
     pass
--- a/exercises/sgf-parsing/sgf_parsing.py
+++ b/exercises/sgf-parsing/sgf_parsing.py
@@ -24,5 +24,6 @@ class SgfTree(object):
     def __ne__(self, other):
         return not self == other
 
+
 def parse(input_string):
     pass
--- a/exercises/sgf-parsing/sgf_parsing_test.py
+++ b/exercises/sgf-parsing/sgf_parsing_test.py
@@ -3,7 +3,7 @@ import unittest
 from sgf_parsing import parse, SgfTree
 
 
-# Tests adapted from `problem-specifications//canonical-data.json` @ v1.0.0
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.2.0
 
 class SgfParsingTest(unittest.TestCase):
     def test_empty_input(self):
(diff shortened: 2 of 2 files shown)


cmccandless:
Hi! Thanks for the PR!

I'm not sure what the intent of these changes is... the added test case is not in the [canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/sgf-parsing/canonical-data.json), and the added `__ne__` method is never used.

However, there are [other changes](https://github.com/exercism/problem-specifications/commit/835fe984eeeeac3c81659f4487987a89de6417ee#diff-0c2c40d7925de6b739f0916705e6b6af) in the canonical data that have not been reproduced here. Would you like to work on those?

If so, please make the appropriate changes, then use [magic words](https://help.github.com/articles/closing-issues-using-keywords/) to link this PR to issue #1620.

elyashiv:
The __ne__ function is actually used in the __eq__ function on some cases. I needed to implement it in order to pass the tests.

The test case I added does test a case that is mentioned in the README, but is never tested.

I will add the other tests when I will have time.

elyashiv:
> Interesting that none of the existing tests had an issue with this. What version of Python are you running with?
I'm running python 2.7.15rc1 on Ubuntu. I'm not an expert on python, but it seems to be a difference between python 2.7 and 3 (https://docs.python.org/3/reference/datamodel.html#object.__ne__ and https://docs.python.org/2.7/reference/datamodel.html#object.__ne__)

elyashiv:
> Interesting that none of the existing tests had an issue with this. What version of Python are you running with?

I'm running python 2.7.15rc1 on Ubuntu. I'm not an expert on python, but it seems to be a difference between python 2.7 and 3 (https://docs.python.org/3/reference/datamodel.html#object.__ne__ and https://docs.python.org/2.7/reference/datamodel.html#object.__ne__)

cmccandless:
Just a couple more things:
- As exercism/problem-specifications#1414 adds the test case you are proposing, I would like to wait until that has been merged before merging this.
- Since we are targeting exercism/problem-specifications#1414, please change the referenced canonical data version in `sgf_parsing_test.py` to `v1.2.0`
- Use [magic words](https://help.github.com/articles/closing-issues-using-keywords/) to link this PR to issue #1620 as updating to v1.2.0 will also cover v1.1.0
- Change to PR title to something like: "\<exercise-name>: update tests to \<new version>"

elyashiv:
Thanks for the comments!

You are right this will close #1620 

Is there a way to mark the PR as blocked by a different one?

cmccandless:
> You are right this will close #1620

The magic words must be in the PR description to properly link.

> Is there a way to mark the PR as blocked by a different one?

Not at this time, no. I'll just have to watch the other PR until it is merged (no problem since it is my PR).

cmccandless:

PR #1633, opened 2018-12-16
Copy Ruby Track's core progression

### LINKED ISSUE #1614: Copying the Ruby track structure
Hello @exercism/python 

Python is one of two tracks that we're really struggling to mentor. The main reason seems to be that the structure of the track isn't massively easy for beginners to upramp on, so mentors are having to solve a *lot* of issues on each exercise. 

We're currently working on a project to build a system to make structuring tracks easier. We'd like to work on the Python track for that as soon as it's ready. However, in the meantime, I think we should re-order exercises to be the same as in Ruby, which has had a huge amount of effort put into it in terms of thinking about structure.

If anyone massively disagrees then please speak up. Otherwise we'll do this in the next week or so. If anyone would like to volunteer to do the PR for this, then that would be super helpful, otherwise I'm tentatively assigning this to @kytrinyx .

F3PiX:
This is to get you all started with the rearrangement of the track, as proposed in #1614
Please note that I don't know Python; this was done based on the changes I made to the Ruby track. 
Hope it helps!
 
1. Copy Ruby's core exercise progression as a first step to help decrease the mentoring workload.
2. For easy reference, I put all the core exercises at the top of the file. 
3. Removed Leap as a core exercises, and choose other cores to unlock the side exercises that used to depend on Leap
4. Raindrops was marked 'deprecated', but is part of the Ruby core; I reinstated it for now.
5. In Ruby Difference of Squares and Grain are still cores, but will be changed to sides. As Python didn't have them as core, I left it like that. 
6. In Ruby, Robot Name is a core, but it is not an exercise in Python. To avoid extra work, I added Word Count instead. https://github.com/exercism/python/pull/1633/commits/00d5afa8fe8de8e1c0fe526797d66bd27ab47162
7. Extra cores that are in Python, not in Ruby:
Bob, Allergies, Sum of Multiples, Kindergarten, Grade School, Saddle Points, BinarySearch, List ops, Bookstore, Markdown, Rest-api.
I added commits for each of them, so that you can easily revert them. My changes are based on either considerations I made to the Ruby track, or a rough estimation in regard to the changes in this PR. 
- [ ] Kept as core: Kindergarten Garden; Grade School; Book Store; Markdown.
- [ ] Changed into side: Bob, Allergies, Sum of Multiples, Saddle Points, Binary Search, List Ops; Rest-API.

*Closes #1614*

Maintainers To do :
- [ ]  Squash PR before merging
- [ ]  Ping Jeremy and ask to un-deprecate Python/Raindrops

### DIFF
--- a/config.json
+++ b/config.json
@@ -211,6 +211,17 @@
       ]
     },
     {
+      "slug": "rest-api",
+      "uuid": "2abe6eed-c7e4-4ad7-91e3-778bc5176726",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 8,
+      "topics": [
+        "parsing",
+        "strings"
+      ]
+    },
+    {
       "slug": "book-store",
       "uuid": "4899b2ef-675f-4d14-b68a-1a457de91276",
       "core": true,
@@ -1376,17 +1387,6 @@
       "topics": [
         "globalization",
         "refactoring",
-        "strings"
-      ]
-    },
-    {
-      "slug": "rest-api",
-      "uuid": "2abe6eed-c7e4-4ad7-91e3-778bc5176726",
-      "core": true,
-      "unlocked_by": null,
-      "difficulty": 8,
-      "topics": [
-        "parsing",
         "strings"
       ]
     },
--- a/config.json
+++ b/config.json
@@ -175,6 +175,17 @@
       ]
     },
     {
+      "slug": "markdown",
+      "uuid": "88610b9a-6d3e-4924-a092-6d2f907ed4e2",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 3,
+      "topics": [
+        "refactoring",
+        "text_formatting"
+      ]
+    },
+    {
       "slug": "twelve-days",
       "uuid": "d41238ce-359c-4a9a-81ea-ca5d2c4bb50d",
       "core": true,
@@ -1214,17 +1225,6 @@
         "conditionals",
         "control_flow",
         "loops"
-      ]
-    },
-    {
-      "slug": "markdown",
-      "uuid": "88610b9a-6d3e-4924-a092-6d2f907ed4e2",
-      "core": true,
-      "unlocked_by": null,
-      "difficulty": 3,
-      "topics": [
-        "refactoring",
-        "text_formatting"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -133,6 +133,21 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -20,87 +20,258 @@
...
-      "core": false,
-      "unlocked_by": "hello-world",
-      "difficulty": 1,
+      "slug": "raindrops",
+      "uuid": "82d82e32-cb30-4119-8862-d019563dd1e3",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 2,
+      "topics": null,
+      "deprecated": false
```
cmccandless: `deprecated` has is assumed `false` if not given, so it can be left out.

```suggestion
```

As for reinstating `raindrops`, I have no problem with this. I tracked down when it was deprecated, and since it was a commit made directly to master :angry: (bf9b7804a67dce1bbea263ec2fe5cb1fb4af7141), whatever discussion may have happened to decide that it should be deprecated is totally unlinked from the change; I rather liked that exercise myself.

config.json
```diff
@@ -20,87 +20,258 @@
...
-        "transforming"
+        "parsing"
       ]
     },
     {
-      "slug": "isbn-verifier",
-      "uuid": "7961c852-c87a-44b0-b152-efea3ac8555c",
+      "slug": "rest-api",
+      "uuid": "2abe6eed-c7e4-4ad7-91e3-778bc5176726",
+      "core": true,
```
cmccandless: `rest-api` still has [some sizable issues](); it should probably be a side exercise for now. It might merit promotion when it is more developed.

config.json
cmccandless: @iHiD 
> probably conditionally having an undeprecate method 

Could use just use the explicit `"deprecated": false`? Most exercises leave it out, so then it becomes:

- If `deprecated` not given, assume not previously or currently deprecated
- If `deprecated: true`, deprecate it
- If `deprecated: false`, undeprecate it

iHiD: I think the config pattern is fine (true if true else missing) but we just need some code in the website to check for things going from deprecated to not-deprecated as that's just not a pattern we've implemented on the parser.

F3PiX:
@cmccandless Back to you. Made the changes. :tada:

cmccandless:
I think this looks good! As always, we may find the need for further tweaks down the road, but I think this should help with the difficulty curve on this track.

iHiD:
It seems to have magically done it itself! :)

iHiD:
Thank you both!

PR #1637, opened 2018-12-18
travis: verify READMEs contain latest .meta contents

cmccandless:
Whenever the contents of an exercise's .meta directory have changed, the README for that exercise should be regenerated as well. This change compares the timestamps of the latest commit for the .meta directory and the latest commit for the README and fails the build if the README's timestamp is less than that of the .meta directory.


sshine:
Doesn't configlet generate READMEs with hints in them and exit with a failure if any of the READMEs were updated like this?

cmccandless:
@sshine Just tested that with the latest version of configlet (3.9.2). Generate returns 0 (success) whether the README contents have been modified or not.

PR #1655, opened 2019-01-15
Retree: implement exercise

### LINKED ISSUE #1654: retree: implement exercise
[Description](https://github.com/exercism/problem-specifications/blob/master/exercises/retree/description.md)
[Canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/retree/canonical-data.json)

AnAccountForReportingBugs:
Closes #1654

### DIFF
--- a/config.json
+++ b/config.json
@@ -1413,6 +1413,20 @@
         ]
     },
     {
+      "slug": "retree",
+      "uuid": "18467e72-e3b0-403c-896a-ce337f96e6f7",
+      "core": false,
+      "unlocked_by": "markdown",
+      "difficulty": 7,
+      "topics": [
+        "functional_programming",
+        "pattern_recognition",
+        "recursion",
+        "transforming",
+        "trees"
+      ]
+    },
+    {
       "slug": "accumulate",
       "uuid": "e7351e8e-d3ff-4621-b818-cd55cf05bffd",
       "core": false,
--- a/config.json
+++ b/config.json
@@ -1419,7 +1419,6 @@
       "unlocked_by": "markdown",
       "difficulty": 7,
       "topics": [
-        "functional_programming",
         "pattern_recognition",
         "recursion",
         "transforming",
--- a/exercises/retree/example.py
+++ b/exercises/retree/example.py
@@ -1,4 +1,4 @@
-def treeFromTraversals(preorder, inorder):
+def tree_from_traversals(preorder, inorder):
     if len(preorder) != len(inorder):
         raise ValueError("traversals must have the same length")
     if set(preorder) != set(inorder):
--- a/exercises/retree/example.py
+++ b/exercises/retree/example.py
@@ -1,4 +1,4 @@
-def TreeFromTraversals(preorder, inorder):
+def treeFromTraversals(preorder, inorder):
     if len(preorder) != len(inorder):
         raise ValueError("traversals must have the same length")
     if set(preorder) != set(inorder):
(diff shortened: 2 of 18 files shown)


AnAccountForReportingBugs:
That is what I have so far, I would appreciate some insight as to what I should put for the retree object in the config.json file.

config.json
```diff
@@ -1412,6 +1412,20 @@
...
+      "uuid": "18467e72-e3b0-403c-896a-ce337f96e6f7",
+      "core": false,
+      "unlocked_by": "markdown",
+      "difficulty": 7,
+      "topics": [
+        "functional_programming",
+        "pattern_recognition",
+        "recursion",
+        "transforming",
+        "trees"
```
cmccandless: I'm not sure `functional_programming` applies, as there is nothing about the exercise that *requires* functional programming.

Same with `transforming`, but if you can make a solid case for it, I'm fine with keeping it.

`recursion` is fine; I attempted an iterative solution to this, and it is not trivial, so `recursion` is a good hint towards a solution.

exercises/retree/example.py
```diff
@@ -0,0 +1,17 @@
+def treeFromTraversals(preorder, inorder):
```
cmccandless: Forgive me; I've been bouncing between too many different languages lately. My previous suggestion was incorrect, this should be `snake_case`.

```suggestion
def tree_from_traversals(preorder, inorder):
```

exercises/retree/retree.py
```diff
@@ -0,0 +1,2 @@
+def treeFromTraversals(preorder, inorder):
```
cmccandless:
```suggestion
def tree_from_traversals(preorder, inorder):
```

exercises/retree/retree_test.py
```diff
@@ -0,0 +1,59 @@
...
+        inorder = ["a"]
+        expected = {"v": "a", "l": {}, "r": {}}
+        self.assertEqual(treeFromTraversals(preorder, inorder), expected)
+
+    def test_tree_with_many_items(self):
+        preorder = ["a", "i", "x", "f", "r"]
+        inorder = ["i", "a", "f", "x", "r"]
+        expected = {"v": "a", "l": {"v": "i", "l": {}, "r": {}},
+                    "r": {"v": "x", "l": {"v": "f", "l": {}, "r": {}},
+                          "r": {"v": "r", "l": {}, "r": {}}}}
```
cmccandless: I've been thinking about this formatting a bit... for more complex trees, it is helpful to expand the tree a bit for readability:

```Python
expected = {
  "v": "a",
  "l": {
    "v": "i",
    "l": {},
    "r": {},
  },
  "r": {
    "v": "x",
    "l": {
      "v": "f",
      "l": {},
      "r": {},
    },
    "r": {
      "v": "r",
      "l": {},
      "r": {}
    }
  }
}
```

cmccandless:
> You recommended checking PEP 8. Was that a reference to my being oversmart with abbreviated names, or the CapWords "TreeFromTraversals"? As the former have been fixed to clearer snake_case, and the latter changes to the PEP 8 non-compliant mixed case, as requested.

I believe I intended it towards the latter, although my suggestion indicates I should also re-read that PEP8 :sweat_smile:

config.json
AnAccountForReportingBugs: 1. will remove functional_programming, as I thought I already did when I saw recursion was an option.
2.  I thought "transforming" was appropriate, as you are literally transforming two lists into an object representing a tree (a nested dict in this case).
3.  You can do this iteratively, but, IIRC, that requires a proper setup with a stack and a deque, and I find recursive makes more sense on a first pass if you aren't running the risk of hitting your recursion limit.

cmccandless:

PR #1160, opened 2018-01-10
[WIP]Implement exercise hangman

khoa102:
Fixes #745

Ported from [csharp track](https://github.com/exercism/csharp/blob/f8e46551fe37b7dd4da2ce46586ab6a6f8636154/exercises/hangman/HangmanTest.cs)

PR #1690, opened 2019-02-26
Implement exercise hangman

### LINKED ISSUE #745: hangman: implement exercise

junming403:
Fixes #745 

Solution based on the work of https://github.com/exercism/python/pull/1160

Config file lint with tool `configlet`, shall I keep all the changes in the config?

Ready for review

### DIFF
--- a/config.json
+++ b/config.json
@@ -43,13 +43,13 @@
       ]
     },
     {
-      "uuid": "574d6323-5ff5-4019-9ebe-0067daafba13",
       "slug": "high-scores",
+      "uuid": "574d6323-5ff5-4019-9ebe-0067daafba13",
       "core": true,
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "control-flow (if-else statements)",
+        "control_flow_if_else_statements",
         "sequences",
         "text_formatting"
       ]
@@ -351,7 +351,6 @@
       ]
     },
     {
-      "uuid": "505e7bdb-e18d-45fd-9849-0bf33492efd9",
       "slug": "run-length-encoding",
       "uuid": "505e7bdb-e18d-45fd-9849-0bf33492efd9",
       "core": false,
@@ -1333,6 +1332,18 @@
       ]
     },
     {
+      "slug": "hangman",
+      "uuid": "adad6be5-855d-4d61-b14a-22e468ba5b44",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 5,
+      "topics": [
+        "events",
+        "reactive_programming",
+        "strings"
+      ]
+    },
+    {
       "slug": "custom-set",
       "uuid": "bb07c236-062c-2980-483a-a221e4724445dcd6f32",
       "core": false,
@@ -1402,15 +1413,15 @@
       ]
     },
     {
-        "slug": "dnd-character",
-        "uuid": "58625685-b5cf-4e8a-b3aa-bff54da0689d",
-        "core": false,
-        "unlocked_by": null,
-        "difficulty": 1,
-        "topics": [
-            "randomness",
-            "integers"
-        ]
+      "slug": "dnd-character",
+      "uuid": "58625685-b5cf-4e8a-b3aa-bff54da0689d",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "integers",
+        "randomness"
+      ]
     },
     {
       "slug": "retree",
@@ -1432,8 +1443,8 @@
       "unlocked_by": null,
       "difficulty": 2,
       "topics": [
-          "conditionals",
-          "math"
+        "conditionals",
+        "math"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -1338,8 +1338,6 @@
       "unlocked_by": null,
       "difficulty": 5,
       "topics": [
-        "events",
-        "reactive_programming",
         "strings"
       ]
     },
--- a/exercises/hangman/README.md
+++ b/exercises/hangman/README.md
@@ -1,0 +1,52 @@
+# Hangman
+
+Implement the logic of the hangman game using functional reactive programming.
+
+[Hangman][] is a simple word guessing game.
+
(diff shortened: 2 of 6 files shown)


exercises/hangman/README.md
```diff
@@ -0,0 +1,52 @@
+# Hangman
+
+Implement the logic of the hangman game using functional reactive programming.
+
+[Hangman][] is a simple word guessing game.
+
+[Hangman]: https://en.wikipedia.org/wiki/Hangman_%28game%29
+
```
cmccandless: Since this file is supposed to be generated with `configlet generate`, I'm not comfortable with manually editing it, even to remove inaccurate (to this track) data.

Here are the options I can think of:
- Add a note to this exercise's `.meta/HINTS.md` that tells users to ignore the part about a 3rd-party framework.
- Leave inaccurate data as-is (least preferable)
- Submit a PR to exercism/problem-specifications to remove the reference to a 3rd-party framework, leaving it to individual tracks to add that information (i.e. [C# track does this already](https://github.com/exercism/csharp/blob/3f7ce0626176da8130781dcf8c6484968df8c69f/exercises/hangman/.meta/hints.md))
- Mark this exercise as `foregone` since there is no builtin FRP framework. There is a precedent for this with the `lens-person` exercise.
- *\[Insert other options here]*

config.json
```diff
@@ -1432,8 +1443,8 @@
       "unlocked_by": null,
       "difficulty": 2,
       "topics": [
-          "conditionals",
-          "math"
+        "conditionals",
+        "math"
       ]
     },
     {
```
cmccandless: I'm less concerned with making the additional edits to pass `configlet lint` here than I am with the fact that those violations made it through numerous CI builds without `configlet lint` properly flagging them. It's possible we are not invoking `configlet lint` correctly in Travis.

exercises/hangman/hangman_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+from hangman import Hangman
+
+
+# Tests adapted from csharp//hangman/HangmanTest.cs
+
+class HangmanTests(unittest.TestCase):
+    def test_initially_9_failures_are_allowed(self):
+        game = Hangman('foo')
+        self.assertEqual(game.get_status(), "ongoing")
```
cmccandless: It is preferable to import constants (i.e. `STATUS_ONGOING`, etc) from the solution and let the student decide what those values represent. See [the test suite for `yacht`](https://github.com/exercism/python/blob/cb8a7a7619160456ed6cbe83f18d3f27d7b2985b/exercises/yacht/yacht_test.py) for a good example of this.

exercises/hangman/hangman_test.py
```diff
@@ -0,0 +1,95 @@
...
+    def test_after_10_failures_the_game_is_over(self):
+        game = Hangman('foo')
+
+        for i in range(10):
+            game.guess('x')
+
+        self.assertEqual(game.get_status(), 'lose')
+        try:
+            game.guess('x')
+        except Exception as e:
```
cmccandless: I'm a little surprised `flake8` doesn't complain about this, but it's not considered best practice to `except Exception`; it is better to be more specific about which exception you're expecting. If you really have to, use `except BaseException` as `Exception` is only a subset of `BaseException`.

All that aside, `unittest` has a method for expected exceptions, and this track uses a variant of that:
```
def do_the_thing():
    raise ValueError('bad stuff happened!')

# ...
	def test_should_raise_exception(self):
		with self.assertRaisesWithMessage(ValueError):
            do_the_thing()

    # Utility functions
    def setUp(self):
        try:
            self.assertRaisesRegex
        except AttributeError:

    def assertRaisesWithMessage(self, exception):
        return self.assertRaisesRegex(exception, r".+")
```

A couple of things in there:
- `assertRaisesRegex` is `assertRaisesRegexp` in Python 2, so we check for both
- RE: [track README](https://github.com/exercism/python/blob/054ba4c3ba4fd501129f203a22d7431cb55e465b/README.md#conventions), we don't check for specific error messages here, only that one was provided.

config.json
junming403: Shall I keep those changes or I just change them back?

exercises/hangman/README.md
junming403: Thank you for the feecback, I have added a hint about that.

junming403:
I have updated according to the review, except the part regarding config file. Thank you!

config.json
cmccandless: Keeping them is fine. I was mainly making a point of reference for another issue to point to.

junming403:
Sorry for the careless mistake, I have fixed them, ready for review. Thank you.

junming403:
Hi Sir, Would you mind help me review this again? @cmccandless

cmccandless:
Apologies for the silence.

I've been struggling with whether to include this exercise at all, since we are intentionally ignoring the exercise purpose as stated in its README (FRP). I think for now, we will go ahead and merge it in, with the option to redesign it later if we find a good solution for simulating FRP.

config.json
```diff
@@ -1332,6 +1331,18 @@
...
     },
+    {
+      "slug": "hangman",
+      "uuid": "adad6be5-855d-4d61-b14a-22e468ba5b44",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 5,
+      "topics": [
+        "events",
+        "reactive_programming",
```
cmccandless: Since we are technically not implementing FRP or RP in this implementation.

```suggestion
```

config.json
```diff
@@ -1332,6 +1331,18 @@
...
       ]
     },
+    {
+      "slug": "hangman",
+      "uuid": "adad6be5-855d-4d61-b14a-22e468ba5b44",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 5,
+      "topics": [
+        "events",
```
cmccandless: Doesn't apply to this implementation.

```suggestion
```

junming403:
Thank you for your review, I have updated the config files. 
Is there any other exercises I can help with?

PR #1709, opened 2019-03-06
hello_world.py

davesabra:
Hello World
first program, exercism python track,
sure hope this is the right place to drop it


cmccandless:
Hi! If you are trying to submit a solution, please follow the [CLI setup instructions], which will explain how to use Exercism's command line interface. Then use the following command to submit your solution!:
```
exercism submit /path/to/file [/path/to/file2 ...]
```

[CLI setup instructions]: https://exercism.io/cli-walkthrough

PR #1713, opened 2019-03-07
Update palindrome_products_test.py

AnAccountForReportingBugs:
Changes two tests to assertFactorsEqual in order to allow implementations that don't use lists to pass.


cmccandless:

PR #1714, opened 2019-03-07
Update palindrome_products_test.py Python3 branch

AnAccountForReportingBugs:
Changes two tests to assertFactorsEqual in order to allow implementations that don't use lists to pass.


cmccandless:
There's no need to submit updates like this to the `drop-python2` branch. I'll be performing periodic rebases to keep that branch up-to-date to minimize the massive merge that would otherwise need to happen.

PR #1720, opened 2019-03-13
Remove Python 2.7 reference on README

tqa236:
@cmccandless I deleted all Python 2.7 references in README. Does this look good for you? I can do that for other files too if this works.


tqa236:
Let's close this pull request. I will open a cleaner one with only 1 commit

Issue #1705, opened 2019-03-05
Remove references to Python 2.7 from config/exercise_readme.go.tmpl


cmccandless:
Please target the `drop-python2` branch in your pull request.

cmccandless:
After this change, exercise READMEs will have to be regenerated with `configlet generate` to apply the change there.

cmccandless:
Closed in #1721

Issue #1704, opened 2019-03-05
Remove references to Python 2.7 from README.md


cmccandless:
Please target the `drop-python2` branch in your pull request.

joaoluizn:
I'm going to work over this one.

joaoluizn:
I'm going to work on this now.

Grociu:
Would this include dropping Python 2 references from all the individual `README.md` files in each of the exercises? Or is there some sort of script that constructs those? 
For reference, this is an example of an individual exercises readme Python 2.7 reference:

> ## Running the tests

To run the tests, run the appropriate command below ([why they are different](https://github.com/pytest-dev/pytest/issues/1629#issue-161422224)):

- Python 2.7: `py.test accumulate_test.py`
- Python 3.4+: `pytest accumulate_test.py`

Alternatively, you can tell Python to run the pytest module (allowing the same command to be used regardless of Python version):
`python -m pytest accumulate_test.py`

cmccandless:
@Grociu Exercise READMEs are outside the scope of this issue.

[`configlet`](configlet) is used to generate exercise READMEs using the [track readme template] and exercise Description.md from the [problem specifications] repository.

[configlet]: https://github.com/exercism/configlet
[track readme template]: https://github.com/exercism/python/blob/master/config/exercise_readme.go.tmpl
[problem specifications]: https://github.com/exercism/problem-specifications

joaoluizn:
@cmccandless should those be closed?

![image](https://user-images.githubusercontent.com/32312434/59552967-a9378380-8f64-11e9-91f8-c3a7cf0fa78d.png)

cmccandless:
Yeah, I forgot automatic closing only works when merging with master. Thanks for pointing that out.

PR #1721, opened 2019-03-13
Remove Python 2 references in exercise README template

tqa236:
@cmccandless I deleted all Python 2.7 references in README. Does this look good for you? I can do that for other files too if this works.

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -21,11 +21,9 @@ raise Exception("Meaningful message indicating the source of the error")
 
 ## Running the tests
 
-To run the tests, run the appropriate command below ([why they are different](https://github.com/pytest-dev/pytest/issues/1629#issue-161422224)):
+To run the tests, run `pytest {{ .Spec.SnakeCaseName }}_test.py`
 
-- Python 3.4+: `pytest {{ .Spec.SnakeCaseName }}_test.py`
-
-Alternatively, you can tell Python to run the pytest module (allowing the same command to be used regardless of Python version):
+Alternatively, you can tell Python to run the pytest module:
 `python -m pytest {{ .Spec.SnakeCaseName }}_test.py`
 
 ### Common `pytest` options
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -23,7 +23,6 @@ raise Exception("Meaningful message indicating the source of the error")
 
 To run the tests, run the appropriate command below ([why they are different](https://github.com/pytest-dev/pytest/issues/1629#issue-161422224)):
 
-- Python 2.7: `py.test {{ .Spec.SnakeCaseName }}_test.py`
 - Python 3.4+: `pytest {{ .Spec.SnakeCaseName }}_test.py`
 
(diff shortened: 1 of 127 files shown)


cmccandless:
@tqa236 Which issue are you targeting? Your title suggests #1704, but your changes apply more to #1705.

tqa236:
Hello, I think it's more about #1705 than #1704. Unfortunately I haven't modified the file `exercise_readme.go.tmpl` yet but I ran a script to remove all Python 2 related path in the README of each exercise. I didn't know that the READMEs are automatically generated. What do you suggest that I should do here?

cmccandless:
I would suggest the following workflow:
1. Leave your current changes in place. Since it is all generated files, they won't cause issues if something else changes later 
1. Make the appropriate changes to `exercise_readme.go.tmpl` (we might exchange a few rounds of feedback on this one)
1. Once the template is settled, use `configlet generate` to apply the template and commit any changes to the index
1. I'll do a final once-over. If everything looks good, we'll merge!

tqa236:
The workflow is fine for me. I just deleted the line about Python 2 as a starting point for discussion. There's some comparisons of Python 2 and 3 that in my opinion unnecessary too. I want to hear your opinion about that.

config/exercise_readme.go.tmpl
```diff
@@ -23,7 +23,6 @@ raise Exception("Meaningful message indicating the source of the error")
 
 To run the tests, run the appropriate command below ([why they are different](https://github.com/pytest-dev/pytest/issues/1629#issue-161422224)):
 
-- Python 2.7: `py.test {{ .Spec.SnakeCaseName }}_test.py`
 - Python 3.4+: `pytest {{ .Spec.SnakeCaseName }}_test.py`
```
cmccandless: Track documentation will mention elsewhere that we support Python3.4+, I'm not sure it's necessary to do so here. Something like the following will suffice for this template:

> To run the tests, run `pytest {{ .Spec.SnakeCaseName }}_test.py`

config/exercise_readme.go.tmpl
```diff
@@ -23,7 +23,6 @@ raise Exception("Meaningful message indicating the source of the error")
...
 
 Alternatively, you can tell Python to run the pytest module (allowing the same command to be used regardless of Python version):
```
cmccandless: Within the same mindset as the above comment, I don't think we need the part in parenthesis here:

> Alternatively, you can tell Python to run the pytest module:

cmccandless:
Template looks good. Go ahead and generate the READMEs again.

tqa236:
How can we do that exactly? I run the command `configlet generate` but it doesn't work.

cmccandless:
You'll need to:
1. Clone exercism/problem-specifications
2. Run `configlet generate -p $WHERE_YOU_CLONED .` in the root directory for this repository

tqa236:
I generate the new READMEs with the latest `problem-specifications`, so there's some unrelated change as well.

tqa236:
Do you know why the CI doesn't mark as completed? It is already completed in the Travis CI website.

cmccandless:
@tqa236 I'm not sure why it doesn't mark as completed, but it is supposed to block incomplete from merging, and it didn't block me, so the backend is working correctly.

Thanks for working on this!

PR #1735, opened 2019-03-27
high-scores: Added new testcase for data corruption

wonhyeongseo:
The current test does not check if scores is corrupted by the use of `sort` in the personal_top_three method. From checking multiple solutions and spotting this many times, and BeneKenobi wanting a test case (uuid=bf0004960b7b4ed094743ab835f1a5f4), I wrote this.


cmccandless:
There is a pull request in exercism/problem-specifications for this case. Whether it is within scope of the tests is under discussion [there](https://github.com/exercism/problem-specifications/pull/1486).

wonhyeongseo:
Sorry for overlooking that. I'll close the pull-request. Thank you @cmccandless!

PR #1744, opened 2019-04-09
HighScore: Two new tests for methods interference

simmol:
When solving the problem event using the tests.
Is very easy to sort the scores in personal_top_tree() or personal_best() which will break latest().
The test never catch that because they are testing only one method at a time.

But knowing the difference between sorting in place and returning new sorted list is valuable.
Also is good to know early you have to be careful with objects states.

### DIFF
--- a/exercises/high-scores/high_scores_test.py
+++ b/exercises/high-scores/high_scores_test.py
@@ -47,6 +47,19 @@ class HighScoreTest(unittest.TestCase):
         expected = [40]
         self.assertEqual(HighScores(scores).personal_top_three(), expected)
 
+    def test_personal_top_three_does_not_break_latest(self):
+        scores = [20, 10, 30]
+        expected = 30
+        high_scores = HighScores(scores)
+        high_scores.personal_top_three()
+        self.assertEqual(high_scores.latest(), expected)
+    
+    def test_personal_best_does_not_break_latest(self):
+        scores = [20, 30, 10]
+        expected = 10
+        high_scores = HighScores(scores)
+        high_scores.personal_best()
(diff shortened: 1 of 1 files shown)


cmccandless:
This issue has come up before in #1735.

Currently, side effect testing is outside the scope of this exercise. There is a discussion [here](https://github.com/exercism/problem-specifications/pull/1486) to include such tests, but I would like to wait and see the consensus there before adding these tests here.

simmol:
What is the scope of the exercise ? 

Because for me at least if we introduce classes, data integrity is the lesson you need to learn as early as possible.
If the main point is list manipulation, that list is mutable is very valid point.

So for me this kind of tests are logical to exist. 
I will see the Issue you mentioned and write there also.

cmccandless:
Looking at the original PR that created the exercise (exercism/problem-specifications#1378), the intent was for an easy exercise for working with arrays/lists. Some tracks don't bother with a class implementation at all for this exercise. I'm wondering if that would be a better route for this track as well.

simmol:
That is something that bugged me also when i saw it for the first time :)
Normally python is learned trough extensive use of functions and then moving to classes.

Changing the exercise to produce just three functions that manipulate list in different manners make sense. I am not sure how to handle that change tough ? 
Isn't the Description of the exercise the same for different tracks ?

Sorry to bother you with so many questions.
I am still not sure how the whole system is organised.

If you like we can talk it trough in the maintenance channel i am willing to put some effort to improve/fix this exercise and tests. It is one of the first people see and do.

simmol:
I will do some work to change the exercise to use functions instead of class with 3 functions.
This should simply it for learners and mentor a like.

PR #1764, opened 2019-04-22
Simplify High-scores exercise. Replace Class with functions

simmol:
As discussed with @cmccandless in #1744, i submit a PR for simplifying high score exercise.  

Exercise text edited.
Tests made to expect functions. One test removed, cause was no longer valid.
Make new example.py using functions.
Add the functions as starting point in the exercise file. (This also make the tests not complain of missing imports and such.)
Added __init__.py to the package for the people that still run 2.7+
Added pytest cache to gitignore file

### DIFF
--- a/exercises/high-scores/.meta/HINTS.md
+++ b/exercises/high-scores/.meta/HINTS.md
@@ -1,4 +1,4 @@
-In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and agin in problems both simple and complex.
+In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and again in problems both simple and complex.
 
 - [**Data Structures (Python 3 Documentation Tutorial)**](https://docs.python.org/3/tutorial/datastructures.html)
 - [**Lists and Tuples in Python (Real Python)**](https://realpython.com/python-lists-tuples/)
--- a/exercises/high-scores/.meta/HINTS.md
+++ b/exercises/high-scores/.meta/HINTS.md
@@ -1,0 +1,5 @@
+In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and agin in problems both simple and complex.
+
+- [**Data Structures (Python 3 Documentation Tutorial)**](https://docs.python.org/3/tutorial/datastructures.html)
+- [**Lists and Tuples in Python (Real Python)**](https://realpython.com/python-lists-tuples/)
+- [**Python Lists (Google for Education)**](https://developers.google.com/edu/python/lists)
--- a/exercises/high-scores/.meta/HINTS.md
+++ b/exercises/high-scores/.meta/HINTS.md
@@ -1,5 +1,0 @@
-In this exercise you're going to create a **class** and use lists to organize scores.  _Don't worry, it's not as complicated as you think!_ 
-
--   [**A First Look at Classes**](https://docs.python.org/3/tutorial/classes.html#a-first-look-at-classes) from the Python 3 documentation. 
--   [**How to Define a Class in Python**](https://realpython.com/python3-object-oriented-programming/#how-to-define-a-class-in-python) from the Real Python website.  
--   [**Data Structures in Python**](https://docs.python.org/3/tutorial/datastructures.html) from the Python 3 documentation.
--- a/exercises/high-scores/README.md
+++ b/exercises/high-scores/README.md
@@ -4,6 +4,12 @@ Manage a game player's High Score list.
 
 Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write methods that return the highest score from the list, the last added score and the three highest scores.
 
+In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and agin in problems both simple and complex.
+
+- [**Data Structures (Python 3 Documentation Tutorial)**](https://docs.python.org/3/tutorial/datastructures.html)
+- [**Lists and Tuples in Python (Real Python)**](https://realpython.com/python-lists-tuples/)
+- [**Python Lists (Google for Education)**](https://developers.google.com/edu/python/lists)
+
 
 ## Exception messages
 
--- a/exercises/high-scores/README.md
+++ b/exercises/high-scores/README.md
@@ -4,11 +4,12 @@ Manage a game player's High Score list.
 
 Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write methods that return the highest score from the list, the last added score and the three highest scores.
 
-In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and agin in problems both simple and complex.
+In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and again in problems both simple and complex.
 
 - [**Data Structures (Python 3 Documentation Tutorial)**](https://docs.python.org/3/tutorial/datastructures.html)
 - [**Lists and Tuples in Python (Real Python)**](https://realpython.com/python-lists-tuples/)
 - [**Python Lists (Google for Education)**](https://developers.google.com/edu/python/lists)
+
 
 
 ## Exception messages
--- a/exercises/high-scores/README.md
+++ b/exercises/high-scores/README.md
@@ -2,7 +2,8 @@
 
 Manage a game player's High Score list.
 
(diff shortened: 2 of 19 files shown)


exercises/high-scores/README.md
```diff
@@ -2,15 +2,7 @@
...
-
-In this exercise you're going to create a **class** and use lists to organize scores.  _Don't worry, it's not as complicated as you think!_ 
-
--   [**A First Look at Classes**](https://docs.python.org/3/tutorial/classes.html#a-first-look-at-classes) from the Python 3 documentation. 
--   [**How to Define a Class in Python**](https://realpython.com/python3-object-oriented-programming/#how-to-define-a-class-in-python) from the Real Python website.  
--   [**Data Structures in Python**](https://docs.python.org/3/tutorial/datastructures.html) from the Python 3 documentation.
-
-
+Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write three functions that from the original score list will return the highest score, the last added score and the three highest scores respectivly.
```
cmccandless: This file is generated with the command `configlet generate`. The content is derived from the exercise Description.md in exercism/problem-specifications, and from the track's exercise .meta/HINTS.md. For this exercise, the only content in HINTS.md is information about classes, so I recommend simply removing that file and running `configlet generate`.

exercises/high-scores/high_scores.py
```diff
@@ -1,3 +1,10 @@
-class HighScores(object):
-    def __init__(self, scores):
-        pass
+def latest(scores):
+    pass  # Put your code here
```
cmccandless: Stub files for other exercises don't include this comment, and since this isn't the first exercise, students should be familiar with how the stub works by now.

```suggestion
    pass
```

simmol:
All Changes made.

exercises/high-scores/.meta/HINTS.md
```diff
@@ -1,5 +0,0 @@
-In this exercise you're going to create a **class** and use lists to organize scores.  _Don't worry, it's not as complicated as you think!_ 
-
--   [**A First Look at Classes**](https://docs.python.org/3/tutorial/classes.html#a-first-look-at-classes) from the Python 3 documentation. 
--   [**How to Define a Class in Python**](https://realpython.com/python3-object-oriented-programming/#how-to-define-a-class-in-python) from the Real Python website.  
--   [**Data Structures in Python**](https://docs.python.org/3/tutorial/datastructures.html) from the Python 3 documentation.
```
cmccandless: I'd like to ping @BethanyG for suggestions on new HINTS content. She put some good work into the current HINTS.md (less than a month ago), and I would feel bad about dumping that work without offering the chance to contribute an alternative.

BethanyG: I'm happy to write up some new hints...but I think the bigger issue here is that if you remove making a class in this exercise (which is a fairly easy one), then when students have to make a class (and deal with side effect questions) in the *next* exercise..well, we've just kicked the can down the road.  The same issues that apply to High Scores apply to Matrix (and Kindergarten Garden, etc.) ...and eventually, a python learner has to to deal with classes.  They're part of the language (it really is objects all the way down), and widely used in frameworks like Django.  The sooner students get (even a little) practice in a controlled fashion, the less they struggle later.  

So the way I see it is that we either have to craft an exercise or two around classes and mutability **_specifically_**, or re-work tests and mentor notes on current exercises to address it.   I don't think the answer is to re-write this exercise (and the next) to not need a class.  If we don't want a class-based exercise here because we want to focus on lists, we could move this exercises to a later stage (and address its shortcomings through a re-write of tests).  We could replace it with something like **Say**   or **Run Length Encoding** - or any other number of exercises that focus on lists and arrays.  But only 6 of 18 core exercises require writing a class currently, and I'm not sure we are doing a great job of giving students practice in that area.

BethanyG: Apologies - forgot one last point:  We'll need to re-work the mentor notes for this exercise as well.  I don't think I will be able to get to either that or the hints until this afternoon, PST.

cmccandless: > So the way I see it is that we either have to craft an exercise or two around classes and mutability specifically

I'm not opposed to that.

>  If we don't want a class-based exercise here because we want to focus on lists, we could move this exercises to a later stage (and address its shortcomings through a re-write of tests).

There are very few easy exercises as it is. I'd prefer not to overcomplicate what was meant to be a simple list exercise.

> All that being said....if we do indeed want to make this exercise function based, I can try to make some new hints this week. However, we then should move these hints to the Matrix exercise, since it will then be the exercise to introduce classes. We maybe should also provide a hint on mutability there as well.

I can get behind that.

BethanyG: Ok - I wasn't able to make suggestions in the pull (I don't have access) - but here's my proposed hint text:
***************

In this exercise, you're going to use and manipulate **_lists_**.  Python **lists** are very versatile, and you'll find yourself using them again and agin in problems both simple and complex.

- [**Data Structures (Python 3 Documentation Tutorial)**](<https://docs.python.org/3/tutorial/datastructures.html>)
- [**Lists and Tuples in Python (Real Python)**](<https://realpython.com/python-lists-tuples/>)
- [**Python Lists (Google for Education)**](<https://developers.google.com/edu/python/lists>)

*****************

I'll submit a pull for revised mentor notes and for the new hint on Matrix.

BethanyG: See https://github.com/exercism/website-copy/pull/956 for the updated mentor notes.

simmol: Thanks @BethanyG :)
I will add the new hints right now :)

And actually will push the old hints to Matrix in this PR. So we did not loose them.

For me Matrix is better Class based exercise. But we can make other also.

cmccandless: Would you mind creating a separate PR for the changes to `matrix`? We try to limit PR changes to one exercise at a time (there are few exceptions).

simmol: Well doesn't that count as exception? We have not changed the exercise in Matrix, just add the hints because previously they ware available at this exercise?

I can open other PR is not a big deal, but we have to merge them one after the other. I thought it would be better if we push this as one change. 

I will make new one tomorrow morning and will edit this one to remove matrix docs.

simmol:
Moved changes to Matrix in separate PR: #1766 
@cmccandless Hope we are good for merge now :)

exercises/high-scores/.meta/HINTS.md
cmccandless: I don't have an issue with both exercises temporarily having the same generic hints about classes, so I went ahead and merged #1766.

exercises/high-scores/.meta/HINTS.md
```diff
@@ -1,5 +1,5 @@
-In this exercise you're going to create a **class** and use lists to organize scores.  _Don't worry, it's not as complicated as you think!_ 
+In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and agin in problems both simple and complex.
```
cmccandless:
```suggestion
In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and again in problems both simple and complex.
```

exercises/matrix/README.md
```diff
@@ -40,6 +40,7 @@ And its columns:
 - 8, 3, 6
 - 7, 2, 7
 
+
```
cmccandless:
```suggestion
```

simmol:
I had applied the comments. Sorry for late response but I was on vacation.

exercises/high-scores/README.md
```diff
@@ -4,12 +4,11 @@ Manage a game player's High Score list.
 
 Your task is to build a high-score component of the classic Frogger game, one of the highest selling and addictive games of all time, and a classic of the arcade era. Your task is to write methods that return the highest score from the list, the last added score and the three highest scores.
 
-In this exercise you're going to create a **class** and use lists to organize scores.  _Don't worry, it's not as complicated as you think!_ 
-
--   [**A First Look at Classes**](https://docs.python.org/3/tutorial/classes.html#a-first-look-at-classes) from the Python 3 documentation. 
--   [**How to Define a Class in Python**](https://realpython.com/python3-object-oriented-programming/#how-to-define-a-class-in-python) from the Real Python website.  
--   [**Data Structures in Python**](https://docs.python.org/3/tutorial/datastructures.html) from the Python 3 documentation.
+In this exercise, you're going to use and manipulate lists. Python lists are very versatile, and you'll find yourself using them again and agin in problems both simple and complex.
```
cmccandless: You'll need to run `configlet generate` one more time to get Travis to pass. We should be good after that!

simmol: Will see to do it later today or first thing in the morning.

simmol:
@cmccandless I think this is all for this PR :) 
Lets merge it and continue with other things that need to be fixed/improved :)

cmccandless:

PR #1770, opened 2019-05-01
binary-search-tree: changed assertTreeEqual to assertEqual

delta48:
According to conventions in the readme file

### DIFF
--- a/exercises/binary-search-tree/binary_search_tree_test.py
+++ b/exercises/binary-search-tree/binary_search_tree_test.py
@@ -9,20 +9,20 @@ class BinarySearchTreeTest(unittest.TestCase):
 
     def test_data_is_retained(self):
         expected = TreeNode('4', None, None)
-        self.assertTreeEqual(BinarySearchTree(['4']).data(), expected)
+        self.assertEqual(BinarySearchTree(['4']).data(), expected)
 
     # Test inserting data at proper node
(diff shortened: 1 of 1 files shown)


cmccandless:
Hi, thanks for the PR!

Which convention are you referring to? I don't see any that forbids the use of utility methods like assertTreeEqual. Also, the rename you're suggesting overrides an existing definition for `unittest.TestCase.assertEqual()` that is used in `compare_tree`, so this change makes the tests non-functional (see [build output from Travis](https://travis-ci.org/exercism/python/jobs/526744681)).

delta48:
"We use assertIs(actual, True) and assertIs(actual, False) rather than assertTrue(actual) or assertFalse(actual)"
https://github.com/exercism/python/tree/track-anatomy/restructuring

cmccandless:
That specifically applies to `assertTrue` and `assertFalse`. The reasoning there is that `assertTrue` and `assertFalse` can pass on [truthy and falsey](https://docs.python.org/3/library/stdtypes.html#truth-value-testing) values, while `assertIs(True)` will only pass on `True and `assertIs(False)` will only pass on `False`.

That is not what is happening here. `assertTreeEqual` is a helper method that does not already exist in `unittest.TestCase` that specifically verifies the returned BinarySearchTree is structured as expected.

PR #1779, opened 2019-05-09
Fixed: testing wrong plants for given student

trackq:
The additional test seems to be testing the wrong plants for the given students.


trackq:
haha, my bad. 😅

PR #1781, opened 2019-05-21
add parentheses to method area_code

YuriyCherniy:
Without parentheses the test_area_code is trying compare this:` <bound method Phone.area_code of <phone_number.Phone object at 0x7f0426e266d8>> != '223'` instead of this: `'223' == '223'` so the test cannot be passed.

### DIFF
--- a/exercises/phone-number/phone_number_test.py
+++ b/exercises/phone-number/phone_number_test.py
@@ -81,7 +81,7 @@ class PhoneNumberTest(unittest.TestCase):
     # Track specific tests
     def test_area_code(self):
         number = Phone("2234567890")
-        self.assertEqual(number.area_code, "223")
+        self.assertEqual(number.area_code(), "223")
 
     def test_pretty_print(self):
         number = Phone("2234567890")


cmccandless:
This is not a bug. `Phone.area_code` should be a property, not a method.

YuriyCherniy:
> This is not a bug. `Phone.area_code` should be a property, not a method.

But why without parentheses test  returns `FAILED (failures=1)`? and with parentheses `OK`?

cmccandless:
Because you've implemented `Phone.area_code` as a method in your solution. This is incorrect. Correct example:
```Python
class Phone(object):
    def __init__(self, number):
        self.area_code = <value of area code>
```

YuriyCherniy:
> Because you've implemented `Phone.area_code` as a method in your solution. This is incorrect. Correct example:
> 
> ```python
> class Phone(object):
>     def __init__(self, number):
>         self.area_code = <value of area code>
> ```

Ok. I got it. Thank you.

Issue #1762, opened 2019-04-17
Refactor exercises to match canonical data


cmccandless:
There are a number of exercises that fall into one of the following cases:

## Function names that differ from those specified in the canonical data
`armstrong-numbers` uses `is_armstrong` instead of `is_armstrong_number`:

https://github.com/exercism/python/blob/222c02976c2c0e75d03440893f543c760ca39ba9/exercises/armstrong-numbers/example.py#L1-L2

From the [canonical-data](https://github.com/exercism/problem-specifications/blob/51418ec78bcca075b586d4c38d66d2c322b488fe/exercises/armstrong-numbers/canonical-data.json#L5-L12):
```
    {
      "description": "Zero is an Armstrong number",
      "property": "isArmstrongNumber",
      "input": {
        "number": 0
      },
      "expected": true
    },
```

*\*Note: in these cases, the camelCase names used in the canonical data should be converted to snake_case names in the tests, example solution, and solution stub.*

## Test input format differs from canonical data
`tournament` passes a multiline string instead of a list of single-line strings.

https://github.com/exercism/python/blob/222c02976c2c0e75d03440893f543c760ca39ba9/exercises/tournament/tournament_test.py#L51-L57

From the [canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/tournament/canonical-data.json#L92-L106):
```
    {
      "description": "There can be more than one winner",
      "property": "tally",
      "input": {
        "rows": [
          "Allegoric Alaskans;Blithering Badgers;loss",
          "Allegoric Alaskans;Blithering Badgers;win"
        ]
      },
      "expected": [
        "Team                           | MP |  W |  D |  L |  P",
        "Allegoric Alaskans             |  2 |  1 |  0 |  1 |  3",
        "Blithering Badgers             |  2 |  1 |  0 |  1 |  3"
		]
	},
```

## Test output format differs from canonical data
`saddle-points` expects a set of tuples instead of a list of dict objects

https://github.com/exercism/python/blob/222c02976c2c0e75d03440893f543c760ca39ba9/exercises/saddle-points/saddle_points_test.py#L16-L18

From the [canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/saddle-points/canonical-data.json#L8-L27):
```
    {
      "description": "Can identify single saddle point",
      "comments": [
        "This is the README example."
      ],
      "property": "saddlePoints",
      "input": {
        "matrix": [
          [9, 8, 7],
[...]

PR #1802, opened 2019-05-29
Markdown: sync expected test results and input data with problem-specifications.

BethanyG:
Part of https://github.com/exercism/python/issues/1762

**NOTE:**  `flake8` will report the following for the exercise stub (in this case it isn't really a *stub*, but more a file to refactor):  _**"C901 'parse' is too complex (19)**_".  Which is the point : the file needs to be refactored to reduce complexity.

- Changed  function name to **parse** in tests, `example.py`, and exercise input to conform with `canonical-data.json`.
- Changed abbreviated variable name **res** to **result** in `example.py`, for clarity.
- Changed other single-letter variable names to words in `example.py`, for clarity.

### DIFF
--- a/exercises/markdown/example.py
+++ b/exercises/markdown/example.py
@@ -1,16 +1,16 @@
 import re
 
 
-def parse_markdown(markdown):
+def parse(markdown):
     lines = markdown.split('\n')
     html = ''
     in_list = False
     in_list_append = False
     for line in lines:
-        res = parse_line(line, in_list, in_list_append)
-        html += res['line']
-        in_list = res['in_list']
-        in_list_append = res['in_list_append']
+        result = parse_line(line, in_list, in_list_append)
+        html += result['line']
+        in_list = result['in_list']
+        in_list_append = result['in_list_append']
     if in_list:
         html += '</ul>'
     return html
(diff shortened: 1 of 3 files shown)


BethanyG:
arg.  apologies.  Forgot to update this branch.  Will update for those that follow.

cmccandless:
> NOTE: flake8 will report the following for the exercise stub (in this case it isn't really a stub, but more a file to refactor): "C901 'parse' is too complex (19)".

`flake8` passed in the Travis builds. You might have a non-default configuration for your `flake8` installation.

exercises/markdown/example.py
```diff
@@ -22,9 +22,9 @@ def wrap(line, tag):
 
 def check_headers(line):
     pattern = '# (.*)'
-    for i in range(6):
+    for index in range(6):
```
cmccandless: Out of curiosity: does `flake8` complain about this one, or was that your personal style preference?

cmccandless:

exercises/markdown/example.py
BethanyG: Apologies - for whatever reason, this comment did not get emailed to me...so I am only seeing it now.  That change is a personal preference.  I can always refrain if it is bugging you/not good practice in this situation.

cmccandless: Doesn't bother me; I just see `for i in range(n)` so frequently that I generally give that one a pass on variable naming rule.

PR #1806, opened 2019-05-31
two_fer: sync with canonical data

Grociu:
part of #1762 

- changed the argument of test function to None from empty string.
- rewritten the example to fit the new test

### DIFF
--- a/exercises/two-fer/example.py
+++ b/exercises/two-fer/example.py
@@ -1,5 +1,5 @@
-def two_fer(name=""):
-    if not name.strip():
+def two_fer(name=None):
+    if not name:
         return "One for you, one for me."
     else:
         return "One for {}, one for me.".format(name)
--- a/exercises/two-fer/two_fer_test.py
+++ b/exercises/two-fer/two_fer_test.py
@@ -7,7 +7,7 @@ from two_fer import two_fer
 
 class TwoFerTest(unittest.TestCase):
     def test_no_name_given(self):
-        self.assertEqual(two_fer(), 'One for you, one for me.')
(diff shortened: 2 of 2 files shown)


exercises/two-fer/example.py
```diff
@@ -1,5 +1,5 @@
-def two_fer(name=""):
-    if not name.strip():
```
Grociu: This no longer worked with None as the argument for test_no_name_given

exercises/two-fer/two_fer_test.py
```diff
@@ -7,7 +7,7 @@
 
 class TwoFerTest(unittest.TestCase):
     def test_no_name_given(self):
-        self.assertEqual(two_fer(), 'One for you, one for me.')
+        self.assertEqual(two_fer(None), 'One for you, one for me.')
```
Grociu: As per canonical data switched null to None as this is the equivalent.

cmccandless:

yawpitch:
I'm going to say this change should **not** have been made.

It makes no sense to force students two write an entirely un-Pythonic solution due to what is effectively an oversight in the canonical data. It is insensible to make a `two_fer(None)` call rather than a `two_fer()` call when attempting to teach default arguments, as the default argument can _never_ be involved with the tests as rewritten.

SleeplessByte:
FWIW: `None` might be equivalent to `null` when that `null` is _JavaScript_, but I would say that normally we would have used `undefined` in the JS equivalent of this canonical data. `null` from JSON doesn't blindly translate.

PR #1807, opened 2019-05-31
transpose: sync with canonical data

Grociu:
part of #1762 

- changed the argument name to lines
- rewritten two test functions to construct their own multiline strings as per:

>   "comments": [
>     "JSON doesn't allow for multi-line strings, so all multi-line input is ",
>     "presented here as arrays of strings. It's up to the test generator to join the ",
>     "lines together with line breaks."
> ],

- reordered the tests to fit the canonical data scheme
- renamed functions in the example to fit the general theme.


cmccandless:

PR #1808, opened 2019-05-31
sum_of_multiples: sync with canonical data

Grociu:
part of #1762 

- changed the stub, tests, and example to fit with the [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/sum-of-multiples/canonical-data.json)
- order of arguments changed
- test function names changed to canon
- the function name is supposed to be `sum()` which is a name of a built-in function. This was addressed in the `example.py` by `old_sum = sum`
- one of the test function names would be too long if used exactly as in canon. Used a truncated name for now. Discuss inline.

### DIFF
--- a/exercises/sum-of-multiples/example.py
+++ b/exercises/sum-of-multiples/example.py
@@ -1,5 +1,8 @@
-def sum_of_multiples(limit, multiples):
-    return sum(value for value in range(limit)
-               if any(value % multiple == 0
-                      for multiple in multiples
-                      if multiple > 0))
+old_sum = sum
+
+
+def sum(factors, limit):
+    return old_sum(value for value in range(limit)
+                   if any(value % multiple == 0
+                          for multiple in factors
+                          if multiple > 0))
--- a/exercises/sum-of-multiples/sum_of_multiples.py
+++ b/exercises/sum-of-multiples/sum_of_multiples.py
@@ -1,2 +1,2 @@
-def sum_of_multiples(limit, multiples):
+def sum(factors, limit):
     pass
--- a/exercises/sum-of-multiples/sum_of_multiples_test.py
+++ b/exercises/sum-of-multiples/sum_of_multiples_test.py
@@ -1,6 +1,6 @@
 """
 You can make the following assumptions about the inputs to the
-'sum_of_multiples' function:
+'sum' function:
     * All input numbers are non-negative 'int's, i.e. natural numbers
       including zero.
     * A list of factors must be given, and its elements are unique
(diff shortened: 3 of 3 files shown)


exercises/sum-of-multiples/example.py
```diff
@@ -1,5 +1,8 @@
-def sum_of_multiples(limit, multiples):
-    return sum(value for value in range(limit)
-               if any(value % multiple == 0
-                      for multiple in multiples
-                      if multiple > 0))
+old_sum = sum
```
Grociu: This is needed to preserve the sum function in some way, as it is used in the example.

exercises/sum-of-multiples/sum_of_multiples.py
```diff
@@ -1,2 +1,2 @@
-def sum_of_multiples(limit, multiples):
+def sum(factors, limit):
```
Grociu: Note the order of arguments had to be switched.
In addition, we might want to include `old_sum = sum` in the stub, or something like that to help the student with the general confusion of reusing `sum`

exercises/sum-of-multiples/sum_of_multiples_test.py
```diff
@@ -9,60 +9,59 @@
...
+        self.assertEqual(sum([5, 25], 51), 275)
 
-    def test_multiples_with_large_factors(self):
-        self.assertEqual(sum_of_multiples(10000, [43, 47]), 2203160)
+    def test_much_larger_factors(self):
+        self.assertEqual(sum([43, 47], 10000), 2203160)
 
-    def test_multiples_of_one_will_be_all(self):
-        self.assertEqual(sum_of_multiples(100, [1]), 4950)
+    def test_all_numbers_are_multiples_of_1(self):
```
Grociu: Function name uses a numeral. Is this OK? It's like this in canon.

exercises/sum-of-multiples/sum_of_multiples_test.py
```diff
@@ -9,60 +9,59 @@
...
 
-    def test_multiples_with_a_zero_factor(self):
-        self.assertEqual(sum_of_multiples(4, [0, 3]), 3)
+    def test_the_factor_0_does_not_affect_the_sum_of_other_factors(self):
+        self.assertEqual(sum([3, 0], 4), 3)
 
-    def test_multiples_of_several_factors(self):
-        self.assertEqual(sum_of_multiples(10000,
-                         [2, 3, 5, 7, 11]), 39614537)
+    def test_solutions_must_extend_to_cardinality_greater_than_3(self):
```
Grociu: This function in canon is defined as `"solutions using include-exclude must extend to cardinality greater than 3",` which would make the name too long. I truncated it at this time but it's technically not consistent with canon.

cmccandless:
After further thought, it seems to me that teaching students to shadow builtins is not necessarily a good idea. I am not sure how to handle this one at this time.

exercises/sum-of-multiples/sum_of_multiples.py
yawpitch: We should _not_ shadow builtin Python functions like **sum** in any change we make. If the name in the canonical data conflicts with a Python keyword or builtin the Python name must be retained.

yawpitch:
We cannot and should not shadow builtins. What if the canonical data used a reserved keyword like **print**? We would be _unable_ to sync that from the canonical data, so we'd be forced to be out of sync with the canonical naming. The same _should_ be followed with any _builtin_ name.

Grociu:
Should I rewrite this to use  `sum_of_multiples` as the main function name then?

cmccandless:
Having thought on this further, I am leaning towards dropping this PR and leaving this exercise as-is for the time being:
- Shadowing builtins is not a good practice (in general)
- The remaining changes in this PR are trivial, and can be addressing if/when a test generator is deployed

yawpitch:
I'm going to second dropping this PR. I'm hoping to have some more time to spend on the generator soon, but I believe what I've already got in place could handle the relatively minor deviations from the canonical data.

PR #1809, opened 2019-06-01
sublists: sync with canonical data

Grociu:
part of  #1762 

- renamed the function and arguments in stub, tests and example.
- deleted references to `SUBLIST` `SUPERLIST` `EQUAL` `UNEQUAL` as [canonical_data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/sublist/canonical-data.json) seems to want a string result.
- deleted a test that referenced those variables only.
- made the additional tests more consistent with the indentation scheme of the test file.


exercises/sublist/example.py
```diff
@@ -1,29 +1,23 @@
-SUBLIST = 0
```
Grociu: This code seems unnecessary to me:

- in canon I see ` "expected": "equal"` which signifies a string result to me.
- therefore I deleted all references to `SUBLIST`, `SUPERLIST`, `EQUAL`, `UNEQUAL` and changed the expected results to strings.

exercises/sublist/sublist_test.py
```diff
@@ -1,134 +1,137 @@
...
-            UNEQUAL
+            sublist(['a c'], ['a', 'c']),
+            "unequal"
         )
 
     # additional track specific test
     def test_large_lists(self):
         l1 = list(range(1000)) * 1000 + list(range(1000, 1100))
         l2 = list(range(900, 1050))
-        self.assertEqual(check_lists(l1, l2), SUPERLIST)
```
Grociu: This was inconsistent with the rest as far as indentation goes, so I corrected it.

exercises/sublist/sublist_test.py
```diff
@@ -1,134 +1,137 @@
 import unittest
 
-from sublist import check_lists, SUBLIST, SUPERLIST, EQUAL, UNEQUAL
+from sublist import sublist
 
 
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.1.0
 
 class SublistTest(unittest.TestCase):
-    def test_unique_return_values(self):
```
Grociu: This test referenced only the deleted variables, so it was deemed unnecessary.

Grociu:
Thank you @yawpitch for this write-up and the time you took in making this post. I've learned something new, which is always appreciated.

For the issue at hand, I can rewrite this to include these global constants, but would that be considered consistent with canonical data (by the maintainer), as per the definition of the task? We've made several minor compromises with the other examples already but this seems more major. The 'yacht' exercise uses a similar scheme of NAME variables so that would clarify further syncs as well.

In addition, while these exercises (sublists, yacht, maybe others?) use the aforementioned scheme, it's not explained in any way. It can be somewhat inferred from the test function, but it causes confusion - I did yacht quite early into learning Python, and I was confused. Is there any interest of putting a small explanation/resource of that into the `README.md` maybe?

cmccandless:
> For the issue at hand, I can rewrite this to include these global constants, but would that be considered consistent with canonical data (by the maintainer), as per the definition of the task?

Yes. Synchronization with the canonical data is not helpful if the tests become un-[Pythonic](https://blog.startifact.com/posts/older/what-is-pythonic.html).

> In addition, while these exercises (sublists, yacht, maybe others?) use the aforementioned scheme, it's not explained in any way. It can be somewhat inferred from the test function, but it causes confusion - I did yacht quite early into learning Python, and I was confused. Is there any interest of putting a small explanation/resource of that into the README.md 

Whichever exercise in the track progression first uses returned constants should contain something in its HINTS.md mentioning this. Perhaps exercise stubs should also contain stub entries for the enumeration items?

```Python
SUBLIST = None
SUPERLIST = None
EQUAL = None
```

This exercise currently contains a check to make sure these values are unique; this should probably be done for all exercises that use returned constants.

Additionally, it may be worthwhile to mention this in our [List of Conventions](https://github.com/exercism/python/blob/master/README.md#conventions)

yawpitch:
Struggling to maintain strong parity with the canonical data is not a worthwhile endeavor; the canonical data writers cannot know what abstractions exist in all the track languages, and it’s going to have to be up to the track’s task implementer to determine if literal datatype in the canonical version is a good Python idiom. From this point of view deviating from the canonical data _to teach the language better_ is not a compromise... slavishly following the canonical data, though, _would_ be a compromise.

Enumerated constants are a basic programming pattern and students need to get used to seeing them, so we certainly should use them in problems in which they’re appropriate.

Any time the tests **must** have access to a constant value that the thing under test either requires as an argument or returns as a value then these global constants are a sensible and idiomatic design choice.

Now, in Python 3 there’s the **enum** module, and ideally we could use that instead of raw global names, but there’s an unknown amount of time until we’re able or willing to abandon Python 2 support.

And yes, updating the README is a reasonable first step if it helps reduce confusion. Personally I think their existence in the “slug” file they get when they download the exercise is sufficient, but for beginners it may not be.

cmccandless:
> there’s an unknown amount of time until we’re able or willing to abandon Python 2 support.

Just to make a note here on that: I believe all issues related to [dropping Python 2](https://github.com/exercism/python/milestone/1) have been closed, so those changes will probably be merged sometime in the second half of this year.

yawpitch:
Good to hear. If we’re assuming 3.6+ we can do a lot to improve things for yourselves and for students. Actual enum types, type hinting, and f-strings... though we should be careful too heavily promoting the latter, as they’re still impervious to translation into other languages.

cmccandless:
> f-strings... though we should be careful too heavily promoting the latter, as they’re still impervious to translation into other languages.

Not really; string interpolation is present in a lot of common languages these days: C#, Javascript, and Ruby all have it, although syntax varies a little.

yawpitch:
Sorry, should have been more clear: the Python f-string is not well-suited for internationalization / localization... an f-string that starts in English will remain in English, while template strings that use **str.format** are amenable to translation with standard tools like `gettext`. I was not referring to string interpolation or translation into other programming languages.

Grociu:
Second revision:

- Returned the enumerated constants to the exercise test suite.
- Entered the enumerated constants into the exercise stub, with `None` as the base value.
- Returned the uniqueness test into the test suite, and moved to the end with other tests not included in [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/sublist/canonical-data.json)
- Minor renaming of variables in one of the tests
- Added a HINTS.md file with a basic explanation of the concept based on feedback from @yawpitch

exercises/sublist/.meta/HINTS.md
```diff
@@ -0,0 +1,10 @@
+This exercise stub and the test suite contain several enumerated constants. 
+
+Since Python 2 does not have the enum module, the idiomatic way to write 
+enumerated constants has traditionally been a NAME assigned to an arbitrary, 
+but unique value. An integer is traditionally used because it’s memory 
+efficient.
+It is a common practice to export both constants and functions that work with 
+those constants (ex. the constants in the os, subprocess and re modules). 
+
+You can learn more here: https://en.wikipedia.org/wiki/Enumerated_type
```
Grociu: Feel free to suggest edits to this file.

exercises/sublist/sublist.py
```diff
@@ -1,2 +1,10 @@
-def check_lists(first_list, second_list):
+# Possible sublist categories.
+# Change the values as you see fit.
```
Grociu: These mirror the stub in the [yacht](https://github.com/exercism/python/blob/master/exercises/yacht/yacht.py) exercise.

exercises/sublist/.meta/HINTS.md
yawpitch: I'd have to look at bin/check-readmes.sh to see why that error would occur, but that said I'd propose deleting this file and instead including the hints above as comments at the top of the `sublist.py` file.

yawpitch:
If I'm not mistaken, at this point -- after my comment about HINTS.md is taken into account -- the only _substantive_ change is renaming **check_lists** to **sublist**, correct? The rest is just formatting and a few temporary variable names?

For @cmccandless: is syncing the name of the function under test to the canonical data that important? I ask only because this PR will "break" every currently published Community Solution for the exercise without adding any actual new tests.

exercises/sublist/.meta/HINTS.md
cmccandless: The error occurred because the README.md for the exercise was not re-generated to include this new content.

Grociu:
Third pass:

- Deleted the hint file, and moved the contents to the exercise stub.

yawpitch:
Ok, notwithstanding my concern about invalidating all the existing Community Solutions for the dubious value of having the function name be the same as the one on the canonical data, this is safe to merge without further changes.

I personally don’t think the name being in sync is worth it, nor something most tracks would fight for. The odds of all properties in the canonical data files being named in a manner consistent with good practice in any language are _low_.

But I’ll leave that determination to @cmccandless.

PR #1818, opened 2019-06-02
README.md: Remove references to Python 2.7 from README.md (#1704)

joaoluizn:
Following the patterns from previous commits over branch _**drop-python2**_: 
- Removed the references for Python 2.7 from **README.md**
- Updated the Python _3_ version to _3.6_ following the **ABOUT.md** guidelines on #1760.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,14 +30,14 @@ A list of missing exercise can be found here: https://github.com/exercism/python
 
 ### Testing
 
-All exercises must be compatible with Python versions 2.7 and 3.4 upwards.
+All exercises must be compatible with Python version 3.6 upwards.
 
-To test a single exercise (e.g., with Python 2.7):
+To test a single exercise:
 ```
-python2.7 test/check-exercises.py [exercise-name]
+python3 test/check-exercises.py [exercise-name]
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,14 +30,14 @@ A list of missing exercise can be found here: https://github.com/exercism/python
 
 ### Testing
 
-All exercises must be compatible with Python versions 2.7 and 3.4 upwards.
+All exercises must be compatible with Python version 3.6 upwards.
```
cmccandless: For the time being, we will continue to support 3.4.

```suggestion
All exercises must be compatible with Python version 3.4 upwards.
```

joaoluizn:
https://github.com/exercism/python/blob/drop-python2/docs/ABOUT.md

image

I used 3.6 because the about section, but okay I changed to 3.4, it's okay

cmccandless:
@joaoluizn For the time being, exercises will be compatible with Python 3.4+. We do recommend that students just starting the track install Python 3.6 or later, but for students who already have 3.4 or 3.5 installed and don't wish to meddle with multiple installations, we do support these versions as well.

PR #1821, opened 2019-06-03
two-fer: revert change to explicit None argument

cmccandless:
The changes being reverted by this PR were made to blindly conform to a 1-to-1 interpretation of the canonical data. I have been convinced that these changes are not idiomatic, and should not have been included.

Ref: [community response](https://github.com/exercism/python/pull/1806#issuecomment-497971969)

Maintainer's Note: when test generators are deployed, custom handling may be required for this exercise.

### DIFF
--- a/exercises/two-fer/example.py
+++ b/exercises/two-fer/example.py
@@ -1,5 +1,2 @@
 def two_fer(name=None):
-    if not name:
-        return "One for you, one for me."
-    else:
-        return "One for {}, one for me.".format(name)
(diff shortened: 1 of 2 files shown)


cmccandless:
Pinging @Grociu @yawpitch @Alexhans

Alexhans:
Thanks for recognizing the issue and reverting the change @cmccandless.

It seems the example has an over-indentation problem but that should be an easy fix.

Issue #1826, opened 2019-06-18
anagram: tests should not require specific order for returned list


roadfoodr:
The `anagram` problem description does not specify a requirement that the elements of the returned sublist be in any particular order.  However, tests in the suite check against a particular ordered list (instead of checking that all the elements of the list are present).

Motivation: in my solution, I wanted to return a list of unique anagrams (although the problem description does not require this.)  I included a call to `set()` in the process, which changed the order of elements.

If there is interest in relaxing this requirement in the test suite, I will be happy to attempt a pull request.

cmccandless:
@ydrecords Feel free to submit a PR. For testing purposes, please check your changes to the test suite against both your solution and the example solution as-is to verify that no bugs are introduced.

roadfoodr:
Thank you.  I believe the most straightforward implementation would be to use `assertCountEqual()`.  However, this was first introduced in Python 3.2.  Does the test suite need to be compatible with Python 2.7?

cmccandless:
Closed in #1828

PR #1828, opened 2019-06-19
Anagram: relax test requirement for ordered elements

roadfoodr:
See issue #1826.  
- [x] Tested against reference solution and own solution. 
- [x] Tested full track-level test suite.

Note: this implementation uses `assertCountEqual()`, first available in Python 3.2.

"Update tests to v1.4.1" references: https://github.com/exercism/problem-specifications/pull/1534

### DIFF
--- a/exercises/anagram/anagram_test.py
+++ b/exercises/anagram/anagram_test.py
@@ -6,6 +6,7 @@ from anagram import find_anagrams
 if not hasattr(unittest.TestCase, 'assertCountEqual'):
     unittest.TestCase.assertCountEqual = unittest.TestCase.assertItemsEqual
 
+
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.4.1
 
 class AnagramTest(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


yawpitch:
Thanks for working towards improving the tests, I agree there’s no reason to worry about the sorted order of the results, however we can’t drop Python 2 comparability completely just yet.

Could you please insert this at the top of the file:

```python
# Python 2/3 compatibility
if not hasattr(unittest.TestCase, 'assertCountEqual'):
    unittest.TestCase.assertCountEqual = unittest.TestCase.assertItemsEqual
```

We’ve used it in other exercises to the same end.

roadfoodr:
Done, thanks.  (I was wondering about compatibility since `README.md` describes the use of `assertIs()`, also new in Python 3.)

yawpitch:
The `TestCase.assertIs` method was added to 2.7 some time ago? So no compatibility issue there.

@cmccandless I’d say this one is ready to (squash and) merge.

cmccandless:

PR #1829, opened 2019-06-19
yacht: sync with canonical data

Grociu:
Part of #1762 
Based upon #1809 

- Updated test suite to v1.2.0 of [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/yacht/canonical-data.json).
- Fixed an error in one of the test cases, and renamed functions to match canon.
- Added the same comment concerning enumerated constants as in the sublists exercise.
- Minor variable and message changes to `example.py`

### DIFF
--- a/exercises/yacht/yacht_test.py
+++ b/exercises/yacht/yacht_test.py
@@ -4,7 +4,7 @@ import yacht
 from yacht import score
 
 
-# Tests adapted from `problem-specifications//canonical-data.json` @ v1.1.0
+# Tests adapted from `problem-specifications//canonical-data.json` @ v1.2.0
 
 class YachtTest(unittest.TestCase):
     def test_yacht(self):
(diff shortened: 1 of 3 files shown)


Grociu:
The test file imports the yacht module and refers to the enumerated constants as `yacht.BIG_STRAIGHT` for example. In the sublists exercise #1809 we imported all the constants into the test suite, and referred to them directly. This scheme can be implemented in this exercise as well, as it would better match canonical data, but it would look somewhat bulky:
`from yacht import score, ONES, TWOS, THREES, ..., YACHT, CHANCE`
`from yacht import *` could be theoretically be used. Would that be needed and proper in this context?

cmccandless:
`flake8` F403 prevents `from yacht import *`. I believe the reasoning behind this is that it is difficult to detect undefined symbols if `flake8` has no way of knowing what is and isn't imported.

You can also wrap multi-imports in parenthesis, like so:
```
from yacht import (
	score,
	ONES,
	TWOS,
	...
	CHANCE,
)
```

This satisfies `flake8` just fine.

cmccandless:
That said, I think `yacht.ONES` is fine and can be left alone.

exercises/yacht/yacht_test.py
```diff
@@ -59,7 +59,7 @@ def test_yacht_can_be_scored_as_four_of_a_kind(self):
         self.assertEqual(score([3, 3, 3, 3, 3], yacht.FOUR_OF_A_KIND), 12)
 
     def test_full_house_is_not_four_of_a_kind(self):
-        self.assertEqual(score([3, 5, 4, 1, 2], yacht.FOUR_OF_A_KIND), 0)
+        self.assertEqual(score([3, 3, 3, 5, 5], yacht.FOUR_OF_A_KIND), 0)
```
cmccandless: Good catch.

PR #1834, opened 2019-07-03
complex-numbers: add __eq__ skeleton

pranasziaukas:
All of the needed method skeletons (`__add__`, `__mul__`, etc.) are provided in the initial solution file except for `__eq__`.

Proposing to add it as well for the sake of consistency.

### DIFF
--- a/exercises/complex-numbers/complex_numbers.py
+++ b/exercises/complex-numbers/complex_numbers.py
@@ -2,6 +2,9 @@ class ComplexNumber(object):
     def __init__(self, real, imaginary):
         pass
 
+    def __eq__(self, other):
+        pass
+
     def __add__(self, other):
         pass
 
--- a/exercises/complex-numbers/complex_numbers_test.py
+++ b/exercises/complex-numbers/complex_numbers_test.py
@@ -35,6 +35,13 @@ class ComplexNumbersTest(unittest.TestCase):
         input_number = ComplexNumber(1, 2)
         self.assertEqual(input_number.imaginary, 2)
 
+    def test_complex_number_equality_and_inequality(self):
+        input_number = ComplexNumber(1, 2)
+        other_number = ComplexNumber(2, 1)
+        expected = ComplexNumber(1, 2)
+        self.assertEqual(input_number, expected)
+        self.assertNotEqual(other_number, expected)
+
     def test_imaginary_unit(self):
         first_input = ComplexNumber(0, 1)
         second_input = ComplexNumber(0, 1)
--- a/exercises/complex-numbers/complex_numbers_test.py
+++ b/exercises/complex-numbers/complex_numbers_test.py
@@ -35,12 +35,11 @@ class ComplexNumbersTest(unittest.TestCase):
         input_number = ComplexNumber(1, 2)
         self.assertEqual(input_number.imaginary, 2)
 
-    def test_complex_number_equality_and_inequality(self):
-        input_number = ComplexNumber(1, 2)
-        other_number = ComplexNumber(2, 1)
-        expected = ComplexNumber(1, 2)
-        self.assertEqual(input_number, expected)
-        self.assertNotEqual(other_number, expected)
+    def test_equality_of_complex_numbers(self):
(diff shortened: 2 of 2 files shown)


pranasziaukas:
> Looks good to me. It’s kind of strange we don’t have a direct test for equality, but instead indirect tests on component equality, but having this in the slug is fine.

I guess a direct test on equality by itself would look like
```
self.assertEqual(ComplexNumber(1, 2), ComplexNumber(1, 2))
```
which is a bit trivial.

Otherwise, equality is tested for all the algebraic operations. I think it's fair enough.

yawpitch:
It would be trivial, but from my quick glance at the tests earlier it appears trivial to also pass all the existing tests without ever defining __eq__. We’re assuming it’s presence through the other tests, but never assuring it’s presence.

pranasziaukas:
> It would be trivial, but from my quick glance at the tests earlier it appears trivial to also pass all the existing tests without ever defining __eq__. We’re assuming it’s presence through the other tests, but never assuring it’s presence.

How can you pass the current tests trivially without defining `__eq__`?

pranasziaukas:
I observed that you can have
```
def __eq__(self, other):
    return True
```
and pass the current tests with other methods being rubbish.

In that case a direct test for complex number (in)equality is useful. What do you think?

yawpitch:
> > It would be trivial, but from my quick glance at the tests earlier it appears trivial to also pass all the existing tests without ever defining **eq**. We’re assuming it’s presence through the other tests, but never assuring it’s presence.
> 
> How can you pass the current tests trivially without defining `__eq__`?

That was my faulty quick read on mobile, sorry; you do indeed have to define `__eq__` (in Python 3, you could get away with just `__cmp__` in Python 2), we just never actually verify that it's a meaningful and correct comparison.

pranasziaukas:
Separate tests are fine by me.

cmccandless:
Sorry, busy week getting ready for a move, and I didn't have the time to chime in on this before it was merged.

I have no problem with what was merged, but I wonder if the equality tests should also include the following:
1. test_inequality_different_real_part
2. test_inequality_different_imaginary_part

@yawpitch @pranasziaukas thoughts?

pranasziaukas:
I believe that the current setup strongly suggests to store parts of a complex number in `self.real` and `self.imaginary` (or something comparable) which results in a simple numerical comparison. In that case, if e.g. `number.real == 1` is true then it's not going to hold for any different value.

So I don't see how you could cheat the current test suite without creating some monstrosity with a custom equality.

TL;DR Looks good to me as it is.

yawpitch:
I'd tend to say the single inequality test is fine, since both values are being changed in the test. Hence if they've botched `__eq__` for either the real or imaginary part the test will fail, which should be sufficient.

cmccandless:
I was thinking more if there was an implementation like this:
```Python
class ComplexNumber(object):
    def __eq__(self, other):
        return self.real == other.real
```

`ComplexNumber(2, 1) == ComplexNumber(2, 3)` should be `False`, but would evaluate to `True`. I might be overthinking it.

pranasziaukas:
Okay, @cmccandless I misunderstood your idea yesterday.

In the case you gave it seems like a better idea to test
```
    def test_inequality_of_complex_numbers(self):
        number = ComplexNumber(1, 2)
        number_different_real_part = ComplexNumber(2, 2)
        number_different_imaginary_part = ComplexNumber(1, 1)
        self.assertNotEqual(number, number_different_real_part)
        self.assertNotEqual(number, number_different_imaginary_part)
```
And maybe that's actually better. 
But honestly I don't know enough about the culture at Exercism to know how deep you wish to go down the rabbit hole :)

yawpitch:
I’d personally say that’s exactly as deep down the rabbit hole as we should go for now.

yawpitch:
@pranasziaukas do you want to add the improved inequality tests that handle both real and imaginary parts? I'd be happy to merge this either way.

cmccandless:
That will require a new PR, as this one has already been merged.

yawpitch:
Woops, should have been more clear, edited my previous.

PR #1835, opened 2019-07-04
say: test `int`, not `float`

pranasziaukas:
2 (out of 15) test cases pass `float` instead of `int`, namely `1e6` and `1e9`. I feel that this is not consistent (besides, decimal parts have nothing to do with this exercise) and suggest using `10**6` and `10**9` respectively.

### DIFF
--- a/exercises/complex-numbers/complex_numbers.py
+++ b/exercises/complex-numbers/complex_numbers.py
@@ -2,6 +2,9 @@ class ComplexNumber(object):
     def __init__(self, real, imaginary):
         pass
 
+    def __eq__(self, other):
+        pass
+
     def __add__(self, other):
         pass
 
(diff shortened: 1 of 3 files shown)


pranasziaukas:
As an additional example, passing `float` voids this (in my opinion, elegant) solution:
```
def say_up_to_10(number):
    return ['zero', 'one', 'two', 'three', 'four',
            'five', 'six', 'seven', 'eight', 'nine'][number]
```
`TypeError: list indices must be integers or slices, not float`

pranasziaukas:
I think I made a mess by having some commits from an older PR...

Should I close this and reopen from scratch?

yawpitch:
Actually, while we're at it, use the literal form `1_000_000_000` instead of the powers.

pranasziaukas:
I'll make changes when I'm back home.

However, I believe that the literal form is for Python 3.6+, is that okay?

yawpitch:
Yes, it was. You could also omit the underscores (ie 1000000000); not as pretty, but consistent with the canonical data.

PR #1836, opened 2019-07-04
say: pass 'int' instead of 'float' in test suite

pranasziaukas:
This is absolutely the same as #1835, just messed the commit history the first time.


pranasziaukas:
As expected, Travis is angry about the literal form in Python <3.5.
Switched to the simplest form `1000000` although I'm not fully sure if that's better (simpler/more readable/more understandable) compared to `10**6`.

yawpitch:
We don’t really need to worry about readability inside tests; besides, once we’ve got a test generator working it will just transcribe the literal found in the canonical data, it won’t try to reduce it to a more readable form.

yawpitch:
Thanks for the fix!

PR #1838, opened 2019-07-10
dot-dsl: remove mutable default arguments

pranasziaukas:
Unintentional mutable default arguments lead to surprising results and are confusing to newcomers to say the least ( https://docs.python-guide.org/writing/gotchas ).

A mutable default argument
```python
def __init__(self, attrs={}):
    self.attrs = attrs
```
can be replaced with 
```python
def __init__(self, attrs=None):
    self.attrs = {} if attrs is None else attrs
```
or possibly even more compact
```python
def __init__(self, attrs=None):
    self.attrs = attrs or {}
```

However, this particular exercise forces all of the arguments to be passed through the test suite (specifically, `test_malformed_EDGE`). In that case, I think it makes sense not to provide any defaults for the constructors of `Node` and `Edge` at all, and an immutable default (or, again, no default at all) for the constructor of `Graph`.

### DIFF
--- a/exercises/dot-dsl/dot_dsl.py
+++ b/exercises/dot-dsl/dot_dsl.py
@@ -2,7 +2,7 @@ NODE, EDGE, ATTR = range(3)
 
 
 class Node(object):
-    def __init__(self, name, attrs={}):
+    def __init__(self, name, attrs):
         self.name = name
         self.attrs = attrs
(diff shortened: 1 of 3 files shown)


yawpitch:
@cmccandless any thoughts on this one before I merge?

PR #1843, opened 2019-07-10
kindergarten-garden: improve example.py

yawpitch:
The example.py has a mutable default, improperly named constants, and incorrectly invokes name mangling on a "non-public" class attribute. It's also inefficient.

### DIFF
--- a/exercises/kindergarten-garden/example.py
+++ b/exercises/kindergarten-garden/example.py
@@ -24,8 +24,12 @@ class Garden(object):
             start = idx * 2
             stop = start + 2
             self.cups.setdefault(student, [])
-            self.cups[student].extend(self.PLANTS[p] for p in front[start:stop])
-            self.cups[student].extend(self.PLANTS[p] for p in back[start:stop])
+            self.cups[student].extend(
+                self.PLANTS[p] for p in front[start:stop]
+            )
+            self.cups[student].extend(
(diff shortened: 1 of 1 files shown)


cmccandless:
This can be merged once the `flake8` violation has been fixed:
```bash
$ flake8
./exercises/kindergarten-garden/example.py:27:80: E501 line too long (80 > 79 characters)
The command "flake8" failed and exited with 1 during .
```

exercises/kindergarten-garden/example.py
```diff
@@ -1,18 +1,31 @@
 class Garden(object):
 
-    __plant_names = {"C": "Clover", "G": "Grass",
-                     "R": "Radishes", "V": "Violets"}
+    STUDENTS = [
```
cmccandless: Since this is already a separate file/module, is it redundant to nest `STUDENTS` and `PLANTS` under `Garden`?

yawpitch: Judgement call ... STUDENTS could be seen as independent of the **Garden** class, but **PLANTS** is pretty bound to the class. I've seen a number of students prefer to have the lookup table as a class constant, and by making them both class bound the file's a bit cleaner to read.

yawpitch:
> This can be merged once the `flake8` violation has been fixed:

Sigh ... there's the issue with flake8 thinking 79 is 80. Fixed.

PR #1857, opened 2019-07-23
Create test suite generator (requires per-exercise templates)

cmccandless:
Using the WIP extension instead of draft PR so that CI scripts will run on these changes.


cmccandless:
[Example template for `two-fer`](https://github.com/cmccandless/python/blob/55258ceea0ccb502a1b0a411bc7a81d121bc0f0a/exercises/two-fer/.meta/template.j2)

bin/generate_tests.py
```diff
@@ -0,0 +1,103 @@
...
+
+DEFAULT_SPEC_LOCATION = os.path.join('..', 'problem-specifications')
+RGX_WORDS = re.compile(r'[-_\s]|(?=[A-Z])')
+
+logging.basicConfig()
+logger = logging.getLogger('generator')
+logger.setLevel(logging.WARN)
+
+
+def snake_case(string):
```
yawpitch: I'm pretty sure the `to_snake` from #1547 will handle more cases here.

bin/generate_tests.py
```diff
@@ -0,0 +1,103 @@
...
+    return '_'.join(RGX_WORDS.split(string)).lower()
+
+
+def title_case(string):
```
yawpitch: It's a bit more accurate to call this `camel_case` rather than `title_case` ... title case usually means "Every Significant Word Gets an Uppercase First Letter but with Spaces Maintained".

bin/generate_tests.py
```diff
@@ -0,0 +1,103 @@
...
+        spec = load_canonical(slug, spec_path)
+        template_path = os.path.join(slug, '.meta', 'template.j2')
+        try:
+            template = env.get_template(template_path)
+            tests_path = os.path.join(
+                exercise, f'{snake_case(slug)}_test.py'
+            )
+            rendered = template.render(**spec)
+            if check:
+                tmp_f, tmp_path = tempfile.mkstemp()
```
yawpitch: Seems to me we could do the diff without resort to subprocess, if we're just looking to see if the files would match.

cmccandless: I was inclined to think the `diff` tool would be faster at comparing whole files... although they are fairly small files, even for the larger exercises.

cmccandless:
> Since we're configuring flake to an 80 character limit we should do the same with black ... by default it's 88 characters and there will be occasional disconnect.

I followed [`black`'s recommendation on this](https://github.com/psf/black#line-length); line length is still 80, but there's a tolerance of 10%.

bin/generate_tests.py
yawpitch: The overhead of the subprocess will trump that until the files get _big_ ... if we're looking for _identical_ content then just opening any existing file and comparing content is sufficient.

yawpitch:
I do think that a slight modification of my branches generate.py file could be used to create a pretty good rough pass of the template text for quite a number of exercises, BTW.

cmccandless:
> though a few more docstrings would be nice

Done.

PR #1862, opened 2019-07-24
two-fer: add test template

cmccandless:


cmccandless:
The issue in the CI build is resolved in #1863. This PR should not be merged before that one.

PR #1863, opened 2019-07-24
hello-world: add test template

cmccandless:


yawpitch:
Try something like:

```python
from tempfile import NamedTemporaryFile
[...]
with NamedTemporaryFile("w", delete=False) as tmp:
    tmp.write(rendered)
format_file(tmp.name)
if check:
    try:
        if not filecmp.cmp(tmp.name, tests_path):
            logger.error(f'{slug}: check failed; tests must be regenerated with bin/generate_tests.py')
            sys.exit(1)
    finally:
        os.remove(tmp.name)
else:
    shutil.move(tmp.name, tests_path)
    print(f'{slug} generated at {tests_path}')
```

PR #1866, opened 2019-07-24
high-scores: add test template

cmccandless:


cmccandless:
Generating tests directly from canonical data has reminded me that we had a test we chose not to implement. The `scores` test is designed for class-based implementations, and is effectively an implementation of the identity function in function-based implementations like the one we follow here.

My inclination is to skip this test still, but it will require writing an exception into the template.

yawpitch:
Implementing `score` makes no obvious sense, so yeah I'd add the filter to the template. We're going to have this issue (along with name mapping) in a number of cases.

cmccandless:
@yawpitch unless you have any further comments, this PR is ready for merge. I can merge it later when I have the chance, or you can merge it on approval.

Issue #1745, opened 2019-04-13
high-scores: incorrect solution can pass the tests


ali1234:
Consider the following partial solution to this problem:

```
class HighScores(object):
    def __init__(self, scores):
        self.scores = scores

    def latest(self):
        return self.scores[-1]

    def personal_best(self):
        self.scores.sort()
        return self.scores[-1]
```

This will pass the tests for the functions it implements. However, if latest is called after calling personal_best, it can give the wrong answer. The current tests do not check for this situation.

cmccandless:
This issue has come up before in #1735.

Currently, side effect testing is outside the scope of this exercise. There is a discussion [here](https://github.com/exercism/problem-specifications/pull/1486) to include such tests, but I would like to wait and see the consensus there before adding these tests here.

Another solution discussed in #1744 would be to convert this track's implementation to use stand-alone functions rather than a class instance, which might make things more clear to students anyway.

simmol:
PR #1764 will be merged soon that will make the exercise simpler (using separate functions).

PR #1868, opened 2019-07-27
Test that original list is not modified for high scores tests

adamzev:
The mentor notes mention the importance of using `sorted()` instead of `sort()` (which mutates the list and makes subsequent calls to `latest` incorrect) but this case is not currently tested for and easy for mentors to miss.

### DIFF
--- a/exercises/high-scores/high_scores_test.py
+++ b/exercises/high-scores/high_scores_test.py
@@ -43,7 +43,7 @@ class HighScoresTest(unittest.TestCase):
         
     def test_all_do_not_modify_original_list(self):
         scores = [40, 20, 50, 30]
-        expected = scores.copy()
+        expected = [40, 20, 50, 30]
 
         personal_top_three(scores)
         self.assertEqual(scores, expected)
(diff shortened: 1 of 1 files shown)


cmccandless:
This is an issue that has been raised multiple times (#1735, #1745), and was one of the driving forces for converting the exercise from a class-based implementation to a funciton-based one.

If the mentor notes indicate that immutability is a requirement, they are wrong and need amended.

If the mentor notes simply state that immutability is a strong recommendation, that is perfectly valid; this exercise is an early core exercise, and is meant to be a simple practice in working with arrays. Immutability is a subject for a later exercise.

adamzev:
It is difficult for me to tell from the mentor notes what is a suggestion and what is a requirement for accepting the exercise. All the reasonable solutions use `sorted`. My assumption was that more was required than just passing the tests or the acceptance would have been automatic and based on that. I'll close this, read the other threads (I did try a search or two first, sorry about the duplicate) and look into clarifying the mentor notes.

adamzev:
> If the mentor notes indicate that immutability is a requirement, they are wrong and need amended.

I wouldn't say the current mentor notes are wrong, just a bit too open for interpretation. I created a [pull request on the notes](https://github.com/exercism/website-copy/pull/1195).

PR #1870, opened 2019-07-30
hamming: add test template

cmccandless:
- add 'is error_case' Jinja2 test
- create Jinja2 macros file, add utility method macro


cmccandless:
Note: using utility() macro for future template implementations will ensure that all exercises use the same source for utility methods.

PR #1871, opened 2019-08-01
rna-transcription: fix example.py

yawpitch:
Remove valid character test, as that's not required by spec any longer.
Also removed some unwieldy Python 2 compatibility code. When we drop support for Python 2 we can just change this to a **str.maketrans** call.

### DIFF
--- a/exercises/rna-transcription/example.py
+++ b/exercises/rna-transcription/example.py
@@ -1,18 +1,4 @@
-import sys
-
-if sys.version_info[0] == 2:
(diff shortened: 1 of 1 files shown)


cmccandless:
That "unweildy Python 2 code" seems to be required for our Python2 CI build.

yawpitch:
Ahh, how quickly we start to forget the idiosyncrasies of the old version. Sorry, thought I'd run both `pytest3` and `pytest2` locally.

PR #1873, opened 2019-08-03
high scores: Add testing to make sure methods don't mutate the scores list

andrewthetechie:
from a conversation in slack: https://exercism-team.slack.com/archives/CASLGCCSZ/p1564786254008900

This is a common issue we are seeing as python track mentors. This should catch students using pop to return values and doing sorts or reverse rather than sorted() or reverse()

### DIFF
--- a/exercises/high-scores/high_scores_test.py
+++ b/exercises/high-scores/high_scores_test.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 import unittest
 
 from high_scores import latest, personal_best, personal_top_three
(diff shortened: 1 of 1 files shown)


yawpitch:
For anyone that comes across this we are currently considering mutation acceptable in the function-based `high-scores` exercise. There is a PR to update the mentor notes to reflect this [here](https://github.com/exercism/website-copy/pull/1195) and also see prior discussion in #1868, #1745, and #1735.

PR #1877, opened 2019-08-04
Fix function name in nth prime

robkeim:
The tests are expecting the function to be named prime and not nth_prime:
https://github.com/exercism/python/blob/master/exercises/nth-prime/nth_prime_test.py


robkeim:
Thanks for the super quick review @yawpitch!

PR #1879, opened 2019-08-04
raindrops: update for track anatomy

### LINKED ISSUE #1859: raindrops: update example to better fit Track Anatomy Level 1 status
As part of Track Anatomy the `raindrops` exercise is being positioned to introduce conditionals (and the bool) _before_ `high-scores` introduces the `list` ... the example.py (and the mentor notes) should foreground that approach rather than the list comprehension syntax currently present.

### LINKED ISSUE #1878: raindrops: update description
With its move to a new position `raindrops` needs an updated README.md file that better explains what's involved in solving the exericse.

yawpitch:
Updates the example.py and README.md to conform to new position in the Core exercises.
Closes #1859 and closes #1878.

### DIFF
--- a/exercises/raindrops/example.py
+++ b/exercises/raindrops/example.py
@@ -1,10 +1,16 @@
-DROPS = ((3, 'Pling'), (5, 'Plang'), (7, 'Plong'))
-
-
-def convert(number):
+def convert(number: int) -> str:
     """
     Converts a number to a string according to the raindrop sounds.
     """
+  
+    result = ''
+    if number % 3 == 0:
+      result += 'Pling'
+    if number % 5 == 0:
+      result += 'Plang'
+    if number % 7 == 0:
+      result += 'Plong'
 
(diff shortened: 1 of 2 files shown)


yawpitch:
@F3PiX this PR updates the exercise README.md and the example solution (used for our internal tests).

exercises/raindrops/example.py
```diff
@@ -1,10 +1,16 @@
-DROPS = ((3, 'Pling'), (5, 'Plang'), (7, 'Plong'))
-
-
-def convert(number):
+def convert(number: int) -> str:
```
cmccandless: As awesome as type annotations are, they are not valid syntax in Python 2. Once we drop Python 2 (sometime in the next few months), this feature will be fine, but Travis fails when it's used right now.

yawpitch: Grr. Fixed.

PR #1881, opened 2019-08-05
isogram: add test template

cmccandless:
- add canonical_ref() generator macro

### DIFF
--- a/exercises/isogram/isogram_test.py
+++ b/exercises/isogram/isogram_test.py
@@ -2,11 +2,10 @@ import unittest
 
 from isogram import is_isogram
 
-
 # Tests adapted from `problem-specifications//canonical-data.json` @ v1.7.0
 
-class IsogramTest(unittest.TestCase):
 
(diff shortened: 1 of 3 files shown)


exercises/isogram/isogram_test.py
```diff
@@ -46,11 +45,6 @@ def test_duplicated_character_in_the_middle(self):
     def test_same_first_and_last_characters(self):
         self.assertIs(is_isogram("angola"), False)
 
-    # Additional tests for this track
-
-    def test_isogram_with_duplicated_letter_and_nonletter_character(self):
```
cmccandless: According to the test title, the purpose of this test is verifying correct behavior with duplicated non-letter characters. Since this track-specific test was added, the tests `isogram_with_duplicated_hyphen` and `made_up_name_that_is_an_isogram` were added and cover this behavior adequately, so the track-specific test is now redundant and can be dropped.

PR #1894, opened 2019-08-07
triangle: adding a new scalene test for two smaller sides equal

YuriyCherniy:
Adding a new test for scalene(sides) function and renaming another test to match the new test. Current tests for scalene(sides) function can be hacked by this code:
```
def scalene(sides):
    sides.sort()
    if sum(sides[:2]) < sides[2] or sides[1] == sides[2]:
        return False
    return True
```
while the new test not.


exercises/triangle/triangle_test.py
```diff
@@ -59,9 +59,12 @@ def test_no_sides_are_equal(self):
     def test_all_sides_are_equal(self):
         self.assertIs(scalene([4, 4, 4]), False)
 
-    def test_two_sides_are_equal(self):
+    def test_first_two_sides_are_equal(self):
```
cmccandless:
```suggestion
    def test_two_sides_are_equal(self):
```

exercises/triangle/triangle_test.py
```diff
@@ -59,9 +59,12 @@ def test_no_sides_are_equal(self):
...
         self.assertIs(scalene([4, 4, 3]), False)
 
+    def test_second_two_sides_are_equal(self):
```
cmccandless:
```suggestion
	# Additional tests for this track

    def test_two_sides_are_equal_and_third_side_is_larger(self):
```

cmccandless:

PR #1896, opened 2019-08-07
Use import with context for generators

cmccandless:
Import macros with context allows standard header/footer macros to access variables from the exercise spec so they do not need to be provided as arguments, simplifying test templates significantly.


cmccandless:
I recommend disabling whitespace diff on these changes; I refactored some error handling code to use a single `try` block.

PR #1900, opened 2019-08-07
Add track Contributing Guide and Generator documentation

cmccandless:


cmccandless:
> maybe replacing "fake_exercise" with something like "[EXERCISE]", and slightly re-arranging the order of the extra files to keep them related to the order of the files they modify.

Initially I had `example_exercise`, but that looked confusing with `example_exercise.py` and `example.py`. I like your idea better.

PR #1907, opened 2019-08-17
matrix: add tests for column matrix

weakish:


cmccandless:
Can you demonstrate a solution that passes the existing tests, but not the suggested ones? The existing test cases on a single-number matrix would seem to cover this already.

weakish:
If it is parsed into an generalized nested list based data structure, depending on the design of the data structure, `1 2 3` and `1\n2\n3` may be indistinguishable, both producing a flat list `[1 2 3]`. But what tests assumed is `[[1 2 3]]` and `[[1][2][3]]`. Thus there will be a shape mismatched.

The right direction is making the parsing always produce nested lists. But if instead trying to cope the problem via treating flat list specially, then it is possible to reach a solution that passes the existing tests, but not the suggested ones. (Because the special treatment only deals with the flat list shape mismatch issue, not the indistinguishable column matrix issue.)

numpy's array is such a data structure (roughly optimized implementation of nested list):

```python
from io import StringIO

import numpy as np


class Matrix(object):
    def __init__(self, matrix_string):
        self.value = np.loadtxt(StringIO(matrix_string), dtype="int", ndmin=1)

    def row(self, index):
        if self.value.ndim == 1:
            return [self.value[index - 1]]
        else:
            return list(self.value[index - 1, ...])

    def column(self, index):
        if self.value.ndim == 1:
            return [self.value[index - 1]]
        else:
            return list(self.value[..., index - 1])
```

The above code passed all existing test, but one of the suggested test
`MatrixTest.test_extract_column_from_a_column_matrix`:

```
>       self.assertEqual(matrix.column(1), [1, 2, 3])
E       AssertionError: Lists differ: [1] != [1, 2, 3]
E       
E       Second list contains 2 additional elements.
E       First extra element 1:
E       2
E       
E       - [1]
E       + [1, 2, 3]
```

cmccandless:
OK, I see where you're coming from. Your suggestion has called to my attention that, while the tests only deal with square matrices the description never states that the user need only handle NxN matrices. One of the following needs to occur:

- Add a statement to the exercise description that the user's program need not worry about non-square matrices
- Add tests for non-square matrices (such as this PR suggests)

For the Python track, at least, this exercise occurs quite early in the core progression, so my vote would be for the first option above. I'm going to create an issue in exercism/problem-specifications to address this.

@weakish, for this PR, I suggest we proceed as thus:

Since this exercise is [generated], adding test cases works a little differently (we are moving towards all exercises being generated; this procedure will be the new standard in this track). Please review the [contributing guide] for more information.
- The Travis build is failing because the modified tests file does not match the generator output; following convention will resolve this.
- The contributing guide is quite new; in fact, I believe this is the first PR of this kind since I wrote it up. If there is anything unclear, creating a new issue describing the confusion would be greatly appreciated.

If you have any further questions, I will be more than happy to address them as quickly as I can!

[generated]: https://github.com/exercism/python/blob/master/docs/GENERATOR.md
[contributing guide]: https://github.com/exercism/python/blob/master/CONTRIBUTING.md

weakish:
> so my vote would be for the first option above.

My vote is the same. :-)

cmccandless:
Since exercism/problem-specifications#1576 was merged, this test suite should be regenerated to include the added tests.

PR #1908, opened 2019-08-18
high-scores: validate that functions do not modify the list data

lekaf974:
Update the test suite to validate that the solution used are not modifying the list data.
With this test the students will have to not use the methods that can modify the list like sort(), pop(), reverse(), etc..

### DIFF
--- a/exercises/high-scores/high_scores_test.py
+++ b/exercises/high-scores/high_scores_test.py
@@ -41,6 +41,20 @@ class HighScoresTest(unittest.TestCase):
         expected = [40]
         self.assertEqual(personal_top_three(scores), expected)
 
+    def test_latest_not_modify_scores_data(self):
+        scores = [100, 0, 90, 30]
+        latest(scores)
+        self.assertEqual(scores, [100, 0, 90, 30])
+
+    def test_personal_best_not_modify_scores_data(self):
+        scores = [40, 100, 70]
+        personal_best(scores)
+        self.assertEqual(scores, [40, 100, 70])
+
+    def test_personal_top_three_not_modify_scores_data(self):
+        scores = [10, 30, 90, 30, 100, 20, 10, 0, 30, 40, 40, 70, 70]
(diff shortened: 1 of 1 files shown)


yawpitch:
Thank you for the PR, but I'm afraid I'm going to have to close this. As discussed in #1874, #1873, #1868, #1745, and #1735 we are _not_ going require the student to avoid mutation of the list passed to the various functions.

lekaf974:
Ok,

I did not read this previous PRs

Regards.

yawpitch:
Sorry about the wasted effort. We updated the [mentor notes](https://github.com/exercism/website-copy/blob/master/tracks/python/exercises/high-scores/mentoring.md) to reflect this, but we really need a better way of communicating this decision other than closed PRs. Unfortunately Github doesn't really provide a great mechanism for this.

cmccandless:
@yawpitch Perhaps we could leverage issue templates for that?

yawpitch:
@cmccandless we could try, but most of the PRs we've ended up closing have been ones created by people who didn't created an associated Issue first. They also didn't do a search of the closed PRs... I tend to think neither an Issue / PR Template nor a Pinned Issue will get any more attention.

What I had considered is whether or not we should simply add a track-specific comment along the lines of "## NOTE: DO NOT IMPLEMENT TESTS FOR MUTATION" to the generated test file.

Another way to look at this is that perhaps we've got it wrong and we _should_ ban mutation by informing the student that they should return a _copy_ of the original list. After all mutation is a fundamental issue the instant you introduce the list type... maybe we shouldn't keep trying to pretend otherwise.

cmccandless:
> Another way to look at this is that perhaps we've got it wrong and we should ban mutation by informing the student that they should return a copy of the original list. After all mutation is a fundamental issue the instant you introduce the list type... maybe we shouldn't keep trying to pretend otherwise.

Without endorsing one direction or the other: that would certain please several mentors who have raised this issue in the past.

PR #1909, opened 2019-09-03
Add test for all numeric name

cspollar:
Some student solutions are using string concatenation instead of a string formatting. Their solutions raise a TypeError when an integer is entered.

This test is designed to push students toward a more Pythonic/safer solution.

### DIFF
--- a/exercises/two-fer/two_fer_test.py
+++ b/exercises/two-fer/two_fer_test.py
@@ -15,6 +15,9 @@ class TwoFerTest(unittest.TestCase):
     def test_another_name_given(self):
         self.assertEqual(two_fer("Bob"), "One for Bob, one for me.")
 
+    def test_numeric_name_given(self):
+        self.assertEqual(two_fer(7), "One for 7, one for me.")
+
 
 if __name__ == "__main__":
     unittest.main()


yawpitch:
Thank you for the contribution, however I’m not sure we should be forcing the student to consider “nonsensical” user inputs, as then we’d need to apply that thinking to all the exercises that assume **str**  inputs (is `bob` and `kindergarten-garden`). IMHO it’s fine for solutions to this exercise to assume that either a **str** is provided as an argument for **name** or no value is received at all.

Concatenation is a valid, but less than ideal, solution to the problem as stated, so moving to a more general solution might make more sense as a mentor comment, in the same way as suggesting f-strings instead of str.format instead of prints-style interpolation. Also `two-fer` already has an auto-mentor that’s prepared to make T’ng law suggestions, so we really just need to make sure it’s doing so.

@cmccandless I haven’t looked at the auto-mentor in a while and I know there’s been some changes on the runner. Maybe we should re-check it’s outputs?

cmccandless:
@yawpitch I didn't think that auto-mentor disapproval comments were quite live yet...? I could be wrong here.

yawpitch:
Comments are being generated by the Python auto-analyser (https://exercism.io/mentor/analyses/20865), but status messages etc haven't been updated so that comments work on the site, so only the "approve_as_optimal" status has any effect.

The Go analyzer though has been making comments for a while (https://exercism.io/mentor/solutions/f7eedb776a5f4a10bebab6513af75eba) ... so I think we just need to update the anaylzer and re-deploy.

cspollar:
@yawpitch I agree, the auto-analyser looks like a great solution to this problem. Thanks for your feedback.

yawpitch:
Ok, so at the moment the _deployed_ auto-analyzer flags the use of concatenation as an issue, but doesn't end up actually rendering a comment that's useful on the site, however a few days ago there was a discussion on Slack on how we'll be improving the analyzer deployment mechanism (https://github.com/exercism/automated-analysis/issues/76), and once the changes in that issue have been implement we'll be able to push a version that _does_ make that comment meaningfully available.

So I'm going to close this PR as we don't want to handle this in tests. In the meantime we'll have to continue manually commenting on uses of concatenation, with the hope that _soonish_ we'll have an auto-generated prompt to reduce mentor overhead.

Thanks again for the submission.

PR #1916, opened 2019-09-19
Resolved Issue #1915

### LINKED ISSUE #1915: etl: update tests to v2.0.1
The [test suite](https://github.com/exercism/python/blob/master/exercises/etl/etl_test.py) for etl is out of date and needs updated to conform to the [latest canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/etl/canonical-data.json).

alirezaghey:
Currently, there are two minor issues with the generated test file which appear to be with the black auto formatter:
1- Dictionary entries are ALWAYS on separate lines, which makes them less easy to read when dealing with small entries. Like [here](https://github.com/exercism/python/compare/master...alirezaghey:issue-%231915?expand=1#diff-6162c6d7d91d9dc13cbdf39e188b596fR35)
2- These very same entries have a redundant comma after their last entry. Like [here](https://github.com/exercism/python/compare/master...alirezaghey:issue-%231915?expand=1#diff-6162c6d7d91d9dc13cbdf39e188b596fR60)

**edit**:
Resolved #1915

### DIFF
--- a/exercises/etl/.meta/template.j2
+++ b/exercises/etl/.meta/template.j2
@@ -3,7 +3,7 @@
     {%- set input = case["input"] -%}
     def test_{{ case["description"] | to_snake }}(self):
         legacy_data =  { {% for key, value in case["input"]["legacy"].items() %}
-                            {{key}}: {{value}}{{ "," if not loop.last else "" }}{% endfor %} }
+                            {{key}}: {{value}},{% endfor %} }
         data = {{case["expected"]}}
         self.assertEqual(
             {{ case["property"] | to_snake }}(legacy_data), data
--- a/exercises/etl/.meta/template.j2
+++ b/exercises/etl/.meta/template.j2
@@ -3,8 +3,7 @@
     {%- set input = case["input"] -%}
     def test_{{ case["description"] | to_snake }}(self):
         legacy_data =  { {% for key, value in case["input"]["legacy"].items() %}
-                             {{key}}: {{value}}, 
-                        {% endfor %} }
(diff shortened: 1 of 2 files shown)


alirezaghey:
Great! Thank you for your swift review.
Two questions though:
1- Are the formattings part of the Black auto formatter?
2- Why's the CI check failing? Is this a problem with the checks in the generator file?

yawpitch:
Yes the formatting is black’s doing. We can expect it to make some changes relative to the hand-made templates. We could reduce that diff noise by going through and doing a PR that ran black on every non-auto-generated test file, but we’ve been handling it exercise by exercise as we introduce generator templates.

As for the failure that appears to be a bit of a race condition with #1914 and #1912 not being completed. I’ll try and clear that up today if possible.

alirezaghey:
Resolves issue #1915

exercises/etl/.meta/template.j2
```diff
@@ -0,0 +1,20 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{% macro test_case(case) -%}
+    {%- set input = case["input"] -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        legacy_data =  { {% for key, value in case["input"]["legacy"].items() %}
+                            {{key}}: {{value}}{{ "," if not loop.last else "" }}{% endfor %} }
```
cmccandless: Since `black` is going to add the comma anyway, might I suggest simplifying the template?

```suggestion
                            {{key}}: {{value}},{% endfor %} }
```

PR #2012, opened 2019-10-02
wordy: add test template

dvermd:
After some digging, here is a template for wordy exercise #2002

### DIFF
--- a/exercises/wordy/.meta/template.j2
+++ b/exercises/wordy/.meta/template.j2
@@ -4,17 +4,15 @@
 class {{ exercise | camel_case }}Test(unittest.TestCase):
     {% for case in cases -%}
     def test_{{ case["description"] | to_snake }}(self):
+        {%- set question = case["input"]["question"] %}
         value = "{{ case["input"]["question"] }}"
         {%- if case is error_case %}
         with self.assertRaisesWithMessage(ValueError):
-            {{ case["property"] }}(value)
+            {{ case["property"] }}("{{ question }}")
         {%- else %}
-        expected = {{ case["expected"] }}
-        self.assertEqual({{ case["property"] }}(value), expected)
+        self.assertEqual({{ case["property"] }}("{{ question }}"), {{ case["expected"] }})
         {%- endif %}
 
     {% endfor %}
 
-    {{ macros.utility() }}
-
-{{ macros.footer() }}
+{{ macros.footer(has_error_case) }}
--- a/exercises/wordy/.meta/template.j2
+++ b/exercises/wordy/.meta/template.j2
@@ -1,0 +1,20 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.header() }}
+
+class {{ exercise | camel_case }}Test(unittest.TestCase):
+    {% for case in cases -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        value = "{{ case["input"]["question"] }}"
+        {%- if case is error_case %}
+        with self.assertRaisesWithMessage(ValueError):
+            {{ case["property"] }}(value)
+        {%- else %}
+        expected = {{ case["expected"] }}
+        self.assertEqual({{ case["property"] }}(value), expected)
+        {%- endif %}
+
+    {% endfor %}
+
+    {{ macros.utility() }}
+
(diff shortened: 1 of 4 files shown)


exercises/wordy/example.py
```diff
@@ -13,7 +13,7 @@
 
 def answer(question):
     if not (question.startswith("What is ") and question.endswith("?")):
-        raise ValueError("Ill-formed question")
+        raise ValueError("unknown operation")
```
cmccandless: Please revert this. The correct message was already being used.
```suggestion
        raise ValueError("Ill-formed question")
```

exercises/wordy/.meta/template.j2
```diff
@@ -0,0 +1,20 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.header() }}
+
+class {{ exercise | camel_case }}Test(unittest.TestCase):
+    {% for case in cases -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        value = "{{ case["input"]["question"] }}"
```
cmccandless:
```suggestion
        {%- set question = case["input"]["question"] -%}
```

exercises/wordy/.meta/template.j2
```diff
@@ -0,0 +1,20 @@
...
+        {%- if case is error_case %}
+        with self.assertRaisesWithMessage(ValueError):
+            {{ case["property"] }}(value)
```
cmccandless:
```suggestion
            {{ case["property"] }}({{ question }})
```

exercises/wordy/.meta/template.j2
```diff
@@ -0,0 +1,20 @@
...
+        {%- else %}
+        expected = {{ case["expected"] }}
```
cmccandless:
```suggestion
```

exercises/wordy/.meta/template.j2
```diff
@@ -0,0 +1,20 @@
...
+        self.assertEqual({{ case["property"] }}(value), expected)
```
cmccandless:
```suggestion
        self.assertEqual({{ case["property"] }}({{ question }}), {{ case["expected"] }})
```

exercises/wordy/.meta/template.j2
```diff
@@ -0,0 +1,20 @@
...
+        {%- endif %}
+
+    {% endfor %}
+
+    {{ macros.utility() }}
```
cmccandless: See Suggested Change for line 20.
```suggestion
```

exercises/wordy/.meta/template.j2
```diff
@@ -0,0 +1,20 @@
...
+
+{{ macros.footer() }}
```
cmccandless: `has_error_case` is a built-in variable for the generator that is `True` if there is any test case that expects an error. The `footer` macro takes it as an optional parameter.

```suggestion
{{ macros.footer(has_error_case) }}
```

dvermd: This one, I copied it from the hamming exercise. To stick with this advice, we could create an issue and PR for that ?

dvermd:
I've updated my code, should I add another commit not to loose the review history or do you prefer I squash them and push force ?

exercises/wordy/.meta/template.j2
cmccandless: There's actually a larger issue that footer macro shouldn't need that variable to be passed in, but removing it will break some existing templates, so that's a larger fix.

cmccandless:
Add another commit please.

dvermd:
The new commit is here, should I mark all the conversations as resolved or this is not required ?

exercises/wordy/.meta/template.j2
```diff
@@ -4,17 +4,15 @@
 class {{ exercise | camel_case }}Test(unittest.TestCase):
     {% for case in cases -%}
     def test_{{ case["description"] | to_snake }}(self):
+        {%- set question = case["input"]["question"] %}
         value = "{{ case["input"]["question"] }}"
```
cmccandless: The `value = ...` line is still there; please remove it.

PR #2016, opened 2019-10-03
triangle : add test template

mpatibandla:


exercises/triangle/.meta/template.j2
```diff
@@ -0,0 +1,14 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.header(["equilateral", "isosceles", "scalene"]) }}
+
+{% for case in cases -%}
+class {{ case["description"] | camel_case }}Test(unittest.TestCase):
+    {% for case in case["cases"] %}
+    def test_{{ case["description"] | to_snake }}(self):
+        input = {{ case["input"]["sides"] }}
+        expected = {{ case["expected"] }}
+        self.assertEqual({{ case["property"] }}(input), expected)
```
cmccandless: I'd recommend putting everything inside the `assertEqual` call instead of using the `input` and `expected` variables here. You can spread things out like so for template readability, and the generator will put everything on a single line in generation:
```
self.assertEqual(
    {{ case["property"] }}({{ case["input"]["sides"] }}),
    {{ case["expected"] }}
)
```

mpatibandla:
Made changes as requested. Thanks !

exercises/triangle/.meta/template.j2
```diff
@@ -0,0 +1,12 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{{ macros.header(["equilateral", "isosceles", "scalene"]) }}
+
+{% for case in cases -%}
+class {{ case["description"] | camel_case }}Test(unittest.TestCase):
+    {% for case in case["cases"] %}
+    def test_{{ case["description"] | to_snake }}(self):
+        self.assertEqual({{ case["property"] }}({{ case["input"]["sides"] }}), {{ case["expected"] }})
```
cmccandless: OK, smaller diff let me see this one. From our [conventions](https://github.com/exercism/python#conventions):
> We use assertIs(actual, True) and assertIs(actual, False) rather than assertTrue(actual) or assertFalse(actual) (#419).

```suggestion
        self.assertIs({{ case["property"] }}({{ case["input"]["sides"] }}), {{ case["expected"] }})
```

mpatibandla: changed to assertIs as per conventions. Thanks.

Issue #2026, opened 2019-10-03
Template generation does not work in Windows


crsmi:
When running `bin/generate_tests.py -v [test-name]` in a Windows environment an error is thrown.

```
DEBUG:generator:zebra-puzzle\.meta\template.j2
INFO:generator:zebra-puzzle: no template found; skipping
```

This is because the Jinja environment/loader expects forward slashes instead of backslashes.

I think this can be easily fixed by adding `import posixpath` and using `posixpath.join` instead of `os.path.join` for creating the template string.

cmccandless:
> I think this can be easily fixed by adding import posixpath and using posixpath.join instead of os.path.join for creating the template string.

Sounds like another PR in the works! Try it, test it, push it!

crsmi:
Yeah, I found this out while working on #2004. The `posixpath` fix worked for me.

In the future, should I open an issue if I can easily have a pull request ready (for documentation purposes), or should I just create the pull request?

cmccandless:
You can just create the pull request, but it would be nice if any information that would've been in the issue is also included in the PR (screenshots of errors, etc).

crsmi:
Cool, I'll make sure to do that in the future. Thanks for the feedback!

PR #2027, opened 2019-10-03
Update template path formatting

crsmi:
Allows for template creation in a Windows environment.

The Jijna environment/loader expects forward slashes for the path used to find a template. In a Windows environment `os.path.join` produces a string with backslashes. We can use `posixpath.join` instead to make sure that the template path is always created with forward slashes.

### DIFF
--- a/bin/generate_tests.py
+++ b/bin/generate_tests.py
@@ -17,6 +17,7 @@ import filecmp
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


cmccandless:
Thanks! We love cross-platform compatibility fixes!

crsmi:
Closes #2026

PR #2031, opened 2019-10-04
Generalize test templates

dvermd:
This PR generalize even more the test creation leaving only the testcase to be written in `template.j2`.
It handles additional and hardcoded tests.
Imports and ignore can still be used.


dvermd:
The checks seem to fail because of network error. I don't how or even if I have the credentials to rerun them.

cmccandless:
I've restarted the failed build.

yawpitch:
I’m going to have to state that a little more firmly: we do not need to generalize “hard-coded” tests; we can just incorporate them into the template file directly.

And that is IFF we can even support the argument those tests need to exist.

This test does not need to exist. The dummy file should be updated to set a unique integer value for each of the “enum” values, and a comment in the file should inform the student not to modify those values such that they’re not unique. By this point in their journey the student does not need to know how to assign four different constants to four different names and there’s no pedagogical value in forcing them to do it.

Similar reasoning can likely be applied to any other such case we encounter. If not we should instead expand the “additional tests” functionality rather than shoehorn in a section of template.

dvermd:
The initial commit was on top of another PR that got merged by the time the new commits arrived.
Locally, I rebased these new commits on exercism/master but instead of force pushing to my branch, I used the github gui to resolve the conflicts which ended up with a merge commit.
Do you prefer I push force ?

dvermd:
I forgot the has_error_case parameter of the footer macro. This commit fixes that.

cmccandless:
I think removing the `sublist` changes from this PR is a good approach.

dvermd:
The sublist changes in this PR were merely there as a usage demo, but should never have been interleaved with master_template changes. 

I reverted them from this PR.

The commit ba25e6ef043222dd64844477ab15742c7506e122 may also been reverted as it should not be needed anymore thanks to #2084.

Waiting for this to be merged before creating the sublist PR.

config/master_template.j2
```diff
@@ -0,0 +1,15 @@
...
+    {{ test_case(case) }}
+    {% endfor %}
+
+    {% if additional_cases | length -%}
+    # Additional tests for this track
+    {% for case in additional_cases -%}
+    {{ test_case(case) }}
+    {% endfor %}
+    {%- endif %}
+{{ macros.footer(has_error_case) }}
```
cmccandless: This is implicit now. For a template that extends this one, it should be possible to override this with `{%- set has_error_case = True -%}` if necessary.

```suggestion
{{ macros.footer() }}
```

config/master_template.j2
```diff
@@ -0,0 +1,15 @@
...
+
+class {{ exercise | camel_case }}Test(unittest.TestCase):
+    {% for case in cases -%}
+    {{ test_case(case) }}
+    {% endfor %}
+
+    {% if additional_cases | length -%}
+    # Additional tests for this track
+    {% for case in additional_cases -%}
+    {{ test_case(case) }}
```
cmccandless: It might be a good idea to define a macro above for `additional_test_case` that uses `test_case`, but may be overridden by an extending template if necessary. It is common for the additional cases to follow a different structure than the canonical ones.

config/master_template.j2
dvermd: Commit reverted

config/master_template.j2
dvermd: The code of master template does not handle nested cases. 
I added a commit to handle them correctly. Is this OK or should I still create the macro you asked for ?

cmccandless: > It is common for the additional cases to follow a different structure than the canonical ones.

What I meant was that the additional test cases sometimes have a different formula. Contrived Example:
```
class WeirdAdditionalTest(unittest.TestCase):
    def test_normal_case1(self):
        self.assertEqual(property(CONSTANT1), expected)

    def test_normal_case2(self):
        self.assertEqual(property(CONSTANT2), expected)

	# Additional tests for this track
	def test_additional_case(self):
        self.assertNotEqual(CONSTANT1, CONSTANT2)
```

If you want to properly handle nested cases, I suggest something like the following, since nested `cases` arrays can be multiple levels deep:
```
{%- macro test_cases_recursive(cases) -%}
{% for case in cases -%}
	{% if "cases" in case -%}
# case["description"]
{{ test_cases_recursive(case["cases"]) }}
	{% else -%}
{{ test_case(case) }}
	{% endif -%}
{% endfor -%}
{% endmacro %}

    {{ test_cases_recursive(cases) }}
```

PR #2036, opened 2019-10-06
Call black as a lib instead of an external tool

dvermd:
This PR shows how to use black as a lib instead of an external tool which needs a process to be spawned.
It results on a 0,25s performance improvement on my machine.

One thing to consider for this PR is that black is still tagged as beta. Its CLI API used today may be less prone to change than its internal API.
Hope this still help.

### DIFF
--- a/bin/generate_tests.py
+++ b/bin/generate_tests.py
@@ -24,7 +24,8 @@ import sys
 from glob import glob
 from itertools import repeat
 from string import punctuation, whitespace
-from subprocess import CalledProcessError, check_call
+from black import reformat_one, WriteBack, FileMode, Report
+from pathlib import Path
 from tempfile import NamedTemporaryFile
 
 from jinja2 import Environment, FileSystemLoader, TemplateNotFound, UndefinedError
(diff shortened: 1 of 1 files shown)


yawpitch:
To be honest I'd be more comfortable relying on the CLI interface, as they've not documented the use as a library and the CLI seems to be the only thing they have any investment in maintaining as stable. The most recent change on the reformat_one interface completely re-ordered arguments, and while that was 8 months ago I'd not be confident it wouldn't happen again.

yawpitch:
@cmccandless what do you think?

cmccandless:
I'm inclined to keep using the CLI. That seems to be the standard recommended usage.

dvermd:
I understand your position and maintenance anticipation. 
Though this PR won't make it to master, I had fun reading through black source code and integrating it.

cmccandless:
In academic interest: do your changes also cause the generator to exit with a non-zero code if black fails to format the file (most common case for this is invalid Python). If not, that might be a fun exercise as well.

dvermd:
You're right, it did not exit with non-zero code and I just fixed it to.

cmccandless:
Another possibility for avoiding the subprocess spin-up per exercise:

https://github.com/psf/black#blackd

dvermd:
Launching blackd is one spawn, this optimization would only be interested when generating many/all exercises ? 
Would that be of any interest for exercism/python to have that ? 
Anyway, I might give it a try to play with requests or pycurl.

yawpitch:
There _might_ be an argument for `blackd` in the CI process worker when building _many_ tests ... but realistically reducing the call overhead is not a priority.

Issue #2011, opened 2019-10-02
Generated tests reference a wrong path for canonical-data.json


dvermd:
The tests generated with template.j2 contains the comment
`# Tests adapted from problem-specifications//canonical-data.json @ vx.y.z`

The path should contain the exercise name.

I have a fix for that one

cmccandless:
This is intentional to keep the comment consistent and short. Technically, the full path is:
`problem-specifications/exercises/<exercise>/canonical-data.json`

But for the longest exercise name (parallel-letter-frequency), this is incredibly long, so we truncate the path for readability.

dvermd:
Ok, maybe we can add some `...` between the `//`

cmccandless:
I'd prefer to wait until all/most exercises have a template to change this line. There are tools that depend on the format of this line, so every exercise and the dependent tools need to be updated at the same time.

PR #2055, opened 2019-10-14
Add exercise name to the header comment

GKotfis:
This small change will add exercise name to the generated comment in header:
Before:
```python
# Tests adapted from `problem-specifications//canonical-data.json` @ v3.2.0
```

After:
```python
# Tests adapted from `problem-specifications//robot-simulator//canonical-data.json` @ v3.2.0
```

### DIFF
--- a/config/generator_macros.j2
+++ b/config/generator_macros.j2
@@ -10,7 +10,7 @@
 {% endmacro -%}
 
 {% macro canonical_ref() -%}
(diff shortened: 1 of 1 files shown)


dvermd:
@GKotfis, you may look at #2011 to see that this is indeed the desired behavior

yawpitch:
Closing this one because, yes, we don't want to make such a change until all / most of the exercises have a template, at which point we can do it en masse.

PR #2057, opened 2019-10-14
rna_transcription: add test template

dvermd:
Resolves: #1979

### DIFF
--- a/exercises/rna-transcription/.meta/template.j2
+++ b/exercises/rna-transcription/.meta/template.j2
@@ -1,26 +1,13 @@
 {%- import "generator_macros.j2" as macros with context -%}
-{% macro test_case(case) -%}
-    {%- set input = case["input"] -%}
-    def test_{{ case["description"] | to_snake }}(self):
-        self.assertEqual(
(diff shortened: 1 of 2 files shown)


exercises/rna-transcription/.meta/template.j2
```diff
@@ -0,0 +1,26 @@
...
+    {{ test_case(case) }}
+    {% endfor %}
+    {% if additional_cases | length -%}
+
+    # Additional tests for this track
+
+    {% for case in additional_cases -%}
+    {{ test_case(case) }}
+    {% endfor %}
+    {%- endif %}
```
cmccandless: This section for additional tests may be removed; there are no such tests in this exercise.

dvermd: I just removed this section. As the section was surrounded by an if additional_cases, I thought that would be OK.

cmccandless: I didn't technically hurt, it was just code that never runs, and doesn't need to be there.

PR #2058, opened 2019-10-14
saddle-points: add test template

### LINKED ISSUE #1984: saddle-points: add test template
## Task
Create a test template file for `saddle-points`

### Details
1. If this is your first time contributing to this repository, welcome!
    - Contributing to a repository on GitHub requires a [fork](https://help.github.com/en/articles/fork-a-repo)
1. Create a branch on your local repository called `saddle-points-add-test-template`
    - Run: `git checkout -b saddle-points-add-test-template`
1. Clone problem-specifications repository; this is used to populate the template
    - Run: `git clone https://github.com/exercism/problem-specifications ../problem-specifications`
1. Create `exercises/saddle-points/.meta/template.j2`
    - See [Generator Documentation] for more information
1. Generate test file from new template
    - Run: `bin/generate_tests.py saddle-points`

At this point, if the generator fails, or you are otherwise stuck,
please feel free to reach out! It is perfectly acceptable to submit
an incomplete PR in order to get feedback or assistance.

### Resources
- [Contributing Guide]
- [Generator Documentation]
- [Closing issues using keywords]

[Contributing Guide]: https://github.com/exercism/python/blob/master/docs/CONTRIBUTING.md
[Generator Documentation]: https://github.com/exercism/python/blob/master/docs/GENERATOR.md
[Closing issues using keywords]: https://help.github.com/en/articles/closing-issues-using-keywords

crsmi:
Resolves #1984 

This one was a bit weird. The canonical data seems to be a little bit off compared to the previous test file. I'm wondering if the following are okay.

- [ ] The test names are all a bit different.
- [ ] There is one less test in canonical data than the old test file: `test_matrix_with_one_elem_has_single_saddle_point`
- [ ] I had to sort the "expected" values for the `test_can_identify_saddle_points_in_a_non_square_matrix` test. I'm not sure this is ideal. (Line 28 in the template)


exercises/saddle-points/.meta/template.j2
```diff
@@ -0,0 +1,58 @@
...
+
+{% macro list_points(points) -%}
+    {% if points -%}
+    {{ points | sort(attribute="column") }}
+    {% else -%}
+    [{}]
+    {% endif %}
+{%- endmacro -%}
+
+{% macro error_case(case) -%}
```
cmccandless: `error_case` is a build-in filter for the generator (example usage below); it is best practice not to override built-ins.
```
{% if error_case(case) -%}
with self.assertRaisesWithMEsage(ValueError):
    {{ case["property"] | to_snake }}(matrix)
{% else -%}
self.assertEqual(
	{{- case["property"] | to_snake }}(matrix),
	{{- list_points(expected) }})
{% endif -%}
```

exercises/saddle-points/.meta/template.j2
```diff
@@ -0,0 +1,58 @@
...
+        {% if input["matrix"][0] -%}
+        matrix = {{ input["matrix"] }}
+        {%- else -%} matrix = [] {% endif %}
+        {% if expected|length > 2 -%}
+        expected = {{ list_points(expected) }}
+        self.assertEqual({{ case["property"] | to_snake }}(matrix), expected)
+        {% else -%}
+        self.assertEqual(
+            {{- case["property"] | to_snake }}(matrix), {{ list_points(expected) }})
+        {%- endif %}
```
cmccandless: To make maintaining the template easier, all non-error tests may directly embed `list_points(expected)` in the `assertEqual` call (currently inside the `{% else %}` here). No need to check length of `expected`.

exercises/saddle-points/.meta/template.j2
```diff
@@ -0,0 +1,58 @@
...
+"""
+{%- import "generator_macros.j2" as macros with context -%}
+
+{% macro test_case(case) -%}
+    {%- set input = case["input"] -%}
+    {%- set expected = case["expected"] -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        {% if input["matrix"][0] -%}
+        matrix = {{ input["matrix"] }}
+        {%- else -%} matrix = [] {% endif %}
```
cmccandless: For readability, please adjust the spacing on this:
```suggestion
        {%- else -%}
        matrix = []
        {% endif -%}
```

exercises/saddle-points/.meta/template.j2
```diff
@@ -0,0 +1,58 @@
...
+        self.assertEqual(
+            {{- case["property"] | to_snake }}(matrix), {{ list_points(expected) }})
+        {%- endif %}
+{%- endmacro -%}
+
+{% macro list_points(points) -%}
+    {% if points -%}
+    {{ points | sort(attribute="column") }}
+    {% else -%}
+    [{}]
```
cmccandless: I am not sure why the current implementation prefers `[{}]` over `[]` for an "empty" result; I am in favor of changing this (requires change to example.py). This would eliminate the need for this macro altogether.

@yawpitch, are you good with changing the expectation for empty results?

yawpitch: @cmccandless yeah, I've no idea why that was ever done. The canonical data seems to expect just an empty list anyway, so eliminating the disconnect seems sensible.

exercises/saddle-points/.meta/template.j2
crsmi: Forgot about the built in! I moved the logic to check for the error case into the `test_case` macro.

exercises/saddle-points/.meta/template.j2
crsmi: Yeah, I thought that was a little weird. I've made the change.

I also switched from using `assertEqual` to `assertCountEqual` which will handle the possibility of the points being in a different order in the list and remove the need to sort.

exercises/saddle-points/.meta/template.j2
crsmi: Cleaned up.

exercises/saddle-points/.meta/template.j2
crsmi: Cleaned up.

exercises/saddle-points/.meta/template.j2
crsmi: Looks like `assertCountEqual` is not in Python 2.7. Reverted to sorting, but I added a more robust sorting logic than previously.

exercises/saddle-points/.meta/template.j2
```diff
@@ -0,0 +1,54 @@
...
+        {% endif -%}
+{% endmacro -%}
+
+{% macro list_points(points) -%}
+    {% if points -%}
+    {{ points | sort(attribute="column") }}
+    {% else -%}
+    []
+    {% endif %}
+{%- endmacro %}
```
cmccandless: If the `list_points` macro is no longer used, please remove it.

exercises/saddle-points/.meta/template.j2
```diff
@@ -0,0 +1,46 @@
...
+        {% else -%}
+        matrix = []
+        {% endif -%}
+        {% if case is error_case -%}
+        with self.assertRaisesWithMessage(ValueError):
+            {{- case["property"] | to_snake }}(matrix)
+        {% else -%}
+        self.assertEqual(
+        sorted({{ case["property"] | to_snake }}(matrix), key=lambda d: (d["row"],d["column"])),
+        sorted({{ expected }}, key=lambda d: (d["row"],d["column"]))
```
cmccandless: Getting cleaner... Since every test case is using this sort call twice, why don't we extract the sorting to a separate function? Something like the following:
```

{% macro test_case(case) -%}
...
self.assertEqual(
    sorted_points({{ case["property"] | to_snake }}(matrix)),
    sorted_points({{ expected}})
)
...
{% endmacro -%}


def sorted_points(point_list):
    return sorted(point_list, key=lambda p: (p["row"], p["column"]))


```

exercises/saddle-points/.meta/template.j2
```diff
@@ -0,0 +1,46 @@
...
+        )
+        {% endif -%}
+{% endmacro %}
```
cmccandless:
```suggestion
{% endmacro -%}
```

To remove the added blank line before `import unittest`

crsmi: Done. Had to also, remove the white space eliminator from the beginning of the import macros call so that there still be a single newline between the comment and the first line of code.

exercises/saddle-points/.meta/template.j2
crsmi: That makes sense. Updated!

PR #2061, opened 2019-10-15
generator: handle no local black install

### LINKED ISSUE #2060: Test generator reports "template not found" if "black" is not installed
### File
bin/generate_tests.py
### Procedure
On a system with no local installation of `black`, run `python3.7 bin/generate_tests.py two-fer`
### Expected
An error stating that `black` is not installed
### Actual
An error stating the template for `two-fer` is not found.

yawpitch:
Fix incorrect error message when black is not installed. Decided to handle it in two locations, just in case someone uses this as a library and not via the CLI.

Resolves #2060


cmccandless:
Looks good. I'll merge it once the CI check finishes (had to update branch).

PR #2065, opened 2019-10-18
nth-prime: add template

smalley:
Add meta template and additional_tests.json to generate nth-prime test file including all tests currently contained in the nth_prime_test.py file. test_case macro is split conditionally to handle error cases, single number cases and ranges of primes. Closes: #1962

### DIFF
--- a/exercises/nth-prime/.meta/additional_tests.json
+++ b/exercises/nth-prime/.meta/additional_tests.json
@@ -1,0 +1,13 @@
+{
+  "cases": [
+    {
+      "description": "first twenty primes",
+      "property": "primeRange",
+      "input": {
+        "number": 20
+      },
+      "expected": [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31,
+                   37, 41, 43, 47, 53, 59, 61, 67, 71]
+    }
+  ]
+}
--- a/exercises/nth-prime/.meta/template.j2
+++ b/exercises/nth-prime/.meta/template.j2
@@ -1,0 +1,35 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{% macro test_case(case) -%}
+    {%- set input = case["input"] -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        {% if case is error_case -%}
+          with self.assertRaisesWithMessage(ValueError):
+                {{ case["property"] | to_snake }}({{ input["number"] }})
+        {% else -%}
+          self.assertEqual(
+                {{ case["property"] | to_snake }}({{ input["number"] }}),
+                {{ case["expected"] }}
+            )
+        {%- endif %}
+{%- endmacro %}
+{{ macros.header()}}
+
+def prime_range(n):
+    """Returns a list of the first n primes"""
+    return [prime(i) for i in range(1, n + 1)]
+
+class {{ exercise | camel_case }}Test(unittest.TestCase):
+    {% for case in cases -%}       
+    {{ test_case(case) }}
+    {% endfor %}
+    {% if additional_cases | length -%}
+
+    # Additional tests for this track
+
+    {% for case in additional_cases -%}
+    {{ test_case(case) }}
+    {% endfor %}
+    {%- endif %}
+
+
+{{ macros.footer(has_error_case) }}
--- a/exercises/nth-prime/nth_prime_test.py
+++ b/exercises/nth-prime/nth_prime_test.py
@@ -2,8 +2,13 @@ import unittest
 
 from nth_prime import prime
 
-
 # Tests adapted from `problem-specifications//canonical-data.json` @ v2.1.0
+
+
+def prime_range(n):
+    """Returns a list of the first n primes"""
+    return [prime(i) for i in range(1, n + 1)]
+
 
 class NthPrimeTest(unittest.TestCase):
     def test_first_prime(self):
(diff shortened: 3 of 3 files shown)


exercises/nth-prime/nth_prime_test.py
```diff
@@ -22,11 +22,34 @@ def test_there_is_no_zeroth_prime(self):
         with self.assertRaisesWithMessage(ValueError):
             prime(0)
 
-    # additional track specific test
+    # Additional tests for this track
+
     def test_first_twenty_primes(self):
```
cmccandless: I'm not sure why this test existed to begin with; it doesn't add additional coverage. If getting the 10001st prime works (case `test_big_prime` above), this is redundant. I would vote to remove the additional test altogether.

@yawpitch thoughts?

exercises/nth-prime/.meta/template.j2
```diff
@@ -0,0 +1,35 @@
...
+            self.assertEqual(
+                {{ case["property"] | to_snake }}({{ input["number"] }}),
+                {{ case["expected"] }}
+            )
+        {% elif case["expected"]["error"] is string -%}
+            with self.assertRaisesWithMessage(ValueError):
+                {{ case["property"] | to_snake }}({{ input["number"] }})
+        {% else -%}
+           self.assertEqual( {{ case["expected"] }},
+                {{'['}}{{ case["property"] | to_snake }}(n) for n in range(1,{{case["expected"]|length + 1}}){{']'}})
```
cmccandless:
```suggestion
                [{{ case["property"] | to_snake }}(n) for n in range(1,{{case["expected"]|length + 1}})])
```

exercises/nth-prime/.meta/template.j2
```diff
@@ -0,0 +1,35 @@
...
+    {% if additional_cases | length -%}
+
+    # Additional tests for this track
+
+    {% for case in additional_cases -%}
+    {{ test_case(case) }}
+    {% endfor %}
+    {%- endif %}
+
+{{ macros.utility() }}
```
cmccandless:
```suggestion
```

exercises/nth-prime/.meta/template.j2
```diff
@@ -0,0 +1,35 @@
...
+
+{{ macros.footer() }}
```
cmccandless:
```suggestion
{{ macros.footer(has_error_case) }}
```

exercises/nth-prime/nth_prime_test.py
yawpitch: If I were to guess it's to keep someone from writing:

```python
def prime(n):
    if n == 1:
        return 2
    if n == 2:
        return 3
    if n == 6:
        return 13
    if n == 10001:
        return 10474
    raise ValueError("LOL, whut?")
```

... without any friction.

Arguably there probably should be a test for the first N primes, but I see no reason for that to be a track-specific choice.

cmccandless: I hadn't thought of that. In that case, perhaps keep the additional test, but call it by a different property? Our current structure would allow the following, I think:
```JSON
{
  "cases": [
    {
      "description": "first twenty primes",
      "property": "(lambda n: [prime(n) for n in range(1, n + 1))",
      "input": {
        "number": 20
      },
      "expected": [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31,
                   37, 41, 43, 47, 53, 59, 61, 67, 71]
    }
  ]
}
```

cmccandless: The above would need to somehow bypass the `to_snake` filter for `property`

cmccandless: Another option would be to just hard-code the below into the template, then reference `prime_range` as the property:

```Python
def prime_range(n):
"""Returns a list of the first n primes"""
    return [prime(i) for i in range(1, n + 1)]
```

smalley: I went the route of hard coding the prime_range support function into the template and calling that through the additional test case. Doing so let me drop an additional case from the test function macro which seemed a little cleaner.

exercises/nth-prime/.meta/additional_tests.json
```diff
@@ -0,0 +1,13 @@
+{
+  "cases": [
+    {
+      "description": "first twenty primes",
+      "property": "prime range",
```
cmccandless:
```suggestion
      "property": "primeRange",
```

cmccandless: A bit nit-picky, but the canonical data uses camelCase for its test names. The idea is to keep additional_tests.json as close to the format of canonical-data.json as possible.

exercises/nth-prime/.meta/template.j2
```diff
@@ -0,0 +1,36 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{% macro test_case(case) -%}
+    {%- set input = case["input"] -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        {% if not case["expected"]["error"] is string -%}
```
cmccandless: There's a built-in filter you can use here:
```suggestion
        {% if not case is error_case -%}
```

For readability, I would suggest switching your cases around so you can just use `{% if case is error_case -%}`

smalley: wow, that is a much better way to do that! I totally missed that generate_test.py was adding a filter for that. Thanks for pointing that out!

cmccandless: Documenting which non-standard filters are built in to the generator is somewhere on my (very long) TODO list.

PR #2074, opened 2019-10-20
raindrops: fix example indentation

bsoyka:
Fixed indentation error in example file.


bsoyka:
Related to exercism/website-copy#1354

PR #2080, opened 2019-10-22
rational-numbers add template

GKotfis:
Resovles #1975 

I've a problem with template and `if/elif` conditions. Maybe you can help me why `if` conditions only works for 'Arithmetic' description? When you look at `rational_numbers_test.py`: 74 the tests are only rendered for for that group of cases. Other like 'Absolute value' are omitted.

### DIFF
--- a/exercises/rational-numbers/.meta/template.j2
+++ b/exercises/rational-numbers/.meta/template.j2
@@ -5,7 +5,8 @@
     "sub": "-",
     "mul": "*",
     "div": "/",
-    "exprational": "**"
+    "exprational": "**",
+    "expreal": "**"
 } -%}
 
 {% macro test_case_arithmetic(case) -%}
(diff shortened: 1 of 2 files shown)


exercises/rational-numbers/.meta/template.j2
```diff
@@ -0,0 +1,40 @@
...
+{%- endmacro %}
+
+{% macro test_case_exponentiation(case) -%}
+    {%- set input = case["input"] -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        # xxx
+        pass
+{%- endmacro %}
+
+{{ macros.header() }}
```
cmccandless:
```suggestion
{{ macros.header(imports=["Rational"]) }}
```

exercises/rational-numbers/rational_numbers_test.py
```diff
@@ -96,43 +106,43 @@ def test_raise_zero_to_an_integer_power(self):
...
+    def test_raise_a_negative_rational_number_to_the_power_of_zero(self):
         self.assertEqual(Rational(-1, 2) ** 0, Rational(1, 1))
 
-    # Test exponentiation of a real number to a rational number
-    def test_raise_a_real_number_to_a_positive_rational(self):
-        self.assertAlmostEqual(8 ** Rational(4, 3), 16.0, places=8)
+    # Tests of type: Exponentiation of a real number to a rational number
+
+    def test_raise_a_real_number_to_a_positive_rational_number(self):
+        self.assertEqual(8 ** Rational(4, 3), 16.0)
```
cmccandless: The exponentiation tests should use `assertAlmostEqual` with `places=8`

exercises/rational-numbers/rational_numbers_test.py
```diff
@@ -1,93 +1,103 @@
-from __future__ import division
```
cmccandless: For now, we still need this import to pass tests in Python 2.

PR #2092, opened 2019-10-26
linked-list: add comment to understand meaning of the Node

kalyaniingle:


yawpitch:
Renaming this so we can easily see what exercise it relates to.

yawpitch:
@kalyaniingle is this not already clear from the description.md file?

> You will write an implementation of a doubly linked list. Implement a Node to hold a value and pointers to the next and previous nodes.

A comment in the slug file doesn't seem as helpful as the generated README.md.

exercises/linked-list/linked_list.py
```diff
@@ -1,3 +1,4 @@
+# Represents smallest unit of a list with value, reference to succeding Node and referenece previous Node
```
yawpitch: This might be more clear as `# A Node stores a value in the doubly linked list, as well as a link to any previous and and succeeding Nodes`.

yawpitch:

PR #2093, opened 2019-10-26
binary-search-tree: add comment about TreeNode class

gulabraoingle:


yawpitch:
Renaming to reflect what exercise this is related to.

yawpitch:
@gulabraoingle this is already explained in the description.md:

> A binary search tree consists of a series of connected nodes. Each node contains a piece of data (e.g. the number 3), a variable named left, and a variable named right. The left and right variables point at nil, or other nodes. Since these other nodes in turn have other nodes beneath them, we say that the left and right variables are pointing at subtrees. All data in the left subtree is less than or equal to the current node's data, and all data in the right subtree is greater than the current node's data.

Is that not enough of an explanation?

yawpitch:

PR #2097, opened 2019-10-28
generator: Add support for exercise plugins

cmccandless:
Proposal to allow template authors to specify additional Python code to support their templates.

*Reason:* some things are very, very verbose in Jinja. The example that prompted this was the `react` exercise. The [spec](https://github.com/exercism/problem-specifications/blob/master/exercises/react/canonical-data.json) for this exercise is quite complex. There was a need to find all unique callback names in each test case. In Python, this is quite easy:
```
callbacks = sorted({
    op["name"]
    for op in case["input"]["operations"]
    if op["type"] == "add_callback"
})
```

In Jinja (which does not support comprehensions), this is quite difficult. However, if template authors were able to have a `.meta/plugins.py`, they could include the above snippet (which is totally not re-usable for other exercises, and thus has no place in the generator itself).

### DIFF
--- a/bin/generate_tests.py
+++ b/bin/generate_tests.py
@@ -195,7 +195,7 @@ def generate_exercise(env, spec_path, exercise, check=False):
         else:
             shutil.move(tmp.name, tests_path)
             print(f"{slug} generated at {tests_path}")
-    except (TypeError, UndefinedError) as e:
+    except (TypeError, UndefinedError, SyntaxError) as e:
         logger.debug(str(e))
         logger.error(f"{slug}: generation failed")
         return False
@@ -205,10 +205,6 @@ def generate_exercise(env, spec_path, exercise, check=False):
     except FileNotFoundError as e:
         logger.debug(str(e))
         logger.info(f"{slug}: no canonical data found; skipping")
-    except SyntaxError as e:
-        logger.debug(str(e))
-        logger.info(f"{slug}: SyntaxError in {plugins_source}")
-        return False
     return True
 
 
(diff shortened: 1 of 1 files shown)


cmccandless:
@yawpitch Good suggestion on importlib. I knew there was a way to do it, but wasn't sure of the advantages of one over the other. Your suggestion works great.

> I'm not sure about the last bit, but it seems it might be better to have a jinja failure from a missing "plugins" key rather than fail on some usage of a key with a value of None.

I like this approach, and it does indeed give a failure if the file doesn't exist.

yawpitch:
Might also want to try and see what happens if the plugins.py is present but contains a SyntaxError or otherwise fails to import. Should that fail an entire, possibly multi-exercise generation run? Or should it log the issue, fail that one, and continue.

cmccandless:
I'll test for what it *currently* does, but I would say the expected behavior should be to log the failure and move on, just like template syntax errors do now. The end return code is still set to 1.

cmccandless:
New behavior:
```
$ bin/generate_tests.py -v react
DEBUG:generator:'return' outside function (plugins.py, line 5)
INFO:generator:react: SyntaxError in exercises/react/.meta/plugins.py
$ echo $?
1
```

cmccandless:
Alternatively, this could be grouped with handling TypeError and UndefinedError and just print "generation failed" unless "--verbose" is present.

yawpitch:
I'd say grouping them together would be fine.

bin/generate_tests.py
```diff
@@ -150,54 +150,70 @@ def generate_exercise(env, spec_path, exercise, check=False):
     False: saves rendered to tests file
     """
     slug = os.path.basename(exercise)
+    meta_dir = os.path.join(exercise, ".meta")
+    plugins = None
```
yawpitch: Line 154 is no longer needed.

PR #2107, opened 2019-11-01
two-fer: Add test for handling empty string param

nguyendat:
Ref: https://exercism.io/solutions/a1ec5f564aba409b8377f25ef1b80cee/comments/162304

### DIFF
--- a/exercises/two-fer/two_fer_test.py
+++ b/exercises/two-fer/two_fer_test.py
@@ -14,6 +14,9 @@ class TwoFerTest(unittest.TestCase):
 
     def test_another_name_given(self):
         self.assertEqual(two_fer("Bob"), "One for Bob, one for me.")
+        
+    def test_name_empty_given(self):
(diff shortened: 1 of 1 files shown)


yawpitch:
@nguyendat I'm afraid this test case doesn't really make sense, and neither does the discussion on that exercise: the purpose of a default value is to be used only _if no argument whatsoever_ is supplied for that parameter ... if _any_ string-convertible argument _is_ provided the argument should be used, even if it's nonsensical, otherwise you're second guessing your user's intent, and the user is allowed to have nonsensical intent.

So the "correct" response for `two_fer("")` really is "One for , one for me." ... it might not make the most _sense_, but it's more consistent than assuming that an explicitly passed empty string is the same thing as no string being passed at all. Similarly `two_fer(None)` really should return "One for None, one for me." ... if the user explicitly passes an argument we should tend to assume that was an informed choice, even if it's a silly one.

I'm going to close this as wontfix for that reason, but thank you for the contribution.

PR #2113, opened 2019-11-06
Drop Python2.7 support

cmccandless:


cmccandless:
@exercism/python The work for this has been done for a while. I cherry-picked the commits onto the current master branch to make the changes easily merge-able.

The intent of this project was to drop support in this track by the end of the calendar year. This is a soft-drop, meaning that we will no longer test against Python2.7, but Python2.7 *may* continue to work for those still insisting on using it for a time (until code is added that takes advantage of Python3-only features).

As we are approaching the end of the year, it seemed like a good time to get the ball rolling.

docs/ABOUT.md
```diff
@@ -10,6 +10,6 @@ The most common form of Python is compiled in C; this is often invisible to the
 
 Python was started by Guido van Rossum in 1989; its name is an homage to the comedy troupe Monty Python.
 
-Python 2 is used widely but support [may end by 2020](https://www.python.org/dev/peps/pep-0373/#id2); it is highly recommended that beginners use Python 3 (they are similar, but users will encounter [some differences](http://blog.teamtreehouse.com/python-2-vs-python-3)).
+This track uses Python 3. It is highly recommended that students upgrade to at least Python 3.6 as some features used by this track may not be supported in earlier version of Python.
```
yawpitch: This line seems to indicate that "3.4+" mentioned elsewhere might need tweaking?

cmccandless: *highly recommended*...

My thought is that we recommend 3.6, but we still *support* 3.4. It's not EOL yet, and 3.5 is nowhere near EOL either, but when 3.6 is the default installation for Python3 on many systems now (if the OS *has* a default installation), it doesn't seem correct to suggest installing anything older.

We might amend this note to mention each new release has performance improvements over the last.

On that note, with the recent release of 3.8, should we recommend 3.7 if Python3 is not already installed?

cmccandless:
@yawpitch
> Now we just need to obliterate the (object) from all the class-based exercises.

Done.

yawpitch:
You can lose the parentheses as well ... if the _only_ thing a class inherits from is `object`:

```python
class Foo:
    pass
```

docs/ABOUT.md
yawpitch: Generally speaking we should probably be recommending 3.7+ at the moment, 3.6+ if you're a bit out of date... nothing older. The simple fact is 3.6 has been around since before Trump assumed office. Which, admittedly, isn't _that_ long ago, but _feels_ like several decades.

yawpitch:
Should state that a little more emphatically ... in 3 you _should_ elide parentheses after the class name unless there's something to inherit from or you're assigning a metaclass.

cmccandless:
Parenthesis dropped.

cmccandless:
I'm going to notify the mentors on Slack that we are officially dropping Python2 support, and merge this tomorrow.

yawpitch:
Exciting times!

PR #2114, opened 2019-11-10
Go counting add test template

dvermd:
Resolves: #1951


yawpitch:
Much better with the macro, however it would be cleaner if you had it build a set literal for non-empty lists (ie `{(1, 0), (2, 0)}` rather than `set([(1, 0), (2, 0])`) and only used the set constructor for the empty list case.

Issue #1946, opened 2019-10-02
dominoes: add test template


cmccandless:
## Task
Create a test template file for `dominoes`

### Details
1. If this is your first time contributing to this repository, welcome!
    - Contributing to a repository on GitHub requires a [fork](https://help.github.com/en/articles/fork-a-repo)
1. Create a branch on your local repository called `dominoes-add-test-template`
    - Run: `git checkout -b dominoes-add-test-template`
1. Clone problem-specifications repository; this is used to populate the template
    - Run: `git clone https://github.com/exercism/problem-specifications ../problem-specifications`
1. Create `exercises/dominoes/.meta/template.j2`
    - See [Generator Documentation] for more information
1. Generate test file from new template
    - Run: `bin/generate_tests.py dominoes`

At this point, if the generator fails, or you are otherwise stuck,
please feel free to reach out! It is perfectly acceptable to submit
an incomplete PR in order to get feedback or assistance.

### Resources
- [Contributing Guide]
- [Generator Documentation]
- [Closing issues using keywords]

[Contributing Guide]: https://github.com/exercism/python/blob/master/docs/CONTRIBUTING.md
[Generator Documentation]: https://github.com/exercism/python/blob/master/docs/GENERATOR.md
[Closing issues using keywords]: https://help.github.com/en/articles/closing-issues-using-keywords

mbbroberg:
I ran this one to give it a try and I'm receiving an error after running `bin/generate_tests.py dominoes`:

```
error: cannot format /var/folders/5l/c1y1gcw97szdywgf9rk1100m0000gn/T/tmpwiaj8rye: Cannot parse: 9:16:         value =
Traceback (most recent call last):
  File "bin/generate_tests.py", line 257, in <module>
    generate(**opts.__dict__)
  File "bin/generate_tests.py", line 217, in generate
    if not generate_exercise(env, spec_path, exercise, check):
  File "bin/generate_tests.py", line 169, in generate_exercise
    format_file(tmp.name)
  File "bin/generate_tests.py", line 132, in format_file
    check_call(['black', '-q', path])
  File "/Users/mbbroberg/.pyenv/versions/3.7.3/lib/python3.7/subprocess.py", line 347, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['black', '-q', '/var/folders/5l/c1y1gcw97szdywgf9rk1100m0000gn/T/tmpwiaj8rye']' returned non-zero exit status 123.
```

[...]

PR #2116, opened 2019-11-10
dominoes: add test template

dvermd:
Resolves: #1946


exercises/dominoes/dominoes_test_sav.py
```diff
@@ -0,0 +1,114 @@
+import unittest
```
cmccandless: This file should not be checked in. Please remove it.

yawpitch:
> @yawpitch, any odd side effects to changing the domino format from tuples to lists?

Not that I can see in the current code, though the **swap** method in the example.py creates tuples, so there’s a bit of a mismatch in the internal representation if the inputs are lists... the normalize method on the test suite is set up to effectively ignore tuple vs list, so the only thing the tuples were really doing is inhibiting the urge to swap by mutating their given domino in place.

Shouldn’t take much more than a template patch to get it back to tuples, which arguably are the more appropriate data type.

PR #2129, opened 2019-11-15
connect: add test template

dvermd:
My IDE suggests some changes to example.py that seems OK to me. I included them as a separate commit.
Are these changes OK for exercism ? Do you want me to create another PR for them ?

Resolves: #1938

### DIFF
--- a/exercises/connect/.meta/template.j2
+++ b/exercises/connect/.meta/template.j2
@@ -5,8 +5,6 @@
     def test_{{ case["description"] | to_snake }}(self):
         game = ConnectGame("""{{ case["input"]["board"] | join("\n               ") }}""")
         winner = game.get_winner()
-        self.assertEqual(winner, "{{ case["expected"] }}",
-                         "Test failed: \"%s\", expected winner: %s, got: %s"
-                             % ("{{ case["description"] }}", "{{ case["expected"] | default("None", true)}}", winner if winner else "None"))
+        self.assertEqual(winner, "{{ case["expected"] }}")
 
 {%- endmacro %}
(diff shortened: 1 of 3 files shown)


exercises/connect/.meta/template.j2
```diff
@@ -0,0 +1,12 @@
+{%- extends "master_template.j2" -%}
+{% set imports = ["ConnectGame"] %}
+
+{% macro test_case(case) -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        game = ConnectGame("""{{ case["input"]["board"] | join("\n               ") }}""")
+        winner = game.get_winner()
+        self.assertEqual(winner, "{{ case["expected"] }}",
+                         "Test failed: \"%s\", expected winner: %s, got: %s"
+                             % ("{{ case["description"] }}", "{{ case["expected"] | default("None", true)}}", winner if winner else "None"))
```
cmccandless: The custom message is no longer needed, since every test is its own method now.

exercises/connect/.meta/template.j2
```diff
@@ -0,0 +1,10 @@
+{%- extends "master_template.j2" -%}
+{% set imports = ["ConnectGame"] %}
+
+{% macro test_case(case) -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        game = ConnectGame("""{{ case["input"]["board"] | join("\n               ") }}""")
```
cmccandless: Unless @yawpitch knows a more elegant way to generate these boards, LGTM.

yawpitch: Nothing I can think of off the top of my head that will do it better.

PR #2127, opened 2019-11-14
Add test template for ocr-numbers

tqa236:
Resolve #1963 . #2051 seems not to be active so I finished it for them.

PR #2130, opened 2019-11-18
ocr-numbers: regenerate README

cmccandless:
#2127 should not have passed the CI check `bin/check-readmes.sh`. There is a bug in that script.


cmccandless:
#2132 supersedes this PR

PR #2134, opened 2019-11-20
generate_tests: allow newer Python versions

yawpitch:
Current version forces 3.7, but one of the few pains of being on Arch Linux is that _suddenly_ you're on 3.8, and you can't run the script anymore.

Sigh.

### DIFF
--- a/bin/generate_tests.py
+++ b/bin/generate_tests.py
@@ -15,8 +15,8 @@ Usage:
 import sys
 
 _py = sys.version_info
-if _py.major < 3 or (_py.major == 3 and _py.minor < 7):
-    print("Python version must be at least 3.7")
+if _py.major < 3 or (_py.major == 3 and _py.minor < 6):
+    print("Python version must be at least 3.6")
     sys.exit(1)
 
 import argparse
--- a/bin/generate_tests.py
+++ b/bin/generate_tests.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3.7
+#!/usr/bin/env python3
 """
 Generates exercise test suites using an exercise's canonical-data.json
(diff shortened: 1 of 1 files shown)


bin/generate_tests.py
```diff
@@ -12,6 +12,13 @@
     generate_tests.py --check           Checks if test files are out of sync with templates
     generate_tests.py --check two-fer   Checks if two-fer test file is out of sync with template
 """
+import sys
+
+_py = sys.version_info
+if _py.major < 3 or (_py.major == 3 and _py.minor < 7):
+    print("Python version must be at least 3.7")
```
cmccandless: Just tested this: the generator runs fine on 3.6. 3.5 doesn't support `black==19.3b0`.

No reason we can't allow 3.6 to run the generator until we employ [walruses](https://www.python.org/dev/peps/pep-0572/) or something else new in 3.7.

yawpitch: True ... though we had a soft requirement on 3.7 simply because of what was in the shebang ... do we _want_ to relax that?

cmccandless: :man_shrugging: I'll be honest here: that shebang was the result of my laziness and not wanting to type out `python3.7 bin/generate_tests.py`. My logic at the time for developing the generator in 3.7 was that it was new and shiny; clearly all the features used are supported in 3.6.

Unless we can demonstrate a clear performance boost using 3.7 over 3.6, I don't have a problem relaxing the requirement.

yawpitch: Righto.

yawpitch: Pushed changed will allow 3.6 and up.

PR #2136, opened 2019-11-20
Dnd character add test template

tqa236:
Solve #1945 .

The last 3 tests are quite special, so I just pasted them there verbatim. I don't think they are likely to change in the future, so it's not worth the effort writing templates for them.

### DIFF
--- a/exercises/dnd-character/.meta/template.j2
+++ b/exercises/dnd-character/.meta/template.j2
@@ -1,5 +1,6 @@
 {%- import "generator_macros.j2" as macros with context -%}
-{{ macros.header() }}
+{% set class = exercise | camel_case -%}
+{{ macros.header(["Character", "modifier"]) }}
 
 class {{ exercise | camel_case }}Test(unittest.TestCase):
     {% for case in cases[0]["cases"] -%}
--- a/exercises/dnd-character/.meta/template.j2
+++ b/exercises/dnd-character/.meta/template.j2
@@ -3,31 +3,39 @@
 {{ macros.header(["Character", "modifier"]) }}
 
 class {{ exercise | camel_case }}Test(unittest.TestCase):
-    {% for case in cases[0]["cases"] -%}
+    {% for supercase in cases -%}
+    {% set property = supercase["property"] -%}
+    {% if "cases" in supercase -%}
+    {% for case in supercase["cases"] -%}
     def test_{{ case["description"] | replace("-","n") | to_snake }}(self):
         {% set value = case["input"]["score"] -%}
         {% set expected = case["expected"] -%}
         self.assertEqual({{ case["property"] }}({{ value }}), {{ expected }})
-
     {% endfor %}
 
+    {% elif property == "ability" -%}
     def test_random_ability_is_within_range(self):
-        score = Character().{{ cases[1]["property"] }}()
-        self.assertTrue({{ cases[1]["expected"] | replace("&&","and") }})
+        score = Character().{{ property }}()
+        self.assertIs({{ cases[1]["expected"] | replace("&&","and") }}, True)
 
+    {% elif property == "character" -%}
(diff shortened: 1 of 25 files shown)


cmccandless:
@yawpitch, thoughts on the hard-coded tests?

cmccandless:
@yawpitch, thoughts on the hard-coded tests?

yawpitch:
> @yawpitch, thoughts on the hard-coded tests?

I mean it works for now, but I’d much rather see it take all the names and values from the canonical data, because the whole point of the generation step and the templates is to automatically react to upstream changes in the data.

tqa236:
I think I have a workaround for that. Please wait a little.

tqa236:
I have this error when merging the latest commit about `generate_tests.py`, any idea why?

```bash
  File "bin/generate_tests.py", line 39, in <module>
    from jinja2 import Environment, FileSystemLoader, TemplateNotFound, UndefinedError
ModuleNotFoundError: No module named 'jinja2'

```

tqa236:
What do you think about this one?

exercises/dnd-character/.meta/template.j2
```diff
@@ -0,0 +1,33 @@
...
+    {% for case in cases[0]["cases"] -%}
+    def test_{{ case["description"] | replace("-","n") | to_snake }}(self):
+        {% set value = case["input"]["score"] -%}
+        {% set expected = case["expected"] -%}
+        self.assertEqual({{ case["property"] }}({{ value }}), {{ expected }})
+
+    {% endfor %}
+
+    def test_random_ability_is_within_range(self):
+        score = Character().{{ cases[1]["property"] }}()
```
cmccandless: What if another case is inserted before this one?

tqa236: Likely we'll get a bug and then we can fix it. I mean it's not easy to anticipate all the changes of the test file to cover them all in the template. There's many way we can ask "What if" already: What if the depth is 3 or higher, what if a new test case that raises an error is added to the list of normal test cases,...

Btw, what change do you suggest for this one?

exercises/dnd-character/.meta/template.j2
```diff
@@ -0,0 +1,33 @@
...
+        {% if ability == "hitpoints" -%}
+        {% set statement = statement | replace("constitution","Char.constitution") -%}
+        {%- endif -%}
+        self.assertTrue({{ statement }})
+        {% endfor %}
+
+    def test_each_ability_is_only_calculated_once(self):
+        Char = Character()
+        {% set ability = cases[3]["property"] -%}
+        self.assertTrue({{ cases[3]["expected"] | replace(ability , ["Char.", ability]|join(""))}})
```
cmccandless: To avoid false-positives on [truthy](https://stackoverflow.com/q/39983695/3229611) values, please use `assertIs(_, True)` instead of `assertTrue` and `assertIs(_, False)` instead of `assertFalse`.

```suggestion
        self.assertIs({{ cases[3]["expected"] | replace(ability , ["Char.", ability]|join(""))}}, True)
```

exercises/dnd-character/.meta/template.j2
cmccandless: I'm not sure at the moment... give me some time to think on it.

tqa236:
@cmccandless I added new commits to address your two point

- `assertTrue` -> `assertIs`, it's an easy part.
- About the hardcoded order of tests used in the template before, now I change to use test's property instead. At least now there's no error if we have new test cases in between. However, this solution still has one disadvantage, if there's a new test case with a new property, we can't detect it.

What do you think about it?

cmccandless:
> if there's a new test case with a new property, we can't detect it.

This is a weakness of several exercise templates here, not just this one, and is an acceptable risk.

PR #2142, opened 2019-11-22
Add Python 3.8 to Travis CI config

tqa236:
I added Python 3.8 ([released Oct 14, 2019](https://docs.python.org/3/whatsnew/3.8.html)) to Travis CI config and change `assertEquals` to `assertEqual` in POV exercise to fix the deprecation warning.

```
#  pov

/tmp/tmppvu6epthpov/pov_test.py:131: DeprecationWarning: Please use assertEqual instead.

  self.assertEquals(tree.path_to("a", "c"), expected)
```

As far as I understand, they are just alias now: https://stackoverflow.com/questions/930995/assertequals-vs-assertequal-in-python/931011

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 language: python
 dist: xenial
-  
+
 python:
   - 3.5
   - 3.6
   - 3.7
+  - 3.8
   - nightly
 
 matrix:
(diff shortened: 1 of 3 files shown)


cmccandless:
Maintainer's note: the 3.8 CI job ran longer than the 3.7 because Travis CI currently downloads/installs 3.8 at runtime, while older versions of Python are pre-installed on the image.

PR #2150, opened 2019-12-22
various: comma fixes

yawpitch:
Accidentally running test generation in a future version of Black (19.10b0) revealed that we've got several templates that were generating unnecessary trailing commas. These are elided by Black in the version in requirements-generator.txt, but apparently won't be in the future.

**NOTE**: this should be merged _after_ #2149 ... the changes from that got incorporated by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ bin/configlet.exe
 .cache
 .pytest_cache
 __pycache__
+.venv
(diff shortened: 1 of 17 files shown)


cmccandless:
Should we update requirements-generator.txt to use the newer version of `black` to avoid over-complicating templates?

yawpitch:
Bumping will cause a lot of churn, because it’s going to make most “from” imports multiline. Plus we’ll still have to add the handling for last commas... we just should have been doing that all along.

PR #2156, opened 2020-01-11
say: add test template 

tsamoglou:
I have a problem when I execute generate_test.py  
`error: cannot format /tmp/tmp3247y8ob: Cannot parse: 43:6: , one hundred)
  File "bin/generate_tests.py", line 331, in <module>
    generate(**opts.__dict__)
  File "bin/generate_tests.py", line 297, in generate
    if not generate_exercise(env, spec_path, exercise, check):
  File "bin/generate_tests.py", line 214, in generate_exercise
    format_file(tmp.name)
  File "bin/generate_tests.py", line 176, in format_file
    check_call(["black", "-q", path])
  File "/usr/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['black', '-q', '/tmp/tmp3247y8ob']' returned non-zero exit status 123.
 `

### DIFF
--- a/exercises/say/.meta/template.j2
+++ b/exercises/say/.meta/template.j2
@@ -1,0 +1,20 @@
+{%- import "generator_macros.j2" as macros with context -%}
+
+{%- macro test_call(case) %}
+            {{ case["property"] }}(
+                {{ case["input"]["number"] }}
+            )
+{% endmacro -%}
+
+{{ macros.header() }}
+
+class {{ exercise | camel_case }}Test(unittest.TestCase):
+    {% for case in cases -%}
+    def test_{{ case["description"] | to_snake }}(self):
+        {%- if case is error_case %}
+        with self.assertRaisesWithMessage(ValueError):
+            {{- test_call(case) }}
+        {%- else %}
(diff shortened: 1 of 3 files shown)


yawpitch:
That error indicates some sort of invalid syntax in the temporary file, which usually means there is something wrong in the template.

In this case you can look at the generated /tmp/tmp3247y8ob file and see what’s wrong with it. I find it helpful to run black on it manually to get a better error message when this happens.

M

tsamoglou:
> Seems to me the error_case logic isn’t being tripped, since the utility function isn’t being added but self.assertRaisesWithMessage is used twice, so it should be there.
> 
> Probably that `is` that’s causing the trouble.

I fixed the problem, I dint have "" in strings cases

yawpitch:
The issue is the ‘footer` macro needs to be called at the end of the template (to add the `if __name__` guard, and the template should in turn call the `utility` macro if there is an error case.

tsamoglou:
Ready to merged?

PR #2160, opened 2020-01-15
[WIP] Zipper add test template

tsamoglou:
`{%- if tree_obj is null-%}
null
` 
jinja2.exceptions.TemplateAssertionError: no test named 'null'

what can I do to see if tree_object == null????


cmccandless:
The correct syntax is:
`{%- if tree_object is none -%}`

tsamoglou:
> The correct syntax is:
> `{%- if tree_object is none -%}`

Yeah but in canonical data of zipper in atribbute left the value is null not none 
@cmccandless

cmccandless:
When Python loads JSON objects, null becomes `None`

tsamoglou:
Thank you for your responding I will make the changes @cmccandless

PR #2169, opened 2020-01-30
word-count: Add multiple apostrophes test case

c4llmeco4ch:
This is a localized version of a change I discussed with the main problem-specs repo. They told me that, due to their lockdown, I should bring this to the language-specific repositories.

Effectively, the problem states that contractions ("can't", "won't", etc.) should treat apostrophes as part of the word, but rule 3 states:

Other than the apostrophe in a contraction all forms of punctuation are ignored
As such, I've added a very simple case testing that normal apostrophes are removed. The following code (while bad) should do a good job of showcasing a solution that passes all the live tests but fails the new one:

```
from string import punctuation
def count_words(sentence): 
    dict = {}
    punc = punctuation.replace("'",'')
    for val in punc:
        if val in sentence:
            sentence = sentence.replace(val, ' ')
    for word in sentence.split():
        if (w := word.lower())[0] == "'":
            w = w[1:]
        if w[-1] == "'":
            w = w[:-1]
        if w not in dict:
            dict[w] = 0
        dict[w] += 1
    return dict
```

Thanks!

### DIFF
--- a/exercises/word-count/.meta/additional_tests.json
+++ b/exercises/word-count/.meta/additional_tests.json
@@ -35,7 +35,7 @@
       }
     },
     {
-      "description": "multiple_apostrophes_ignored",
+      "description": "multiple apostrophes ignored",
       "property": "countWords",
       "input": {
         "sentence": "''hey''"
--- a/exercises/word-count/.meta/additional_tests.json
+++ b/exercises/word-count/.meta/additional_tests.json
@@ -33,6 +33,16 @@
         "is": 1,
         "broken": 1
       }
+    },
+    {
+      "description": "multiple_apostrophes_ignored",
+      "property": "countWords",
+      "input": {
+        "sentence": "''hey''"
+      },
+      "expected": {
+        "hey": 1
+      }
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


cmccandless:
Ref: exercism/problem-specifications#1628

yawpitch:
Slight divergence, but what the heck should happen with a plural possessive? For instance, what should the word count be for the phrase "lawyers accept both lawyer's fees and lawyers' fees?"?

c4llmeco4ch:
Sounds good, I'll get on that later today. I appreciate the help from everyone

c4llmeco4ch:
Really sorry for possibly blowing up everyone's feed. After generating tests from the automated script, I'm still getting failures from Travis. Do I need to revert the test script back to the original passing version then push without the test added? Thanks and sorry again for the mess.

c4llmeco4ch:
So I double checked and the reason I was pushing two versions of the test was due to my local version of problem-specs contained the test from my discussion with the main repo. I updated that but I'm still having the same failures from Travis.

yawpitch:
Did you actually set the branch in your local copy of problem-specifications to `master`, rather than just remove the test in your working branch? The version number shouldn't be bumped in the generated tests, since the canonical-data.json it's pulling from hasn't (in `master`) been versioned.

c4llmeco4ch:
I'll take a look Monday. That could be the issue in which case I'll revert both files and start from scratch to see if that'll make it work.

c4llmeco4ch:
Good call @yawpitch

PR #2180, opened 2020-02-20
Fetch configlet python

cmccandless:


.travis.yml
```diff
@@ -16,6 +16,7 @@ matrix:
     - env: JOB=HOUSEKEEPING
       python: 3.7
       install:
+        - pip install requests
         - ./bin/fetch-configlet
         - git clone https://github.com/exercism/problem-specifications spec
         - pip install -r requirements-generator.txt
```
yawpitch: Maybe add requests to the requirements-generator.txt and bump it above the fetch-configlet call?

cmccandless: That was my first idea, but as it isn't technically a requirement for the generator, I was hesitant to add it there. Likewise, creating yet another requirements file for a single package seemed wrong too.

bin/fetch-configlet
```diff
@@ -1,32 +1,47 @@
...
+def parse_arch(arch):
+    if '64' in arch:
+        return '64bit'
+    elif '686' in arch or '386' in arch:
+        return '32bit'
+    return '64bit'
+
+
+def fetch_configlet():
+    os, *_, arch = platform.uname()
```
yawpitch: That could be a bit problematic ... `platform.uname()` returns a namedtuple, and on my machine for instance the last item is actually an empty string (for processor). The `platform.uname().machine` is more reliable, although that's aliased to `platform.machine()` ... but there's a bit near the top of the [docs](https://docs.python.org/3.8/library/platform.html#platform.architecture) that suggests the best approach is `is_64bits = sys.maxsize > 2**32`.

yawpitch: Shadows the `os` builtin ... `platform.platform()` will give you this directly ... BTW have we abandoned triing to work on Mac, cause `os` here would be "Darwin".

bin/fetch-configlet
```diff
@@ -1,32 +1,47 @@
...
+        print("{name}".format(**asset))
+        if asset["name"] == name:
+            print(f"Downloading configlet {version} for {os}-{arch}")
+            download_and_extract(asset["browser_download_url"])
+            return 0
+    return 1
+
+
+if __name__ == "__main__":
+    max_retries = 5
```
yawpitch: Maybe everything under the `__name__` guard to a `def main`?

.travis.yml
yawpitch: True ... not a big fan of having it there all alone, but I suppose it's acceptable.

bin/fetch-configlet
```diff
@@ -1,33 +1,54 @@
...
+
+
+def download_and_extract(url):
+    resp = requests.get(url)
+    print('extracting...')
+    file = tarfile.open(fileobj=io.BytesIO(resp.content), mode='r:gz')
+    file.extractall(path='bin/')
+
+
+def parse_arch(arch):
```
yawpitch: `parse_arch` no longer required.

bin/fetch-configlet
```diff
@@ -1,33 +1,54 @@
...
+    return '64bit'
+
+
+def is_64bites():
+    return sys.maxsize > 2**32
+
+
+def fetch_configlet():
+    os = platform.uname().system.lower()
+    if is_64bites():
```
yawpitch: This could be a ternary `if`.

bin/fetch-configlet
```diff
@@ -1,33 +1,54 @@
...
+
+def parse_arch(arch):
+    if '64' in arch:
+        return '64bit'
+    elif '686' in arch or '386' in arch:
+        return '32bit'
+    return '64bit'
+
+
+def is_64bites():
```
yawpitch: I don't think it takes that many "bites".

bin/fetch-configlet
cmccandless: Yes...

bin/fetch-configlet
```diff
@@ -1,33 +1,55 @@
...
+
+def download_and_extract(url):
+    resp = requests.get(url)
+    print('extracting...')
+    file = tarfile.open(fileobj=io.BytesIO(resp.content), mode='r:gz')
+    file.extractall(path='bin/')
+
+
+def get_os():
+    _os = platform.platform()
```
yawpitch: Not a blocker, but the convention is `name_` to avoid shadowing, `_name` to imply "non-public".

bin/fetch-configlet
```diff
@@ -1,33 +1,55 @@
...
+    if _os.startswith("Darwin"):
+        return "mac"
+    elif _os.startswith("Windows"):
+        return "windows"
+    else:
```
yawpitch: All branches above `return` so `else` is not required.

bin/fetch-configlet
```diff
@@ -1,33 +1,55 @@
...
+def download_and_extract(url):
+    resp = requests.get(url)
+    print('extracting...')
+    file = tarfile.open(fileobj=io.BytesIO(resp.content), mode='r:gz')
+    file.extractall(path='bin/')
+
+
+def get_os():
+    _os = platform.platform()
+    if _os.startswith("Darwin"):
```
yawpitch: IIRC if you look at the source of `platform` the string will just be `Darwin` or `Windows`, so don't really need to check startswith.

cmccandless: I wasn't sure how closely it matched the output of `uname` on bash.

PR #2181, opened 2020-02-25
[sgf-parsing] fix escaped characters

### LINKED ISSUE #2176: [sgf-parsing] fix escaped characters test
Running the tests for `sgf-parsing` throws a DeprecationWarning related to an invalid escape sequence. This is due to the `\]` inclusions in the test string, which really should be `\\]` to work correctly.

Now from the canonical data we are getting `'(;A[\\]b\\nc\\nd\\t\\te \\n\\]])'`, but by the time it's rendered into the _test.py it's become `"(;A[\]b\nc\nd\t\te \n\]])"`, leading to the aforementioned warning.

To fix this we'd seemingly need to flag the literal as raw _before_ parsing the json file, or we'd need to manually correct for improperly escaped non-control characters.

Unfortunately we cannot just render the strings with "r" in front of them in the Jinja template, as that ends up turning `\n` into two characters, `\\` and `n`.

Open to ideas.

yawpitch:
Fixes issue with pytest throwing a warning on invalid escapes.

Closes #2176

### DIFF
--- a/bin/generate_tests.py
+++ b/bin/generate_tests.py
@@ -101,6 +101,47 @@ def parse_datetime(string, strip_module=False):
         return result.replace("datetime.", "", 1)
     return result
 
+INVALID_ESCAPE_RE = re.compile(
+    r"""
+    \\(?!                           # a backslash NOT followed by
+        newline                     # the literal newline
+    |[                              # OR precisely one of
+        \\                          # another backslash
+        '                           # the single quote
+        "                           # the double quote
+        a                           # the ASCII bell
+        b                           # the ASCII backspace
+        f                           # the ASCII formfeed
+        n                           # the ASCII linefeed
+        r                           # the ASCII carriage return
+        t                           # the ASCII horizontal tab
+        v                           # the ASCII vertical tab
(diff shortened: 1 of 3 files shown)


bin/generate_tests.py
```diff
@@ -101,6 +101,47 @@ def parse_datetime(string, strip_module=False):
...
+    |                               # OR
+        N                           # a unicode char name composed of
+        \{                          # an opening brace
+            [A-Z][A-Z\ \-]*[A-Z]    # uppercase WORD, WORDs (or WORD-WORDs)
+        \}                          # and a closing brace
+    |                               # OR
+        u(?:[0-9A-Fa-f]{4})         # a 16-bit unicode char
+    |                               # OR
+        U(?:[0-9A-Fa-f]{8})         # a 32-bit unicode char
+    )""", flags=re.VERBOSE)
```
cmccandless: I had no idea Python `re` allowed you to expand out a regex pattern like this with line comments and everything.

yawpitch: Yep ... good ol `re.VERBOSE`.

PR #2184, opened 2020-03-01
[grep] Use mock for unit testing

### LINKED ISSUE #2147: grep: Use unittest.mock for testing
unittest.mock was not used originally because it was not available in Python 2.7. Since we've dropped support for Python 2.7, we can use the correct module for this now.

Work on this should be postponed until #2106 has been merged.

yawpitch:
Updates `grep` to use mock, per the work of @AnAccountForReportingBugs, with minor modifications (fallback after KeyError wouldn't work as is, and we want it to error loudly anyway).

Closes #2147.

### DIFF
--- a/exercises/grep/grep_test.py
+++ b/exercises/grep/grep_test.py
@@ -43,7 +43,7 @@ def open_mock(fname, *args, **kwargs):
         )
 
 
-@mock.patch("grep.open", name="open", side_effect=open_mock)
+@mock.patch("grep.open", name="open", side_effect=open_mock, create=True)
 @mock.patch("io.StringIO", name="StringIO", wraps=io.StringIO)
 class GrepTest(unittest.TestCase):
     # Test grepping a single file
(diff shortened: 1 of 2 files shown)


exercises/grep/grep_test.py
```diff
@@ -14,152 +15,105 @@
...
+def open_mock(fname, *args, **kwargs):
+    try:
+        return io.StringIO(FILE_TEXT[fname])
+    except KeyError:
+        raise RuntimeError(
+            "Expected one of {0!r}: got {1!r}".format(list(FILE_TEXT.keys()), fname)
+        )
 
 
+@mock.patch("grep.open", name="open", side_effect=open_mock)
```
AnAccountForReportingBugs: Two choices: `@mock.patch("builtins.open", name="open", side_effect=open_mock)` or `@mock.patch("grep.open", name="open", side_effect=open_mock, create=True)`
for it to work everywhere. I forgot that 3.4 doesn't add `create=True` by default for built-in functions the way 3.5 and later do. Note: using the first option will break `open` everywhere for the duration of the test, but that doesn't really matter here as you changed the fallback.

yawpitch: Thanks for that, went with the latter. CI now passing.

PR #2186, opened 2020-03-03
fetch-configlet: Honor rate limits and errors.

### LINKED ISSUE #2182: [bin/fetch-configlet] should backoff gracefully
We really should be inspecting the header and checking if the machine has been rate limited ... if it is we probably shouldn't simply repeatedly hammer on the same dead end.

AnAccountForReportingBugs:
Closes #2182.  Also removed dependency on requests module.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ matrix:
     - env: JOB=HOUSEKEEPING
       python: 3.7
       install:
-        - pip install requests
         - ./bin/fetch-configlet
         - git clone https://github.com/exercism/problem-specifications spec
         - pip install -r requirements-generator.txt
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -18,12 +18,12 @@ def download_and_extract(url):
         if status >= 400:
             print(f"Unexpected {status} status from download server.")
             time.sleep(BACKOFF_TIME)
-            return 1
+            return False
         response = request.read()
     print("extracting...")
     with tarfile.open(fileobj=io.BytesIO(response), mode="r:*") as file:
         file.extractall(path="bin/")
-    return 0
+    return True
 
 
 def get_os():
@@ -69,12 +69,14 @@ def fetch_configlet():
     for asset in data["assets"]:
         if asset["name"] == name:
             print(f"Downloading configlet {version} for {machine_info}")
-            return download_and_extract(asset["browser_download_url"])
+            for _ in range(MAX_RETRIES):
+                if download_and_extract(asset["browser_download_url"]):
+                    return 0
     return 1
 
 
 def main():
-    for i in range(MAX_RETRIES):
+    for _ in range(MAX_RETRIES):
         ret = fetch_configlet()
         if ret == 0:
             break
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -9,17 +9,20 @@ import time
 from urllib.request import urlopen
 
 
+BACKOFF_TIME = 5
+MAX_RETRIES = 5
+
 def download_and_extract(url):
     with urlopen(url) as request:
         status = request.status
         if status >= 400:
             print(f"Unexpected {status} status from download server.")
-            time.sleep(5)
+            time.sleep(BACKOFF_TIME)
             return 1
         response = request.read()
     print("extracting...")
-    file = tarfile.open(fileobj=io.BytesIO(response), mode="r:gz")
-    file.extractall(path="bin/")
+    with tarfile.open(fileobj=io.BytesIO(response), mode="r:*") as file:
+        file.extractall(path="bin/")
     return 0
 
 
@@ -49,17 +52,17 @@ def fetch_configlet():
         return 1
     if status == 403:
         wait_until = datetime.fromtimestamp(
-            headers["X-RateLimit-Reset"], timezone.utc
+            int(headers["X-RateLimit-Reset"]), timezone.utc
         )
         delta = wait_until - datetime.now(timezone.utc)
         seconds = delta.total_seconds()
-        wait = seconds + 5 if seconds > 5 else 5
+        wait = seconds + 5 if seconds > BACKOFF_TIME else BACKOFF_TIME
         print(f"Rate limited, sleeping {wait} seconds.")
         time.sleep(wait)
         return 1
     if status >= 400:
         sys.exit(f"Received unexpected {status} response from API.")
-    data = json.loads(response.decode())
+    data = json.loads(response.decode("utf-8"))
     version = data["tag_name"]
     machine_info = f"{get_os()}-{get_arch()}"
     name = f"configlet-{machine_info}.tgz"
@@ -71,8 +74,7 @@ def fetch_configlet():
 
 
 def main():
-    max_retries = 5
-    for i in range(max_retries):
+    for i in range(MAX_RETRIES):
         ret = fetch_configlet()
         if ret == 0:
             break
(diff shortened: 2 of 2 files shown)


bin/fetch-configlet
```diff
@@ -1,16 +1,26 @@
...
-    file.extractall(path='bin/')
+    with urlopen(url) as request:
+        status = request.status
+        if status >= 400:
+            print(f"Unexpected {status} status from download server.")
+            time.sleep(5)
+            return 1
+        response = request.read()
+    print("extracting...")
+    file = tarfile.open(fileobj=io.BytesIO(response), mode="r:gz")
```
yawpitch: Shadows `file` builtin; I know this late in the function that's a minimal issue, but best practice is still to avoid shadowing.

bin/fetch-configlet
```diff
@@ -23,21 +33,40 @@ def get_os():
...
+    data = json.loads(response.decode())
     version = data["tag_name"]
     machine_info = f"{get_os()}-{get_arch()}"
     name = f"configlet-{machine_info}.tgz"
     for asset in data["assets"]:
         if asset["name"] == name:
             print(f"Downloading configlet {version} for {machine_info}")
-            download_and_extract(asset["browser_download_url"])
-            return 0
+            return download_and_extract(asset["browser_download_url"])
```
yawpitch: For the sake of argument let's assume there's a temporary hiccup server side that will cause this to return 1 on the first attempt to download one of the assets ... this hiccup _will_ be fixed by a 5 second wait; _should_ we repeat the effort of fetching **data** and finding the appropriate asset again?

Basically shouldn't **download_and_extract** have its own backoff loop?

bin/fetch-configlet
yawpitch: Not that it's very impactful, but do we need to specify the compression mode? I believe `mode="r"` or `mode=r:*` will handle gzip compression transparently, while specifying `mode="r:gz" will fail were Github to ever decide to switch to one of the other **tarfile** supported compression schemes.

bin/fetch-configlet
```diff
@@ -1,16 +1,26 @@
...
 def download_and_extract(url):
-    resp = requests.get(url)
-    print('extracting...')
-    file = tarfile.open(fileobj=io.BytesIO(resp.content), mode='r:gz')
-    file.extractall(path='bin/')
+    with urlopen(url) as request:
+        status = request.status
+        if status >= 400:
+            print(f"Unexpected {status} status from download server.")
+            time.sleep(5)
```
yawpitch: This magic number 5 everywhere is going to be a pain if we ever decide to shorten or lengthen the backoff globally. Maybe a global `BACKOFF_TIME`?

bin/fetch-configlet
yawpitch: Maybe should use use `with tarfile.open`, to ensure the underlying buffer gets closed?

bin/fetch-configlet
```diff
@@ -23,21 +33,40 @@ def get_os():
...
+        )
+        delta = wait_until - datetime.now(timezone.utc)
+        seconds = delta.total_seconds()
+        wait = seconds + 5 if seconds > 5 else 5
+        print(f"Rate limited, sleeping {wait} seconds.")
+        time.sleep(wait)
+        return 1
+    if status >= 400:
+        sys.exit(f"Received unexpected {status} response from API.")
+    data = json.loads(response.decode())
```
yawpitch: Just to avoid ambiguity maybe specify `response.decode("utf-8")`.

bin/fetch-configlet
AnAccountForReportingBugs: I am unfamiliar with the `file` built in you are refrencing? Perhaps you are thinking of `open` or the `__file__` property?

AnAccountForReportingBugs:
I'll go over your notes and edit the travis file. I originally left it alone in case `requests` was used somewhere else.

bin/fetch-configlet
AnAccountForReportingBugs: Good idea.

bin/fetch-configlet
AnAccountForReportingBugs: I'll change that from the original method.

bin/fetch-configlet
AnAccountForReportingBugs: My thought was that we can't guarantee what the API will do in the future and race conditions. They might switch to temporary URLs, or the latest version change out from under us, etc.

bin/fetch-configlet
yawpitch: Ahh, sorry, that’s an old habit from the days when we supported Python 2 dying slower than it should have... not a builtin any longer. Disregard.

bin/fetch-configlet
yawpitch: I should think that if they’ve made an API change such that between parsing the configlet version and finding the platform-specific link for that version they’ve changed the endpoint of the URLs they’ve just served us then we’ve got bigger problems that aren’t going to be fixed by parsing the version number again.

I can see the race condition point, but if the latest version was swapped out form under us via a normal process it would leave the old download URLs intact, so your script would deliver stale data anyway... if the link has been permanently invalidated then retrying the download on its own loop will eventually kick back to
the outer loop and pick up the new version anyways, it just won’t needlessly hammer the latest endpoint — the one we already have rate limit issues with — in the more likely case, when there’s some temporary issue that’s blocking a valid URL from working.

bin/fetch-configlet
```diff
@@ -23,27 +36,45 @@ def get_os():
...
-            download_and_extract(asset["browser_download_url"])
-            return 0
+            return download_and_extract(asset["browser_download_url"])
     return 1
 
 
 def main():
-    max_retries = 5
-    for i in range(max_retries):
+    for i in range(MAX_RETRIES):
```
yawpitch: If **i** is never actually used in the loop, maybe use `_`?

AnAccountForReportingBugs: It was like that originally, but I'll change it for style regardless.

bin/fetch-configlet
AnAccountForReportingBugs: I'll just change it to use MAX_RETRIES in the method then.

.travis.yml
```diff
@@ -16,7 +16,6 @@ matrix:
     - env: JOB=HOUSEKEEPING
       python: 3.7
       install:
-        - pip install requests
```
yawpitch: Removing **requests** here is fine, it was added recently specifically for the previous version of the script.

bin/fetch-configlet
```diff
@@ -23,27 +36,47 @@ def get_os():
...
 def get_arch():
-    return '64bit' if sys.maxsize > 2**32 else '32bit'
+    return "64bit" if sys.maxsize > 2 ** 32 else "32bit"
 
 
 def fetch_configlet():
-    latest = "https://api.github.com/repos/exercism/configlet/releases/latest"
-    resp = requests.get(latest)
-    data = resp.json()
+    with urlopen(
```
yawpitch: Ok, we need to handle urllib.error.HTTPError being thrown here, probably via:

```python
from urllib.errors import HTTPError

[...]

try:
    with urlopen("https://api.github.com/repos/exercism/configlet/releases/latest") as request:
        status = request.status
        headers = dict(request.getheaders())
        response = request.read()
except HTTPError as err:
     status = err.code
     headers = dict(err.headers)
     response = err.reason
```

As right now it looks like a 403 is never actually reaching the `if status == 403` line.

bin/fetch-configlet
```diff
@@ -1,16 +1,29 @@
...
+import json
 import platform
-import requests
 import sys
+import tarfile
+import time
+from urllib.request import urlopen
+
 
+BACKOFF_TIME = 5
```
cmccandless:
```suggestion
BACKOFF_TIME = 5  # in seconds
```

PR #2185, opened 2020-03-03
Remove `paparomeo` from maintainers.

paparomeo:
Unfortunately, for most of the rest of this year, I'm not able to dedicate the minimum weekly time to help with the v3 drive and maintaining the track, so I'm removing myself as maintainer. Please revoke my GitHub privileges also.

PR #2187, opened 2020-03-04
Revert "Remove `paparomeo` from maintainers."

yawpitch:
On second thought @paparomeo what I'll do is mark you as an alumnus here (V2), but remove you from the V3 file. Does that work for you?

Reverts exercism/python#2185


yawpitch:
I've set your almunus status for now here, and opened https://github.com/exercism/v3/pull/866 for the V3 file. Let me know if you want any of the identifying details scrubbed from this one.

PR #2188, opened 2020-03-06
fetch-configlet: use logging module instead of print

cmccandless:


bin/fetch-configlet
```diff
@@ -10,9 +11,13 @@ from urllib.request import urlopen
 from urllib.error import HTTPError, URLError
 
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__file__)
```
yawpitch: This should be `__name__` not `__file__`.

bin/fetch-configlet
```diff
@@ -22,15 +27,15 @@ def download_and_extract(url):
         status = err.code
         response = err.reason
     except URLError as err:
-        print(err.reason)
+        logger.error(err.reason)
         time.sleep(BACKOFF_TIME)
         return False
     if status >= 400:
-        print(f"Unexpected {status} status from download server.")
+        logger.error(f"Unexpected {status} status from download server.")
```
yawpitch: Should we include the err.reason, aliased here as **response**?

bin/fetch-configlet
```diff
@@ -62,11 +67,11 @@ def fetch_configlet():
         headers = err.headers
         response = err.reason
     except URLError as err:
-        print(err.reason)
+        logger.error(err.reason)
```
yawpitch: For debug purposes might be good to do `logger.exception(err)` so we get some traceback here.

bin/fetch-configlet
```diff
@@ -62,11 +67,11 @@ def fetch_configlet():
...
         time.sleep(BACKOFF_TIME)
         return 1
     if status >= 500:
-        print(f"Sleeping due to {status} response from API.")
+        logger.info(f"Sleeping due to {status} response from API.")
```
yawpitch: We might also want the err.reason here.

bin/fetch-configlet
```diff
@@ -76,18 +81,19 @@ def fetch_configlet():
         delta = wait_until - datetime.now(timezone.utc)
         seconds = delta.total_seconds()
         wait = seconds + 5 if seconds > BACKOFF_TIME else BACKOFF_TIME
-        print(f"Rate limited, sleeping {wait} seconds.")
+        logger.info(f"Rate limited, sleeping {wait} seconds.")
         time.sleep(wait)
         return 1
     if status >= 400:
-        sys.exit(f"Received unexpected {status} response from API.")
+        logger.error(f"Received unexpected {status} response from API.")
```
yawpitch: Again with err.response.

bin/fetch-configlet
```diff
@@ -76,18 +81,19 @@ def fetch_configlet():
...
         seconds = delta.total_seconds()
         wait = seconds + 5 if seconds > BACKOFF_TIME else BACKOFF_TIME
-        print(f"Rate limited, sleeping {wait} seconds.")
+        logger.info(f"Rate limited, sleeping {wait} seconds.")
         time.sleep(wait)
         return 1
     if status >= 400:
-        sys.exit(f"Received unexpected {status} response from API.")
+        logger.error(f"Received unexpected {status} response from API: {response}")
+        return 1
```
AnAccountForReportingBugs: FWIW, I originally did that as `sys.exit()` because my thinking was a client error other than a 403 from the API is probably a permanent condition within the lifetime of the script run. Probably doesn't matter either way.

cmccandless: Intent noted. It still seems the wrong approach to allow the function itself to trigger a full program exit. If we wanted to keep that short-circuit functionality, the best approach would probably be to raise some sort of unique exception to allow the caller to handle the failure however it wishes.

bin/fetch-configlet
```diff
@@ -10,9 +11,13 @@ from urllib.request import urlopen
 from urllib.error import HTTPError, URLError
 
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
```
cmccandless: On-record, I usually avoid using `__name__` for logger as a style preference, as the logger name appears in the default log message format.

yawpitch: That may not be the best habit ... `__file__` isn't consistent across operating systems, and isn't even always present, so you may just bust your file irrevocably with a confusing NameError, but you also may end up with a miles-long absolute path in every logging message. That's on top of breaking the logging hierarchy functionality itself, which could be a issue for downstream consumers.

cmccandless: I don't typically use `__file__` either; I think here I used it because I knew the primary use case would be from the parent directory and no further. Normally, I use constant names (i.e. `getLogger("fetch-configlet")`). I probably should have just gone that route this time as well.

PR #2189, opened 2020-03-10
Test submission for etl exercise

almond55:
Submission for exercises/etl

Test ran with 4 successes.


yawpitch:
@almond55 I'm sorry, but that file is intended to be empty, it's the boilerplate slug that students download when they're going to begin trying to solve the exercise.

If you're trying to submit a submission for mentoring you need to use the Exercism CLI tool to do that with `exercism submit /path/to/etl.py`.

Issue #2190, opened 2020-03-14
Submission bug not caught by tests


chivalry:
I think it's possible to run the tests and pass them all without actually checking all of the alphabet. I think I did this with the following function:

```
def is_pangram(sentence):
    sentence = sentence.lower()
    for c in range(ord('a'), ord('z')):
        if chr(c) not in sentence:
            return False
    return True
```

In the above, I'm not actually checking for the letter "z," but my tests pass.

PR #2191, opened 2020-03-14
Add test to confirm all letters are checked.

chivalry:
My attempt to address issue #2190

### DIFF
--- a/exercises/pangram/pangram_test.py
+++ b/exercises/pangram/pangram_test.py
@@ -41,6 +41,14 @@ class PangramTest(unittest.TestCase):
     def test_case_insensitive(self):
         self.assertIs(is_pangram("the quick brown fox jumps over with lazy FX"), False)
 
+    def test_all_letters_checked(self):
+        alphabet = ''
+        for c in range(ord('a'), ord('z')+1):
+            alphabet += chr(c)
+        for c in range(ord('a'), ord('z')+1):
+            s = alphabet.replace(chr(c), '')
+            self.assertIs(is_pangram(s), False)
+
 
 if __name__ == "__main__":
     unittest.main()


chivalry:
Any help on this would be appreciated. I've looked at the failed checks but don't understand what I messed up.

khoivan88:
It is not you, it just requires others to approve it. Another thing you can do is python 3 has string.ascii_lowercase. so you can do something like:
```python
import string

...
for char in string.ascii_lowercase:
    # do something
...
```

AnAccountForReportingBugs:
The build is failing because `pangram` uses a template and [canonical data](https://github.com/exercism/problem-specifications/tree/master/exercises/pangram). Instead of editing the test file directly, you should follow the directions in the [contributing guide](https://github.com/exercism/python/blob/master/CONTRIBUTING.md) to add the test that way. Although, I'm not sure how exactly this test would fit in there, as Jinja2 is not my strong suit. If a test like this were approved, the use of [subTest](https://docs.python.org/3/library/unittest.html#subtests) would make the tracebacks clearer for those implementing the exercise.

khoivan88:
so, if I am correct, @chivalry , you would need to change [`canonical-data.json`](https://github.com/exercism/problem-specifications/blob/master/exercises/pangram/canonical-data.json) by adding this at the near the end:
```json
,
    {
      "description": "missing letter z",
      "property": "isPangram",
      "input": {
        "sentence": "The quick brown fox jumps over the layy dog"
      },
      "expected": false
    }
```

chivalry:
Thanks to both of you. I'll look more closely at the contributing guide and try again today.

chivalry:
Well, I delved into the contributing guide, and begin to see the problem. This isn't an issue with just the Python track, but with all the tracks. To really fix it would require editing not only the problem specifications, but all of the Jinja templates as well. This one's probably above my head, but I'll look into it a bit more.

khoivan88:
This text quoted below (found [here](https://github.com/exercism/docs/blob/master/contributing-to-language-tracks/README.md#test-suites)) seems to address how you would go about suggesting test that would affect all tracks:
> For every change that you make to the test suite, ask yourself: Should this change also affect other language tracks that implement the same exercise?
> If you think it might, open a discussion in the problem-specifications repository and get other track implementors' feedback on the subject.
> If this change should affect the broader Exercism curriculum, then use the issue in problem-specifications to:
> - Hash out all the details together.
> - Figure out the necessary changes to the canonical-data.json for the exercise (if it exists).
> - Draft an issue that can be submitted to all the tracks that implement the exercise using the blazon tool (which automates the tedious parts). More about that here.

cmccandless:
@chivalry The fastest approach to resolving this issue would be to create a track-specific test here, and also open a pull request to add test(s) to problem-specifications. Once that merges, the track-specific tests could be removed; this is not likely to happen soon as the problem-specifications repo is currently frozen while the v3 redesign is underway.

The test as you have written it does not easily transition into the template format, but I have created #2192 to show another approach that should address the gap you have pointed out in #2190. Please take a look at that PR and let me know what you think.

PR #2192, opened 2020-03-17
pangram: add tests for alphabetical min/max

### LINKED ISSUE #2190: Submission bug not caught by tests

cmccandless:
Resolves #2190

Closes #2191 by mutual exclusion.

### DIFF
--- a/exercises/pangram/.meta/additional_tests.json
+++ b/exercises/pangram/.meta/additional_tests.json
@@ -1,0 +1,20 @@
+{
+  "cases": [
+    {
+      "description": "sentence without lower bound",
+      "property": "isPangram",
+      "input": {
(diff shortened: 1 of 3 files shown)


AnAccountForReportingBugs:
I think that looks good. Will catch people who don't realize how `range` works on this exercise while leaving more insidious bugs for the reviewer and not cluttering the tests file with 24 of their friends.

cmccandless:
@exercism/python Can I please get a review from one of you?

PR #2194, opened 2020-03-24
Add CPython specific test case

citizen-stig:
I saw that some students use `is not` for comparing that have different length.
This actually works for strings smaller than 257 symbols, so they don't catch that error

By offloading this case to test case, then can learn about it offline.


cmccandless:
I'm with @yawpitch on this one. Pointing out the flaw in such solutions seems the responsibility of a mentor, automated or otherwise.

I do like the idea of a V3 Concept exercise teaching right and wrong used of the `is` operator. It's something more unique to Python.

citizen-stig:
Thanks for the feedback, I didn't know about V3 ideas and concepts!

PR #2196, opened 2020-05-03
Check supplied argument is actually an integer.

WeeDom:
See https://exercism.io/mentor/solutions/48683af98ace4142b4c13b24f5180853#discussion-post-658995

### DIFF
--- a/exercises/raindrops/raindrops_test.py
+++ b/exercises/raindrops/raindrops_test.py
@@ -61,6 +61,9 @@ class RaindropsTest(unittest.TestCase):
 
     def test_the_sound_for_3125_is_plang_as_it_has_a_factor_5(self):
         self.assertEqual(convert(3125), "Plang")
+        
+    def test_non_int_supplied(self):
+        self.assertRaises('TypeError', convert('not an int'))
 
 
 if __name__ == "__main__":


yawpitch:
Since we control the input values and the spec we provide does not allow for or imply non-`int` input values, I'd consider this test overkill.

Where possible we _do not_ want to teach any habits that disrupt or interfere with duck-typing, and type-checking input values definitely does that.

By adding that `type(number) != int` test the student has automatically stopped their library's downstream user from passing `float`, `decimal`, `fraction`, and `complex` values even if they were otherwise entirely appropriate **number**s. By not using `isinstance(number, int)`  they've even stopped the user from presenting an instance of a valid subclass of `int` ... and both of these are _usually_ examples of overly-aggressive policing of input values that break duck-typing.

I appreciate the effort of submitting a PR, but I don't believe this is one we'd want to merge for the above reasons. The failing check, BTW, is because our test files are generated from upstream data and we'd need to provide these tests in a generation-friendly manner (see https://github.com/exercism/python/blob/master/CONTRIBUTING.md#architecture and specifically the bit about ./meta/additional_tests.json).

WeeDom:
thanks for the comprehensive feedback.

Submitting the PR wasn't as time-consuming as your feedback, I don't think!

yawpitch:
> Submitting the PR wasn't as time-consuming as your feedback, I don't think!

Quite likely. I do tend to type a lot. ;-)

PR #2197, opened 2020-05-03
check name is not total garbage

WeeDom:
The check could be stronger, I know, but I didn't fancy forcing the students down that line. It's just that I've seen a few coming in with various attempts at type-checking the argument, so if they're thinking of it then maybe the test suite could too

### DIFF
--- a/exercises/two-fer/two_fer_test.py
+++ b/exercises/two-fer/two_fer_test.py
@@ -15,6 +15,8 @@ class TwoFerTest(unittest.TestCase):
     def test_another_name_given(self):
         self.assertEqual(two_fer("Bob"), "One for Bob, one for me.")
 
-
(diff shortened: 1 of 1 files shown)


yawpitch:
As mentioned in #2196 we do not really want to encourage explicit type-checking of input values, as such policing of inputs unnecessarily inhibits otherwise valid attempts at duck-typing. The spec given and the tests present only string inputs, which means that reacting to non-string inputs is simply out of the scope of this exercise. Thus it's really up to a mentor to comment (or not) on a student's habit of doing so, but adding this test would _force_ students to write boilerplate that _usually_ represents a bit of a code smell.

So for this one as well I'd say thank you very much for the effort of making a submission, but personally I'd say we shouldn't add this test.

PR #2198, opened 2020-05-03
chore- two_fer add test name is empty string

galambi:
* chore- two_fer add test name is empty string

### DIFF
--- a/exercises/two-fer/README.md
+++ b/exercises/two-fer/README.md
@@ -10,7 +10,7 @@ One for X, one for me.
 
 Where X is the given name.
 
-However, if the name is missing, return the string:
+However, if the name is missing or empty, return the string:
 
 ```text
 One for you, one for me.
--- a/exercises/two-fer/two_fer_test.py
+++ b/exercises/two-fer/two_fer_test.py
@@ -15,6 +15,9 @@ class TwoFerTest(unittest.TestCase):
     def test_another_name_given(self):
         self.assertEqual(two_fer("Bob"), "One for Bob, one for me.")
 
+    def test_empty_name_given(self):
+        self.assertEqual(two_fer(""), "One for you, one for me.")
+
 
 if __name__ == "__main__":
     unittest.main()


galambi:
I raised this PR because I think this exercise is lacking a critical test. If the name is provided but it is an empty string, you get an ugly return message: "One for , one for me."

Perhaps its an opportunity to learn tenary operator. I would expect a solution like this:

def two_fer(name:str="you") -> str:
	"""Return the message with the name included"""
	return f"One for {name if name else 'you'}, one for me."

galambi:
I didn't see a relation between the failed build and my change. But I obviously am the one likely to be mistaken here.
https://travis-ci.org/github/exercism/python/jobs/682703748

If you like the essence of my PR and would like to include it. Please let me know how I can help fix things to get it in.

yawpitch:
The failed build is due to the test being a direct change to an automatically-generated file, see https://github.com/exercism/python/blob/master/CONTRIBUTING.md#architecture for an explanation of how tests must either be added to the .meta/additional_tests.json file, or if more complex to the .meta/template.j2 file.

As for the PR itself, I appreciate the contribution, but I'm going to disagree on the need for this. While an empty string is, indeed, somewhat goofy looking, there's no way the author of a library can meaningfully decide that such a name is **wrong** _per se_ ... if the downstream user has gone to the trouble of presenting the empty string then replacing that string with _any_ other value is, at best, a guess, and is at worst an abuse of the user's autonomy, right? "In the face of ambiguity, resist the temptation to guess" is the bit of the Zen of Python I'm alluding to here.

Realistically the purpose of a default value is _only_ to provide a fallback if no value is provided at all by the user, thus:

```
two_fer()
two_fer("")
two_fer(None)
```

... _should_ arguably each give a different result (or possibly an error), since the downstream user either _intentionally_ omitted an argument (and therefore expects the default), or they _intentionally_ presented a False-y value and we shouldn't override them, even if it means throwing up our hands and failing.

So I'd personally say thank you for the effort, but I'd tend to reject this PR. I'll leave it for a few days to see if there are any other thoughts to consider from others.

galambi:
> The failed build is due to the test being a direct change to an automatically-generated file, see https://github.com/exercism/python/blob/master/CONTRIBUTING.md#architecture for an explanation of how tests must either be added to the .meta/additional_tests.json file, or if more complex to the .meta/template.j2 file.
> 
> As for the PR itself, I appreciate the contribution, but I'm going to disagree on the need for this. While an empty string is, indeed, somewhat goofy looking, there's no way the author of a library can meaningfully decide that such a name is **wrong** _per se_ ... if the downstream user has gone to the trouble of presenting the empty string then replacing that string with _any_ other value is, at best, a guess, and is at worst an abuse of the user's autonomy, right? "In the face of ambiguity, resist the temptation to guess" is the bit of the Zen of Python I'm alluding to here.
> 
> Realistically the purpose of a default value is _only_ to provide a fallback if no value is provided at all by the user, thus:
> 
> ```
> two_fer()
> two_fer("")
> two_fer(None)
> ```
> 
> ... _should_ arguably each give a different result (or possibly an error), since the downstream user either _intentionally_ omitted an argument (and therefore expects the default), or they _intentionally_ presented a False-y value and we shouldn't override them, even if it means throwing up our hands and failing.
> 
> So I'd personally say thank you for the effort, but I'd tend to reject this PR. I'll leave it for a few days to see if there are any other thoughts to consider from others.

Good response. Thanks for the thorough explanation.

Issue #1944, opened 2019-10-02
[V2] diffie-hellman: add test template


cmccandless:
## Task
Create a test template file for `diffie-hellman`

### Details
1. If this is your first time contributing to this repository, welcome!
    - Contributing to a repository on GitHub requires a [fork](https://help.github.com/en/articles/fork-a-repo)
1. Create a branch on your local repository called `diffie-hellman-add-test-template`
    - Run: `git checkout -b diffie-hellman-add-test-template`
1. Clone problem-specifications repository; this is used to populate the template
    - Run: `git clone https://github.com/exercism/problem-specifications ../problem-specifications`
1. Create `exercises/diffie-hellman/.meta/template.j2`
    - See [Generator Documentation] for more information
1. Generate test file from new template
    - Run: `bin/generate_tests.py diffie-hellman`

At this point, if the generator fails, or you are otherwise stuck,
please feel free to reach out! It is perfectly acceptable to submit
an incomplete PR in order to get feedback or assistance.

### Resources
- [Contributing Guide]
- [Generator Documentation]
- [Closing issues using keywords]

[Contributing Guide]: https://github.com/exercism/python/blob/master/docs/CONTRIBUTING.md
[Generator Documentation]: https://github.com/exercism/python/blob/master/docs/GENERATOR.md
[Closing issues using keywords]: https://help.github.com/en/articles/closing-issues-using-keywords

crsmi:
I'd like to take this one.

denvaar:
Me too

cmccandless:
@denvaar I'm not sure the issue really lends itself to two simultaneous contributors. If you and @crsmi wish to collaborate, that is between you, but he did claim the issue prior to you. If @crsmi is not open to collaboration on this issue, please feel free to claim another open issue.

pyropy:
@cmccandless Is this stale issue or I could take it?

cmccandless:
@pyropy 
> Is this stale issue
Yes
> or I could take it?
Also yes. Contributors always welcome!

BethanyG:
Missed that this was fixed and merged.  Closing.

PR #2203, opened 2020-05-11
diffie-hellman: add test template 

pyropy:
-Add jinja2 test template for Diffie Hellman exercise #1944 
- Split `to_snake` into two functions (`to_snake_base` and `to_snake`)
- Register `to_snake_base` jinja2 macro
- Generate new test template

### DIFF
--- a/exercises/diffie-hellman/.meta/template.j2
+++ b/exercises/diffie-hellman/.meta/template.j2
@@ -23,9 +23,9 @@ class {{ exercise | camel_case }}Test(unittest.TestCase):
     {% elif property == "key_exchange"%}
     def test_{{ description }}(self):
         {% for key, val in input.items() -%}
-            {{ key | to_snake }} = {{ val  | string | to_snake_base }}
+            {{ key | to_snake }} = {{ val | string | to_snake(wordchars_only=True) }}
         {% endfor -%}
-        self.assertTrue({{ expected | to_snake_base }})
+        self.assertTrue({{ expected | to_snake(wordchars_only=True) }})
     {%  else %}
     def test_{{ description }}(self):
         self.assertEqual({{ expected }}, {{ property }}(*{{ input.values() | list }} ))
--- a/exercises/diffie-hellman/.meta/template.j2
+++ b/exercises/diffie-hellman/.meta/template.j2
@@ -14,8 +14,10 @@ class {{ exercise | camel_case }}Test(unittest.TestCase):
         for p in primes:
             self.assertTrue(p < private_key(p) < p)
     {%- elif property == "private_key_is_random" %}
-        # Can fail due to randomness, but most likely will not,
-        # due to pseudo-randomness and the large number chosen
+        """
+        Can fail due to randomness, but most likely will not,
+        due to pseudo-randomness and the large number chosen
+        """
         p = 2147483647
         private_keys = [private_key(p) for _ in range(5)]
         self.assertEqual(len(set(private_keys)), len(private_keys))
--- a/exercises/diffie-hellman/.meta/template.j2
+++ b/exercises/diffie-hellman/.meta/template.j2
@@ -1,0 +1,35 @@
+{%- import "generator_macros.j2" as macros with context -%}
+{% set class = exercise | camel_case -%}
+{{ macros.header(["private_key", "public_key", "secret"]) }}
+
+class {{ exercise | camel_case }}Test(unittest.TestCase):
+    {% for case in cases -%}
+    {% set property = case["property"] | to_snake -%}
+    {% set description = case["description"] | to_snake -%}
+    {% set expected = case['expected'] %}
+    {% set input = case['input'] %}
+    {% if property == "private_key_is_in_range" -%}
+    def test_{{ description }}(self):
+        primes = [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
+        for i in primes:
+            self.assertTrue(1 < private_key(i) < i)
+    {% elif property == "private_key_is_random" -%}
+    def test_{{ description }}(self):
+        p = 2147483647
+        private_keys = []
+        for i in range(5):
+            private_keys.append(private_key(p))
+        self.assertEqual(len(set(private_keys)), len(private_keys))
+    {% elif property == "key_exchange"%}
+    def test_{{ description }}(self):
+        {% for key, val in input.items() -%}
+            {{ key | to_snake }} = {{ val  | string | to_snake_base }}
+        {% endfor -%}
+        self.assertTrue({{ expected | to_snake_base }})
+    {%  else %}
+    def test_{{ description }}(self):
+        self.assertEqual({{ expected }}, {{ property }}(*{{ input.values() | list }} ))
+    {%- endif -%}
+    {% endfor %}
+
+{{ macros.footer() }}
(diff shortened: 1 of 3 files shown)


pyropy:
> Well done on the template, but it's not exactly _obvious_ from the docstrings that `to_snake_base` isn't converting whitespace and punctuation, while `to_snake` is ...
> 
> Consider instead something like:
> 
> ```python
> def to_snake(string, wordchars_only=False):
>     """
>     Converts pretty much anything to snake case.
> 
>     By default whitespace and punctuation will be converted to underscores as well, pass wordchars_only=True to preserve these as is.
>     """
>    [...]
> ```
> 
> Then only the one filter is required and when needed it can be called with `| to_snake(wordchars_only=True)`?

Yeah I'll try that out. Wasn't quite sure how would that work with Jinja2

exercises/diffie-hellman/.meta/template.j2
```diff
@@ -0,0 +1,35 @@
...
+class {{ exercise | camel_case }}Test(unittest.TestCase):
+    {% for case in cases -%}
+    {% set property = case["property"] | to_snake -%}
+    {% set description = case["description"] | to_snake -%}
+    {% set expected = case['expected'] %}
+    {% set input = case['input'] %}
+    {% if property == "private_key_is_in_range" -%}
+    def test_{{ description }}(self):
+        primes = [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
+        for i in primes:
```
yawpitch, line 14: It's a very minor thing, but could we change the temp value to **p** here, since each is a prime? It'd also jive more with the comments in the canonical data.

exercises/diffie-hellman/.meta/template.j2
```diff
@@ -0,0 +1,35 @@
...
+            self.assertTrue(1 < private_key(i) < i)
+    {% elif property == "private_key_is_random" -%}
+    def test_{{ description }}(self):
+        p = 2147483647
```
yawpitch, line 18: I'd prefer we did this by mocking **random**, really, but this approach will do for now ... that said, let's do all this generation in a list comprehension: `private_keys = [private_key(2147483647) for _ in range(5)]`

exercises/diffie-hellman/.meta/template.j2
```diff
@@ -0,0 +1,35 @@
...
+        private_keys = []
+        for i in range(5):
+            private_keys.append(private_key(p))
+        self.assertEqual(len(set(private_keys)), len(private_keys))
+    {% elif property == "key_exchange"%}
+    def test_{{ description }}(self):
+        {% for key, val in input.items() -%}
+            {{ key | to_snake }} = {{ val  | string | to_snake_base }}
```
cmccandless: Can you explain why `to_snake` was inadequate here?

exercises/diffie-hellman/.meta/template.j2
```diff
@@ -0,0 +1,35 @@
...
+        {% endfor -%}
+        self.assertTrue({{ expected | to_snake_base }})
+    {%  else %}
+    def test_{{ description }}(self):
+        self.assertEqual({{ expected }}, {{ property }}(*{{ input.values() | list }} ))
```
cmccandless, line 31: A couple of things here. First, the generated code is much cleaner if you perform the unpacking in the template instead:
```suggestion
        self.assertEqual({{ expected }}, {{ property }}({% for val in input.values() %}{{ val }},{% endfor %}))
```

Second, simply unpacking `input.values()` directly into the function arguments assumes the order of `input.values()` is correct; this is not guaranteed. So I would prefer that we list the expected input values explicitly:
```suggestion
        self.assertEqual(
            {{ expected }},
            {{ property }}(
                {{ input["p"] }},
                {{ input["theirPrivateKey"] }},
                {{ input["myPrivateKey"] }},
            )
        )
```

exercises/diffie-hellman/.meta/template.j2
```diff
@@ -0,0 +1,35 @@
...
+{{ macros.header(["private_key", "public_key", "secret"]) }}
+
+class {{ exercise | camel_case }}Test(unittest.TestCase):
+    {% for case in cases -%}
+    {% set property = case["property"] | to_snake -%}
+    {% set description = case["description"] | to_snake -%}
+    {% set expected = case['expected'] %}
+    {% set input = case['input'] %}
+    {% if property == "private_key_is_in_range" -%}
+    def test_{{ description }}(self):
```
cmccandless, line 12: The function declaration is identical for all cases here, so it can be moved outside the `if/elif/else` blocks.

exercises/diffie-hellman/diffie_hellman_test.py
```diff
@@ -1,56 +1,40 @@
...
-    def test_private_key_is_in_range(self):
+class DiffieHellmanTest(unittest.TestCase):
+    def test_private_key_is_in_range_1_p(self):
         primes = [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
         for i in primes:
-            self.assertTrue(1 < diffie_hellman.private_key(i) < i)
+            self.assertTrue(1 < private_key(i) < i)
 
-    # Can fail due to randomness, but most likely will not,
-    # due to pseudo-randomness and the large number chosen
```
cmccandless, line 16: This comment may be included as a function docstring inside the `elif property == "private_key_is_random"` block.

exercises/diffie-hellman/diffie_hellman_test.py
```diff
@@ -1,56 +1,40 @@
...
         self.assertEqual(len(set(private_keys)), len(private_keys))
 
     def test_can_calculate_public_key_using_private_key(self):
-        p = 23
-        g = 5
-        private = 6
-        expected = 8
-
-        actual = diffie_hellman.public_key(p, g, private)
-        self.assertEqual(actual, expected)
```
cmccandless, line 31: I would prefer to keep the verbose variable declarations in this test for clarity on the usage of `public_key`

exercises/diffie-hellman/diffie_hellman_test.py
```diff
@@ -1,56 +1,40 @@
...
+        self.assertEqual(8, public_key(*[23, 5, 6]))
 
     def test_can_calculate_secret_using_other_party_s_public_key(self):
-        p = 23
-        public = 19
-        private = 6
-        expected = 2
-
-        actual = diffie_hellman.secret(p, public, private)
-        self.assertEqual(actual, expected)
```
cmccandless, line 40: I would prefer to keep the verbose variable declarations in this test for clarity on the usage of `secret`

exercises/diffie-hellman/.meta/template.j2
pyropy: Sure thing, I've copied mostly of that from the previous test case but I think your right there. Where there's change to improve something - it should be done.

exercises/diffie-hellman/.meta/template.j2
pyropy: In this case most of the input most values are function calls, therefore old `to_snake` function would replace punctuation (brackets and commas in this case) with underscore. 

I've reveted `to_snake_base` back to one function but now `to_snake` takes optional that @yawpitch suggested.

exercises/diffie-hellman/.meta/template.j2
pyropy: I've done unpacking like you suggested but instead of unpacking values i"ve unpacked keys.

I've also created variables for given keys in order to make tests more verboese.

exercises/diffie-hellman/.meta/template.j2
pyropy: Lol I've changed this anf for some reason tests are failing haha.

PR #2204, opened 2020-05-13
yacht: Fixed wrong info about score

toguko:
The info on the "Score Threes" was showing the wrong information.

If all faces are the same, it will be a Yacht, so the score should be 50, not 15.
FROM:
| Category | Score | Description | Example |
| -------- | ----- | ----------- | ------- |
| Threes | 3 × number of threes | Any combination | 3 3 3 3 3 scores 15 |

TO:
| Category | Score | Description | Example |
| -------- | ----- | ----------- | ------- |
| Threes | 3 × number of threes | Any combination | 3 3 3 3 4 scores 12 |

### DIFF
--- a/exercises/yacht/README.md
+++ b/exercises/yacht/README.md
@@ -14,7 +14,7 @@ category chosen.
 | -------- | ----- | ----------- | ------- |
 | Ones | 1 × number of ones | Any combination	| 1 1 1 4 5 scores 3 |
 | Twos | 2 × number of twos | Any combination | 2 2 3 4 5 scores 4 |
-| Threes | 3 × number of threes | Any combination | 3 3 3 3 3 scores 15 |
+| Threes | 3 × number of threes | Any combination | 3 3 3 3 4 scores 12 |
 | Fours | 4 × number of fours | Any combination | 1 2 3 3 5 scores 0 |
 | Fives | 5 × number of fives| Any combination | 5 1 5 2 5 scores 15 |
 | Sixes | 6 × number of sixes | Any combination | 2 3 4 5 6 scores 6 |


yawpitch:
I can see that this does, in a sense, clear up a potential point of confusion, however I'm not sure the change is necessary.

Per the README, "[the] result can be entered **in any of twelve** categories" and "[t]he score of a throw of the dice **depends on [the] category** chosen" (emphasis mine) ... so it would be perfectly valid to enter **any** of these scores for a roll of `3 3 3 3 3`:

| Category | Score |
| -------- | ----- |
| Ones | 0 |
| Twos | 0 |
| Threes | 15 |
| Fours | 0 |
| Fives | 0 |
| Sixes | 0 |
| Full House | 15 |
| Four of a Kind | 12 |
| Little Straight |  0 |
| Big Straight | 0 |
| Choice | 15 |
| Yacht | 50 |

Now, **obviously** any rational player would choose the "yacht" category, but 15 remains the valid score for counting `3 3 3 3 3` in the "threes" category ... the function we're asking students to create doesn't know anything about the play-correct scoring, it merely converts a given list of rolls to a given score given a specific category.

Based on that I would personally call this change unnecessary. Moreover, if necessary, we should suggest the change on the [upstream description.md][https://github.com/exercism/problem-specifications/blob/master/exercises/yacht/description.md] rather than here in the track.

cmccandless:
In the game of Yacht, each category may only be scored once. So if the player has already scored for Yacht (category), they can't use it for their subsequent turns, even if it qualifies, and must use other categories. So not only is `3 3 3 3 3` a valid roll for the Threes category, in many games it might be essential to do as such.

This is also the reason for returning a 0 for rolls that do not qualify for the given category: if a player rolls a combination that does not match unused categories, they must take a score of 0 for one of their remaining categories.

Thus, the current description is perfectly valid whether or not we consider a single turn in isolation or a full game.

yawpitch:
Thanks @cmccandless, having never played (or indeed heard of) the game I didn't know those details.

@toguko, given the above I'm going to close this as wontfix, as the original reflects a correct input and output and there's no obvious improvement in clarity by changing to the new values, while the old values have the added benefit of actually existing in the canonical test data as well.

Thank you for the submission though.

PR #2208, opened 2020-06-22
Update README.md

WeeDom:
Clarify instructions to try to prevent so many submissions that miss the point and just add a typed out dict of letter:score.

### DIFF
--- a/exercises/scrabble-score/README.md
+++ b/exercises/scrabble-score/README.md
@@ -34,6 +34,17 @@ And to total:
 - = `5 + 9`
 - = 14
 
+## Bad Examples
+
+Don't do this:
+
+```
+scores = { 'a':1, 'b':3, 'c':3..., 'z':10}
+```
+That's missing the point. The challenge is to make a readable, maintainable, structure using dict/tuples/lists/etc and then using your pythonista skills, transform that into a structure which can be easily accessed by the function `score(word`)`.  
+
+Adding up scores is easy - we're looking for data transformation skills here!
+
 ## Extensions
 
 - You can play a double or a triple letter.


yawpitch:
I’m on leave from maintaining ATM, so take this with as much of a grain of salt as you’d like, but as I’m still occasionally mentoring occasionally and I got the notification I thought I’d pipe in.

First, the “bad examples” language doesn’t actually get followed by an example, nor does it improve the student’s understanding of what the function itself is supposed to do, instead it implies that there’s a “better” implementation of the score lookup. If we include this language the title of the heading should better reflect what’s being conveyed. 

Second, the tags for this don’t imply that data transformation is a goal (or even a point to consider)... I’d argue that a literal dict is the most obvious (and a perfectly valid) data structure to use. An **enum.Enum** would be another good option; we don’t need to encourage code golf and we shouldn’t really imply that there’s a benefit to using runtime data transformations to build what is effectively always going to be a constant. So personally I’d either say we make data transformations an explicit goal of this exercise, with improved language specifying that a constant should NOT be used, or we should simply allow through a literal dict, perhaps with a comment that you _could_ build it programmatically you save some LOC.

Personally I’d lean on preferring the status quo and adding something to the mentor notes to the effect of there being many ways of skinning this particular cat, but of course it’s up to the mentors as a whole which is better.

cmccandless:
I'm personally against adding this sort of language directly to the README. As @yawpitch pointed out, there's nothing wrong with a solution that uses a direct mapping (as long as the mapped data set is not too big; 26 key-value-pairs is definitely small enough). If a student produces such a solution, I _do_ think it is reasonable for a mentor to approve, but suggest improvements that can avoid a direct mapping. Learning such methods is useful for real-world applications where the data set might be significantly larger.

The goal of Exercism and the mentoring process is to improve one's proficiency in a given language. I would hate for less experienced programmers to get caught up on a "no direct mapping" rule, and then give up because they were unable to find another way on their own. I would much rather a student submit a passing, but non-optimal, solution to open the mentoring conversation, so that they can grow and learn.

WeeDom:
I'm absolutely on-track with the 'submit non-optimal solution as an opening gambit' suggestion.  

I'll draft up an amendment to the mentoring notes, then, and we can review it there?

Thanks for the, as always, positive feedback. I'm still relatively new, here, and appreciate that I'm not up to speed on everything!

WeeDom:
I'll just add that the way I've been approaching mentoring the examples where they've come back with a 26 letter long dict is something along the lines of :

Nice work - but what we're really looking for is a solution that isn't just typing out a `dict`. We're looking to see your data transformation skills. Your `sum` is great, absolutely perfect, but let's see if you come up with something a little easier to maintain. Use tuples/lists/dicts/whatever to construct a more readable structure, maybe grouped by scores, then showcase your python skills to transform that into a data structure that your `score` can use. Don't change `score` - that's great! Just change how you're constructing the data that's fed into it.

I'll work on the wording, because that does seem a little verbose!

PR #2209, opened 2020-06-26
word-count: Add you're test case

szTheory:
There's a blind spot in the test suite for you're

### DIFF
--- a/exercises/word-count/word_count_test.py
+++ b/exercises/word-count/word_count_test.py
@@ -42,8 +42,8 @@ class WordCountTest(unittest.TestCase):
 
     def test_with_apostrophes(self):
         self.assertEqual(
-            count_words("First: don't laugh. Then: don't cry."),
-            {"first": 1, "don't": 2, "laugh": 1, "then": 1, "cry": 1},
+            count_words("You're unflappable. First: don't laugh. Then: don't cry."),
+            {"you're": 1, "unflappable": 1, "first": 1, "don't": 2, "laugh": 1, "then": 1, "cry": 1},
         )
 
     def test_with_quotations(self):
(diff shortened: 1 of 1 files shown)


cmccandless:
Welcome! Thanks for the PR! As described in our [Contributing Guide](https://github.com/exercism/python/blob/master/CONTRIBUTING.md), exercises here inherit their test cases from the [problem-specifications repository](https://github.com/exercism/problem-specifications/), so we usually prefer that new test cases are added there first, then copied to this repository. Please feel free to open a PR over there for new test cases.

That said, my first reaction to this suggested test: isn't there already a test case for words with an apostrophe in the middle?
```Python
    def test_with_apostrophes(self):
        self.assertEqual(
            count_words("First: don't laugh. Then: don't cry."),
            {"first": 1, "don't": 2, "laugh": 1, "then": 1, "cry": 1},
        )
```

szTheory:
Thanks, gotcha. Yeah there's a test for single letter after apostrophe, but not multiple letters so it's different and could affect regular expressions that only work for "let's" but not "you're" for example.

cmccandless:
A valid point. Might I suggest integrating "you're" into the existing test somehow, instead of writing a separate test?

szTheory:
Good idea, I updated the test to integrate it into an existing one instead of writing a separate test. What do you think?

cmccandless:
I like the improvement! However, this still cannot be merged here because this test is inherited from the [problem-specifications](https://github.com/exercism/problem-specifications/) repository. Changes will need to be made there, then the test code here may be re-[generated](https://github.com/exercism/python/blob/master/docs/GENERATOR.md) to incorporate these changes.

szTheory:
@cmccandless  Thanks! I'm a bit busy for that but maybe I'll circle back to it some other time.

PR #2210, opened 2020-06-30
Added test case with an integer at two fer exercise

AkibAzmain:
Hi, I am a mentor of python track. I often see people trying to concatenate strings like this:

```python
return "One for " + name + ", one for me."
```

But this causes the function to malfunction if argument is an integer. So, I added a test which will check if the function rewritten by user is able to work with integer.

### DIFF
--- a/exercises/two-fer/two_fer_test.py
+++ b/exercises/two-fer/two_fer_test.py
@@ -14,6 +14,9 @@ class TwoFerTest(unittest.TestCase):
 
     def test_another_name_given(self):
         self.assertEqual(two_fer("Bob"), "One for Bob, one for me.")
+    
+    def test_an_integer_given(self):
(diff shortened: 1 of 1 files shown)


cmccandless:
`two-fer` is a very basic exercise. As such, testing with unexpected data types for inputs is considered out-of-scope.

Additionally, the [mentor notes](https://github.com/exercism/website-copy/blob/master/tracks/python/exercises/two-fer/mentoring.md) for `two-fer` mention String Interpolation (or "f-strings") as a recommended talking point. In general [string interpolation is recommended over concatenation](https://stackoverflow.com/a/38722192/3229611), and a solution using interpolation will not error unless the argument provided is of a type that somehow does not have a `__str__()` method.

PR #2217, opened 2020-08-27
phone-number: standardize pretty format

### LINKED ISSUE #2216: phone-number standard format?
The Phone Number exercise seems to be confused on what format it wants as the standard.

In `README.md`, it says this format is standard:

```
(NXX)-NXX-XXXX
```

but the tests for `pretty()` want this format:
```
(NXX) NXX-XXXX
```

I think these two should be unified.

cmccandless:
Resolves #2216


J08K:
Some simple changes. I like it!

PR #2218, opened 2020-09-11
say: fix for numbers divisible by 10, but not ending in 10 or 20

AndrejTS:


cmccandless:
@AndrejTS good catch, thanks!

PR #2221, opened 2020-09-30
Update linked_list.py

Prathameshyadav:


yawpitch (commented):
@Prathameshyadav thank you for the submission, but this file is intentionally left without an implementation, as it’s the “dummy” slug file the student is delivered and which they’re meant to fill in with their work.

PR #2222, opened 2020-09-30
Improved docs

SURAJ-KUNWAR:


yawpitch (commented):
In this case “as well” is redundant as it’s already implied by “both”.

PR #2223, opened 2020-09-30
adding currency currectly.

likeitaash:
8/octo/2020


yawpitch (commented):
This site does not use country-specific currency designations, as users may be in any number of potential countries.

PR #2263, opened 2020-10-15
Test generator: filter test cases using tests.toml

cmccandless:
Recommend using a normal Merge commit instead of squashing, as there are commits here for separate exercises.

Normally, I would have committed those dedicated branches, but the other changes here required the exercises to be updated to pass CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ bin/configlet.exe
 .pytest_cache
 __pycache__
 .venv
+bin/canonical_data_syncer
+bin/canonical_data_syncer.exe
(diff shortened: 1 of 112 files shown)


ErikSchierboom (approved):
LGTM (I don't know Python though)

cmccandless:
@ErikSchierboom I figured I'd tag you in just in case you were tracking somewhere which language tracks had up-to-date test generators.

BethanyG (approved):
`generate_tests.py` & `generator_macros.j2` look good to me.  I am going to also spot-check the problems, but I probably won't look at every single one.

ErikSchierboom:
Thanks for the tag @cmccandless! I appreciate it.

PR #2273, opened 2020-11-11
Update reverse_string.py

Zeaan:
Created a project which reverses a String


cmccandless:
Hello! It looks like you may be trying to submit a solution to the `reverse-string` exercise. The correct way to submit solutions is via the Exercism CLI. Please follow the interactive setup instructions [here](https://exercism.io/cli-walkthrough).

The file you have edited here (reverse_string.py) is the file that the CLI will download as a starting point when you start this exercise. As such, no changes are required to it here.

Zeaan:
Oh ok.

PR #2284, opened 2021-01-28
Update CI test runner for v3

cmccandless:
- Check for exemplar.py (>=v3), then example.py (<v3)
	- See https://github.com/exercism/v3/pull/3086
- Use pathlib instead of os.path

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - master
+      - main
   pull_request:
 
 jobs:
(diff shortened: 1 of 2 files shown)


cmccandless:
Maintainer note: squash when merging

BethanyG (commented):
This looks good to me -- but with two questions:

1.  Is our `master` branch going to be renamed `main`?  If so, we'll need to change the ref.
2. Should we hold this merge until after the "explosion"?  We have some pending exercise PRs to finish up.

Otherwise, I think we are good to go 🌟

cmccandless:
> Is our master branch going to be renamed main? If so, we'll need to change the ref.

I've added main to target branches

> Should we hold this merge until after the "explosion"? We have some pending exercise PRs to finish up.

I don't think that's necessary. If 'exemplar.py' is missing, it looks next for 'example.py', so this should be backwards compatible.

BethanyG (approved):
👍🏽

PR #2320, opened 2021-02-04
Run unit tests with python-test-runner

cmccandless:
add workflow and script for running all exercises against python-test-runner container

Closes #2334 (commits cherry-picked here)


exercises/practice/paasio/test_utils.py
```diff
@@ -0,0 +1,146 @@
+import errno
+import inspect
+import io
+import os
+
+
```
BethanyG, line 6:
```suggestion
LOGGER = logging.getLogger(__name__)
```

BethanyG: I can't suggest on the main test file here, but if you change the import on line 6 there to be:

`from test_utils import MockSock, MockFile, MockException, ZEN, SuperMock, LOGGER`

and remove  `LOGGER = logging.getLogger(__name__)`

then that should fix it.

BethanyG (approved):
Looking good!  🌟 🚀

PR #2328, opened 2021-02-06
use pytest in CI script

cmccandless:
Certain exercises were missing the `unittest.main` footer, preventing their tests from running in CI. As much of our tooling depends on pytest anyway, this PR changes the script to invoke pytest from the current Python executable. Will fail if pytest not installed for that Python version.


BethanyG (approved):
LGTM!  🌟   Thank you for fixing this.

PR #2371, opened 2021-02-26
🤖 Update tests.toml files to latest spec

ErikSchierboom:
Track maintainers found that they wanted to add comments to the tests.toml file to e.g. indicate _why_ a test was not included.
Unfortunately, running configlet sync would re-generate the entire file so any manually added comments were lost.

In this PR we're updating the format of tests.toml files to support adding comments.
We do this by creating a separate table for each test case which has `description` and `include` fields.
Tracks are then free to add additional fields, like a `comment` field, but also any other fields they feel might be useful to them.

configlet has _not_ yet been updated to support this new format, but we hope to do this soon. Sorry for the inconvenience.

For more information, see this discussion: https://github.com/exercism/configlet/issues/186

## Implementation

The PR expects the tests.toml files to be in their original format:

```toml
# <description>
"<uuid>" = <include>
```

This is transformed to:

```toml
[<uuid>]
description = "<description>"
include = <include>
```

## Example

```toml
# one factor has multiples within limit
"361e4e50-c89b-4f60-95ef-5bc5c595490a" = true
```

becomes

```toml
[361e4e50-c89b-4f60-95ef-5bc5c595490a]
description = "one factor has multiples within limit"
include = true
```

## Existing comments

As some tracks have manually added comments to tests, we try to detect them by assuming they are either:

- Added as a line comment _before_ the description (we'll ignore empty lines)
- Added as an inline comment _after_ boolean include value

For any such manually detected comments, we'll add a `comment` field.

## Tracking

https://github.com/exercism/v3-launch/issues/22

### DIFF
--- a/bin/data.py
+++ b/bin/data.py
@@ -3,7 +3,8 @@ from dataclasses import dataclass, asdict
 from itertools import chain
 import json
 from pathlib import Path
-from typing import List, Any
+import toml
+from typing import List, Any, Dict
 
 
 @dataclass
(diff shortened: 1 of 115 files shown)


BethanyG (approved):

ErikSchierboom:
@BethanyG There is a test failure. You probably want to fix that before I merge, right?

BethanyG:
@ErikSchierboom  -- pinged @cmccandless, since I am not sure why the failure is happening.  Nothing changed to produce it.  I'll investigate further today.

cmccandless:
I'll take a look

cmccandless:
Tooling updated

PR #2378, opened 2021-03-03
Default TestCaseTOML.include to True

cmccandless:
Per discussion in:
- https://github.com/exercism/configlet/issues/186
- https://github.com/exercism/configlet/issues/207


BethanyG (approved):
Good catch!

PR #2381, opened 2021-03-05
secret-handshake: make input a binary string instead of integer

### LINKED ISSUE #2380: secret-handshake: Rewrite tests and example solution to make input a binary string instead of integer
For practicing working with binary strings. Example:
`commands(19)`=>`commands('10011')`

### Checklist:
- [ ] Update template to format input as binary string
- [ ] Generate tests
- [ ] Update .meta/example.py

cmccandless:
Resolves #2380

### DIFF
--- a/exercises/practice/secret-handshake/.docs/instructions.append.md
+++ b/exercises/practice/secret-handshake/.docs/instructions.append.md
@@ -1,0 +1,6 @@
+To keep things simple (and to let you focus on the important part of this exercise), your function will receive its inputs as binary strings:
+
+```
+>>> commands("00011")
+["wink", "double blink"]
+```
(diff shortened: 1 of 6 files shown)


BethanyG (changes requested):
Sincere apologies for the delay in getting to this.  This code looks great!  🌟 

I think we'll have to append to the [instructions](https://github.com/exercism/python/blob/main/exercises/practice/secret-handshake/.docs/instructions.md) with something along the lines of:

_"But today, you and your friends have decided to take things a step further, and pass the messages themselves in binary strings."_

Or an adaptation/explanation, since the core instructions are very decimal-number focused right now.

Otherwise, I like this problem variation, and think we should use it.  😄

BethanyG (approved):

PR #2389, opened 2021-03-18
Update to absolute image paths in prep for website

iHiD:
We can't have relative links on the website to images. So this guards against this breaking.


BethanyG (approved):
LGTM!  🌈

PR #2409, opened 2021-04-28
Fix duplicate prerequisite

ErikSchierboom:
There was a duplicate prerequisite for the atbash-cipher exercise, which causes the track to not sync properly.

Closes https://github.com/exercism/website/issues/846


BethanyG (approved):
_If they're not missing, they're duplicated._  🙄    Thanks for the catch!

PR #2416, opened 2021-05-04
🤖 Simplify tests.toml files

ErikSchierboom:
This PR simplifies the `tests.toml` files by removing any `include = true` properties, which are redundant as the default value of this field is `true`.

As the vast majority of the test cases use `include = true`, omitting it makes the `tests.toml` files more readable and make the exceptional cases (`include = false`) easier to identify.

We've updated `configlet` to use this new behavior. See https://github.com/exercism/configlet/issues/186 and https://github.com/exercism/configlet/issues/207 for the corresponding discussion.

## Tracking

https://github.com/exercism/v3-launch/issues/31

### DIFF
--- a/exercises/practice/acronym/.meta/tests.toml
+++ b/exercises/practice/acronym/.meta/tests.toml
@@ -4,36 +4,27 @@
 
 [1e22cceb-c5e4-4562-9afe-aef07ad1eaf4]
 description = "basic"
-include = true
 
 [79ae3889-a5c0-4b01-baf0-232d31180c08]
 description = "lowercase words"
-include = true
 
 [ec7000a7-3931-4a17-890e-33ca2073a548]
 description = "punctuation"
-include = true
 
 [32dd261c-0c92-469a-9c5c-b192e94a63b0]
 description = "all caps word"
-include = true
 
 [ae2ac9fa-a606-4d05-8244-3bcc4659c1d4]
 description = "punctuation without whitespace"
-include = true
 
 [0e4b1e7c-1a6d-48fb-81a7-bf65eb9e69f9]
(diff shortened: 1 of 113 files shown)


BethanyG:
@ErikSchierboom  - configlet is failing all the concept stubs for not having an author in `<concep>/.meta/config.json`.  Do I need to add one in?

BethanyG:
@cmccandless  -- looks like we'll need to update `generate_tests.py` for this, since it is looking for `includes`??

ErikSchierboom:
> configlet is failing all the concept stubs for not having an author in <concep>/.meta/config.json. Do I need to add one in? Can we make it a TODO, or does it have to be an actual user?

You could, for now, use "TODO" as the author username. It does mean that if https://github.com/todo/ ever creates an account at exercism.lol, they will be awarded the reputation 😉

cmccandless:
> @cmccandless  -- looks like we'll need to update `generate_tests.py` for this, since it is looking for `includes=True`??   At least that's what it looks like for `filter_test_cases()` starting at line 201...unless I am reading that incorrectly.

I'll take a look to make sure, but I thought I already corrected for this.

BethanyG:
@ErikSchierboom  -- "fixed" this in #2417 (_Corey and I are co-authors on all the stubs now._), so once that's merged, this can be rebased.  Waiting for approval from @cmccandless  on that.

cmccandless (approved):
I just checked to make sure: our dataclass for tests.toml defaults include to `True`, so this is safe.

https://github.com/exercism/python/blob/main/bin/data.py#L254

ErikSchierboom:
Great!

PR #2418, opened 2021-05-04
Remove fetch-canonical-data-syncer script

ErikSchierboom:
The `fetch-canonical-data-syncer` script is obsolete as canonical data syncing is now built into `configlet`.

See https://github.com/exercism/configlet/issues/124


ee7 (approved):
I've checked that, with this PR,
```
$ git grep 'syncer'
```
produces no output.

cmccandless (approved):
Should be fine. No other tooling currently relies on canonical_data_syncer, so this doesn't break any CI.

CI failures are due to empty author field in exercise configs. This is being addressed elsewhere.

PR #2427, opened 2021-05-12
Gracefully fail with user-friendly error text when unrecognized datac…

cmccandless:
…lass fields are detected

Example:
```bash
$ bin/test_exercises.py acronym
FAIL: dataclass EditorSettings does not have field 'fancy_colors'.
If this field is valid, please add it to the dataclass in data.py.
If adding an object-type field, please create a new dataclass for it.
```


BethanyG (approved):
Nice work!  Thank you for this.  🎉

PR #2452, opened 2021-06-08
Limited 'practices' list to 8 entries or less for practice exercises.

BethanyG:


iHiD (approved):
Wonderful. Thanks :) 

(I've not checked the content, but I'm happy to wave this through if you and configlet are happy!)

neenjaw (approved):

neenjaw:
@iHiD oops! didn't see you approve there! I just saw the notification! 😳

iHiD:
@BethanyG There still seem to be more than 8 exercises for each concept (e.g. see https://exercism.lol/tracks/python/concepts/numbers still has 48). I don't think this is a syncing issue (from checking a few entries but not all 42) but thought I'd let you know here 🙂

SleeplessByte:
@BethanyG as I understood it, it should be at most 8 exercises have `numbers` (and this goes for each concept) in their `practices` list, not each exercise should at most have 8 "items" in their `practices` list.

iHiD:
Yes, what DJ says is correct. If you consider the UI, the problem is how many exercises are shown next to a concept, not how many concepts are shown next to an exercise. 

I think what you've done is also good btw - this will improve the UI for when it shows the concepts practiced at the end of an exercise. It just doesn't solve the initial problem. And the prereqs list is fine.

PR #2496, opened 2021-08-27
Set Test-runner to false for alphametics

BethanyG:


bin/data.py
```diff
@@ -134,6 +134,7 @@ class ExerciseConfig:
     forked_from: str = None
     contributors: List[str] = None
     language_versions: List[str] = None
+    test_runner: bool = True
```
BethanyG, line 137:
```suggestion
```

exercises/practice/alphametics/.meta/config.json
```diff
@@ -26,5 +26,6 @@
     "example": [
       ".meta/example.py"
     ]
-  }
+  },
+  "test_runner": "false"
```
iHiD, line 30: I **think** this needs to be a boolean, not a string.

BethanyG:
```suggestion
  "test_runner": false
```

iHiD:
Thanks :)

PR #2501, opened 2021-08-30
🤖 Remove empty `source_url` property

ee7:
The track CI runs `configlet lint`, which currently exits with a
non-zero exit code if it sees an optional string key that has the value
of the empty string.

This PR therefore removes all such key/value pairs on the track -
they otherwise cause a failing CI check.

---

Tracking: https://github.com/exercism/v3-launch/issues/57


BethanyG (approved):
LGTM!

PR #2525, opened 2021-09-02
Update instructions.md

squeakyboots:
Number 5 is a different header which prevents some markup viewers (Pycharm's for instance) from displaying it correctly as part of the list


BethanyG (approved):
Thanks for the catch and for the PR! 🌟

PR #2531, opened 2021-09-03
Minor change in instructions

CosmoSt4r:


BethanyG (approved):
Thanks for the catch!

PR #2543, opened 2021-09-04
Follow unittest method naming conventions

paiv:
Some tests were not executing since every test method has to be prefixed with `test`


BethanyG (approved):
Thanks for these!

PR #2544, opened 2021-09-04
Added missing 'unexchangeable_value' function to exchange.py

CosmoSt4r:
Added missing function to [Currency Exchange](https://exercism.org/tracks/python/exercises/currency-exchange) exercise's `exchange.py` file.

Function `unexchangeable_value()` exists in tests and instructions but not in `exchange.py` file. I don't think it's intentional since this is a learning exercise.

Also changed single to double triple quotes

### DIFF
--- a/exercises/concept/currency-exchange/exchange.py
+++ b/exercises/concept/currency-exchange/exchange.py
@@ -1,54 +1,67 @@
 def estimate_value(budget, exchange_rate):
-    '''
+    """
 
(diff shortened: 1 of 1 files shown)


BethanyG (approved):
Thank you for this!  It was indeed an omission.  As was not defining the `returns` in the docstrings (_which I just noticed_).  Feel free to add in the `returns` in a follow-on PR if you like.  I will be logging an issue for that shortly.

PR #2549, opened 2021-09-04
Card Games: rephrase the instructions for Task 6

kbuc:
Task 6 requires comparison of the averages of odd- and the even- positioned cards, but the instructions stated odd- and even- valued cards, which could have led to a confusion. Fun fact: the tests worked OK for both interpretations, due to the parities of values corresponding to the parities of indexes.


BethanyG (approved):
Lovely!  Thank you for making this change.  🌟

PR #2552, opened 2021-09-05
Fixed numbers_to_cube example.

marchukov:
numbers_to_cube example in "Working with Lists" used "*3" instead of "**3" while the example
description says that numbers are going to be cubed. Due to this typo the example output was
also wrong and had numbers in the list multiplied by three instead of cubes.

This change fixes the operation to be cube as per the description and also fixes the output.


BethanyG (approved):
Thank you for catching this error!  Much appreciated.  🌟

PR #2562, opened 2021-09-06
Removed prerequisite listings from depricated exercises.

BethanyG:
Removed the prerequisites from all exercises that were flagged  `"deprecated"` in the track `config.json`.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1973,14 +1973,7 @@
         "name": "Accumulate",
         "uuid": "e7351e8e-d3ff-4621-b818-cd55cf05bffd",
         "practices": [],
-        "prerequisites": [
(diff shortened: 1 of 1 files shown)


ee7 (approved):
Context for others: https://github.com/exercism/configlet/pull/415

From [the spec](https://github.com/exercism/docs/blob/14de83e5305b/building/configlet/lint.md#rule-configjson-file-is-valid):

> - The `"exercises.practice[].prerequisites"` value must be an empty
    array if `"exercises.practice[].status"` is equal to `deprecated`

This PR resolves an otherwise-upcoming CI failure - `configlet lint` will enforce this rule in the future.

Issue #2563, opened 2021-09-06
[Practice Exercises] Add Guidance for Exercises that Require Raising Errors or Error handling


BethanyG:
When we transitioned to the current version of exercism, we removed the generic "raising exceptions" notices from all the exercise READMEs.  We felt they were adding confusion and leading students to insert error raising and handling in places where they should not have been inserting it.

For the Python track, we were (and are!) planning `error handling` concept exercises to help students learn how to raise and handle errors properly in their code.  

But those exercises are still in process, and we have **42** [practice exercises](https://exercism.org/docs/building/tracks/practice-exercises) that test for error handling in one way or another.  
**37** of these include `JinJa2` test generation templates.

As a stopgap while we create the `error handling` concept exercises, we need to create practice exercise `.docs/instructions.append.md` files for the `42` exercises that add back a blurb about how to raise an exception.  The original vs proposed language for the `instructions.append.md` is at the bottom, but is open to any edits that make sense and help student understanding.  

 For the 37 `JinJa2` exercises, we will need to modify the test generation templates to test for a meaningful **_exception message_**, and not just the _presence_ of any exception.  We'll then need to regenerate those test files.

For the 5 exercises that remain, we will need to hand-edit the test files as opposed to editing a generation template.

Recommend that a PR for ONE exercise be done as a first pass to make sure the changes do indeed makes sense, then the remainder of them can be done.

*  [Practice Exercise Specs](https://exercism.org/docs/building/tracks/practice-exercises)
*  [Exercism Markdown Specs](https://exercism.org/docs/building/markdown/markdown)
*  [Python Track Test Generator](https://github.com/exercism/python/blob/main/docs/GENERATOR.md)
*  [JinJa2 Tutorial](https://zetcode.com/python/jinja/)

**These are the 42 practice exercises in question:**

[PR 2685](https://github.com/exercism/python/pull/2685)
- [x] [affine-cipher](https://github.com/exercism/python/tree/main/exercises/practice/affine-cipher/) \| [test file](https://github.com/exercism/python/tree/main/exercises/practice/affine-cipher/affine_cipher_test.py) \| [template file](https://github.com/exercism/python/tree/main/exercises/practice/affine-cipher/.meta/template.j2)
[...]

PR #2564, opened 2021-09-06
Add hamming/instructions.append with exception raising details

IsaacG:


exercises/practice/hamming/.docs/instructions.append.md
```diff
@@ -0,0 +1,14 @@
+# Implementation notes: exceptions
+
+This exercise expects the code to raise exceptions in certain circumstances.
+Exceptions are usually raised when an operation cannot be completed for some
+reason and there is no value that can be returned which would be correct.
+Instead, an exception is raised. When raising an exception, the tests expect
```
SleeplessByte, line 6: Doesn't seem to be an "instead" clause as the previous sentence already says there is no clause. `Therefore` might be more applicable.

exercises/practice/hamming/.docs/instructions.append.md
```diff
@@ -0,0 +1,14 @@
+# Implementation notes: exceptions
+
+This exercise expects the code to raise exceptions in certain circumstances.
+Exceptions are usually raised when an operation cannot be completed for some
```
SleeplessByte, line 4: FWIW, general markdown style on Exercism is once sentence per line, not limited to a width. This makes reviews, edits, and diffs easier!

exercises/practice/hamming/.docs/instructions.append.md
```diff
@@ -0,0 +1,14 @@
...
+reason and there is no value that can be returned which would be correct.
```
SleeplessByte, line 5: I think you meant: and there is no applicable value to return?

"which would be correct" is amtrigous. Either the act of no value returning is correct, or returning a value would be correct, or a correct value cannot be returned.

exercises/practice/hamming/.docs/instructions.append.md
```diff
@@ -0,0 +1,14 @@
...
+Instead, an exception is raised. When raising an exception, the tests expect
+that the exception class in instantiated and contains a message.
+
+```python
+# Avoid this. There is no message and the tests will fail:
+raise ValueError
+# Do this. Provide a clear message with as much detail as possible:
```
SleeplessByte, line 12:
```suggestion
# Do this. Provide a clear and actionable message:
```

I don't think we always want to go into detail, but whatever we do, there must be "just" enough information to take an action. A large assssss error message which you can't solve isn't as good as a short, solvable, error. Right?

SleeplessByte (commented):
My 2 cents!

exercises/practice/hamming/.docs/instructions.append.md
IsaacG: Operation cannot be performed. Function usually returns a value. Any value returned would be incorrect. There exists no value where returning that value would be correct.

Maybe,
> Exceptions are usually raised when an operation cannot be completed for some reason and therefore there is no value to return. Any value returned would be an incorrect value.

exercises/practice/hamming/.docs/instructions.append.md
IsaacG: Changed.

exercises/practice/hamming/.docs/instructions.append.md
IsaacG: Changed.

exercises/practice/hamming/.docs/instructions.append.md
SleeplessByte: Yeah much better!

IsaacG:
Related to https://github.com/exercism/python/issues/2563 (claimed)

BethanyG:
@IsaacG  - Thank you for submitting this, but I am going to close this at this time.  Given our (lengthy) slack discussions, the number of exercises we may now have to change, and the wording in #2563 (which needs to be discussed), I think we're not read to go ahead with changes.

Issue #2565, opened 2021-09-08
Task 2 of the Meltdown exercise should be rephrased.


kbuc:
Our reactor will Meltdown!

The intervals:

```
green -> 80-100% efficiency
orange -> 60-79% efficiency
red -> 30-59% efficiency
black -> <30% efficient
```

are not properly closed. In which interval does the 79.5% fall? And 59.73%? 

The exemplar solution does suggest that `voltage`, `current` and `theoretical_max_power` being integers (but why?); 
this fact however does  **not** make the `percentage_range` automatically integer as well. 
 
I see two options:

1. Casting the `percentage_range` to integer (but why?), and staying in the integer scope with the existing intervals.
   The integer percentage should then also be mentioned in the instructions.

2. Keep floating `percentage_range` and include/exclude left and right interval limits with a proper use of `<` and `<=`. 
   
Personally, I would go with the second suggestion, it's much more realistic and let the students also think about the limits. 
In both cases, instructions should be updated.  

*The border cases should also be verified in the tests!*

A proposal for rephrasing the intervals:
```
green -> efficiency of 80% or more,
orange -> efficiency of less than 80% but at least 60%, 
red -> efficiency below 60%, but still 30% or more,
black ->  less than 30% efficient.
```

I'm happy to do a PR after reaching the agreement between options 1 and 2.

BethanyG:
@kbuc - excellent catch!  Thank you for logging this!

I am inclined at first glance to go with `option 2`.  [BUT](https://0.30000000000000004.com/).  😄   So if we are going to do `option 2`, I would want the instructions to be really clear with students, so we don't get solutions that use round or floor or some other funky mechanic.  There are always more decimal places, so a limit needs to be a limit of precision as well.

I'd also suggest that the revised intervals be written as floats (with the expected precision), or otherwise nudge students toward the option that they will want to use `=>` and `>` appropriately, to account for the infinite amount of numbers between the intervals.

And I would be happy if you want to take this on!  In fact, there is also contention around some of the other tasks in the exercise (see #2535 and #2566 ), which could use some fresh eyes - if you wouldn't mind giving it a full once-over?

kbuc:
Since the focus of the exercise is on conditionals, I would not complicate with floating point peculiarities. As long as the student uses proper math (normal division, comparison with bounds included or not), the tests should pass. 

[...]

PR #2566, opened 2021-09-08
Fix meltdown mitigation task 1

### LINKED ISSUE #2535: Task 1 of Meltdown Mitigation is misleading
One of the most common blunders a novice can make is to use if-elif-else blocks instead of using a single return statement.

For instance

```python
if a:
    return True
elif b:
    return True
else:
    return False
```

should simply be written

```python
return a or b
```

The introduction sets up using `if:elif:else:` blocks. I fear that a novice will blindly fall into the trap of solving task 1 using an `if`.

```python
def is_criticality_balanced(temperature, neutrons_emitted):
    if temperature < 800 and neutrons_emitted > 500 and temperature * neutrons_emitted < 500_000:
        return True
    else:
        return False
```

or something even worse, when it should be solved with a single return statement.

```python
def is_criticality_balanced(temperature, neutrons_emitted):
    return temperature < 800 and neutrons_emitted > 500 and temperature * neutrons_emitted < 500_000
```

punkstarman:
fixes #2535

### DIFF
--- a/exercises/concept/meltdown-mitigation/.docs/introduction.md
+++ b/exercises/concept/meltdown-mitigation/.docs/introduction.md
@@ -48,9 +48,9 @@ if x > y:
 elif y > z:
     print("y is greater than x and z")
 else:
-    print("z is great than x and y")
+    print("z is greater than x and y")
 ...
->>> z is great than x and y
+>>> z is greater than x and y
 ```
 
 [Boolean operations][boolean operations] and [comparisons][comparisons] can be combined with conditionals for more complex testing:
@@ -72,6 +72,22 @@ else:
 
 >>> classic_fizzbuzz(13)
 '13'
+```
+
+Bear in mind that in the special case where a function returns a boolean, it should not use an `if`.
+Rather it should return the condition directly.
+
+```python
+# DO NOT use `if` and return `True`
+>>> def is_leap_year(year):
+        if (year % 4 == 0 and year % 100 != 0) or year % 400 == 0:
+            return True
+        else:
+            return False
+
+# instead return the condition directly
+>>> def is_leap_year(year):
+        return (year % 4 == 0 and year % 100 != 0) or year % 400 == 0
 ```
 
 [if statement]: https://docs.python.org/3/reference/compound_stmts.html#the-if-statement
--- a/exercises/concept/meltdown-mitigation/.meta/exemplar.py
+++ b/exercises/concept/meltdown-mitigation/.meta/exemplar.py
@@ -60,7 +60,7 @@ def fail_safe(temperature, neutrons_produced_per_second, threshold):
     - `temperature * neutron per second` is not in the above-stated ranges ==  'DANGER'
     """
     output = temperature * neutrons_produced_per_second
-    operational_percentage = int(output / threshold) * 100
+    operational_percentage = int((output / threshold) * 100)
 
     if operational_percentage < 40:
         return 'LOW'
--- a/exercises/concept/meltdown-mitigation/.meta/exemplar.py
+++ b/exercises/concept/meltdown-mitigation/.meta/exemplar.py
@@ -11,12 +11,8 @@ def is_criticality_balanced(temperature, neutrons_emitted):
     - The product of temperature and neutrons emitted per second less than 500000.
     """
     output = temperature * neutrons_emitted
-    balanced = False
 
-    if (temperature < 800 and neutrons_emitted > 500) and output < 500000:
-        balanced = True
-
-    return balanced
+    return temperature < 800 and neutrons_emitted > 500 and output < 500000
 
 
 def reactor_efficiency(voltage, current, theoretical_max_power):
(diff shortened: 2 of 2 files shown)


exercises/concept/meltdown-mitigation/.meta/exemplar.py
```diff
@@ -67,14 +60,11 @@ def fail_safe(temperature, neutrons_produced_per_second, threshold):
     - `temperature * neutron per second` is not in the above-stated ranges ==  'DANGER'
     """
     output = temperature * neutrons_produced_per_second
-    operational_percentage = int((output / threshold) * 100)
-    safety_range = 'UNKNOWN'
+    operational_percentage = int(output / threshold) * 100
```
mukeshgurpude, line 63: If you haven't found why tests failed?
> `int(output / threshold) * 100`
Above expression returns 0, when output is less than threshhold, which is not expected behavior.
This must be changed to `int((output / threshold) * 100)`, to behave correctly...

BethanyG, line 63: This is causing test failure, and needs to be backed out for the time being.

```suggestion
    operational_percentage = int(output / threshold * 100)
```

exercises/concept/meltdown-mitigation/.meta/exemplar.py
```diff
@@ -40,18 +36,15 @@ def reactor_efficiency(voltage, current, theoretical_max_power):
...
-        efficiency_level = 'orange'
+        return 'orange'
     elif 30 <= percentage_range <= 59:
-        efficiency_level = 'red'
+        return 'red'
     else:
-        efficiency_level = 'black'
-
-    return efficiency_level
+        return 'black'
```
BethanyG, line 47: This task is getting revised via #2565, so I think we should back out these changes.

exercises/concept/meltdown-mitigation/.docs/introduction.md
```diff
@@ -74,6 +74,22 @@ else:
...
+        if (year % 4 == 0 and year % 100 != 0) or year % 400 == 0:
+            return True
+        else:
+            return False
+
+# instead return the condition directly
+>>> def is_leap_year(year):
+        return (year % 4 == 0 and year % 100 != 0) or year % 400 == 0
+```
+
```
BethanyG, line 92: While this is excellent information to discuss, I don't think this exercise (_which is focused on conditionals_) is the place to do it.  This note sets an exception/rule about Booleans without going into the nuance that one can return an expression directly without assigning to an intermediary state variable -- which isn't just a rule around conditions and Booleans. I think it's better left to a mentor discussion/notes or explored in detail in an exercise on structuring functions/handling returns.

exercises/concept/meltdown-mitigation/.docs/introduction.md
```diff
@@ -48,9 +48,9 @@ if x > y:
 elif y > z:
     print("y is greater than x and z")
 else:
-    print("z is great than x and y")
+    print("z is greater than x and y")
 ...
->>> z is great than x and y
+>>> z is greater than x and y
```
BethanyG, line 53: Thanks for catching this!

exercises/concept/meltdown-mitigation/.meta/exemplar.py
```diff
@@ -11,12 +11,8 @@ def is_criticality_balanced(temperature, neutrons_emitted):
     - The product of temperature and neutrons emitted per second less than 500000.
     """
     output = temperature * neutrons_emitted
-    balanced = False
 
-    if (temperature < 800 and neutrons_emitted > 500) and output < 500000:
-        balanced = True
-
-    return balanced
+    return temperature < 800 and neutrons_emitted > 500 and output < 500000
```
BethanyG, line 15: As stated in the issue, this exercise is about `conditionals`, and every task should use or practice them - as contrived as that feels.  This task in the exemplar now displays only the use of Booleans.  If this task cannot be re-written to use/display the use of `conditionals`, then another task should be made or this task should be omitted from the exercise.

BethanyG (changes requested):
In addition to the in-line comments (which did not trigger "request change" for some reason), as @mukeshgurpude has stated, the CI / tests are failing due to an error on line 63.

exercises/concept/meltdown-mitigation/.docs/introduction.md
punkstarman: I strongly disagree, unless someone can find a way to fix Task 1 so that it is not misleading (see [another discussion](https://github.com/exercism/python/pull/2566#discussion_r704648474)).

The learning exercise on Booleans (Ghost Gobble Arcade Game) already has students [returning expressions directly](https://github.com/exercism/python/blob/main/exercises/concept/ghost-gobble-arcade-game/.meta/exemplar.py#L43), and this concept is explained in the Basics exercise [Guido's Gorgeous Lasagna](https://github.com/exercism/python/blob/main/exercises/concept/guidos-gorgeous-lasagna/.docs/introduction.md).

Also, [knowing when not to use a tool is part of mastering that tool](https://twitter.com/kelseyhightower/status/963428093292457984?s=20).

exercises/concept/meltdown-mitigation/.meta/exemplar.py
punkstarman: As it is, the task is misleading, so 
> another task should be made or this task should be omitted from the exercise.
but I do not have the inspiration for this at the moment.

BethanyG:
Closing since we cannot come to agreement on task 1, and because there have been changes to this exercise via [PR2568](https://github.com/exercism/python/pull/2568) that have made substantial changes.

PR #2568, opened 2021-09-10
Meltdown exercise change according to issue #2565

kbuc:
Changes:
- Change Task 2 intervals to cover non-integer comparison.
- Rewrite tests to test more points, especially the border cases.
- Reduce number of tests of only one per function, but test many points
  (asserts in loop) in each test.
- Remove hints of integer input parameters (they can be any numeric type).

### DIFF
--- a/exercises/concept/meltdown-mitigation/.docs/instructions.md
+++ b/exercises/concept/meltdown-mitigation/.docs/instructions.md
@@ -30,13 +30,15 @@ True
 Once the reactor has started producing power its efficiency needs to be determined.
 Efficiency can be grouped into 4 bands:
 
-1. green -> 80-100% efficiency
-2. orange -> 60-79% efficiency
-3. red -> 30-59% efficiency
-4. black -> <30% efficient
+1. green -> efficiency of 80% or more,
+2. orange -> efficiency of less than 80% but at least 60%,
+3. red -> efficiency below 60%, but still 30% or more,
+4. black ->  less than 30% efficient.
 
-These percentage ranges are calculated as `(generated_power/theoretical_max_power)*100`
-where `generated_power = voltage * current`
+The percentage value can be calculated as `(generated_power/theoretical_max_power)*100`
+where `generated_power = voltage * current`.
+Note that the percentage value is usually not an integer number and take care of a
+proper use of the `<` and `<=` comparisons.
 
 Implement the function `reactor_efficiency()`, with three parameters: `voltage`,
 `current`, and `theoretical_max_power`.
--- a/exercises/concept/meltdown-mitigation/.docs/instructions.md
+++ b/exercises/concept/meltdown-mitigation/.docs/instructions.md
@@ -37,7 +37,7 @@ Efficiency can be grouped into 4 bands:
 
 The percentage value can be calculated as `(generated_power/theoretical_max_power)*100`
 where `generated_power = voltage * current`.
-Note that the percentage value is usually not an integer number and take care of a
+Note that the percentage value is usually not an integer number, so make sure to consider the
 proper use of the `<` and `<=` comparisons.
 
 Implement the function `reactor_efficiency()`, with three parameters: `voltage`,
--- a/exercises/concept/meltdown-mitigation/.meta/exemplar.py
+++ b/exercises/concept/meltdown-mitigation/.meta/exemplar.py
@@ -1,8 +1,8 @@
 def is_criticality_balanced(temperature, neutrons_emitted):
     """Verify criticality is balanced.
 
-    :param temperature: temperature
-    :param neutrons_emitted: number of neutrons emitted per second
+    :param temperature: Union[int,float] temperature
+    :param neutrons_emitted: Union[int,float] number of neutrons emitted per second
     :return:  boolean True if conditions met, False if not
 
     A reactor is said to be critical if it satisfies the following conditions:
@@ -22,9 +22,9 @@ def is_criticality_balanced(temperature, neutrons_emitted):
 def reactor_efficiency(voltage, current, theoretical_max_power):
     """Assess reactor efficiency zone.
 
-    :param voltage: voltage
-    :param current: current
-    :param theoretical_max_power: power that correspods to a 100% efficiency
+    :param voltage: Union[int,float] voltage
+    :param current: Union[int,float] current
+    :param theoretical_max_power: Union[int,float] power that corresponds to a 100% efficiency
     :return: str one of 'green', 'orange', 'red', or 'black'
 
     Efficiency can be grouped into 4 bands:
@@ -57,9 +57,9 @@ def reactor_efficiency(voltage, current, theoretical_max_power):
 def fail_safe(temperature, neutrons_produced_per_second, threshold):
     """Assess and return safety range.
 
-    :param temperature:
-    :param neutrons_produced_per_second:
-    :param threshold:
+    :param temperature: Union[int,float]
+    :param neutrons_produced_per_second: Union[int,float]
+    :param threshold: Union[int,float]
     :return: str one of: 'LOW', 'NORMAL', 'DANGER'
 
     - `temperature * neutrons per second` < 40% of `threshold` == 'LOW'
--- a/exercises/concept/meltdown-mitigation/.meta/exemplar.py
+++ b/exercises/concept/meltdown-mitigation/.meta/exemplar.py
@@ -69,7 +69,6 @@ def fail_safe(temperature, neutrons_produced_per_second, threshold):
     output = temperature * neutrons_produced_per_second
     operational_percentage = (output / threshold) * 100
     safety_range = 'UNKNOWN'
-    print(operational_percentage)
 
     if operational_percentage < 40:
         safety_range = 'LOW'
--- a/exercises/concept/meltdown-mitigation/.meta/exemplar.py
+++ b/exercises/concept/meltdown-mitigation/.meta/exemplar.py
@@ -1,8 +1,8 @@
 def is_criticality_balanced(temperature, neutrons_emitted):
     """Verify criticality is balanced.
 
-    :param temperature: Union[int,float] temperature
-    :param neutrons_emitted: Union[int,float] number of neutrons emitted per second
+    :param temperature: temperature value (integer or float)
+    :param neutrons_emitted: number of neutrons emitted per second (integer or float)
     :return:  boolean True if conditions met, False if not
 
     A reactor is said to be critical if it satisfies the following conditions:
@@ -22,9 +22,9 @@ def is_criticality_balanced(temperature, neutrons_emitted):
 def reactor_efficiency(voltage, current, theoretical_max_power):
     """Assess reactor efficiency zone.
 
-    :param voltage: Union[int,float] voltage
-    :param current: Union[int,float] current
-    :param theoretical_max_power: Union[int,float] power that corresponds to a 100% efficiency
+    :param voltage: voltage value (integer or float)
+    :param current: current value (integer or float)
+    :param theoretical_max_power: power that corresponds to a 100% efficiency (integer or float)
     :return: str one of 'green', 'orange', 'red', or 'black'
 
     Efficiency can be grouped into 4 bands:
@@ -57,9 +57,9 @@ def reactor_efficiency(voltage, current, theoretical_max_power):
 def fail_safe(temperature, neutrons_produced_per_second, threshold):
     """Assess and return safety range.
 
-    :param temperature: Union[int,float]
-    :param neutrons_produced_per_second: Union[int,float]
-    :param threshold: Union[int,float]
+    :param temperature: value of the temperature (integer or float)
+    :param neutrons_produced_per_second: neutron flux (integer or float)
+    :param threshold: threshold (integer or float)
     :return: str one of: 'LOW', 'NORMAL', 'DANGER'
 
     - `temperature * neutrons per second` < 40% of `threshold` == 'LOW'
--- a/exercises/concept/meltdown-mitigation/.meta/exemplar.py
+++ b/exercises/concept/meltdown-mitigation/.meta/exemplar.py
@@ -1,14 +1,14 @@
 def is_criticality_balanced(temperature, neutrons_emitted):
     """Verify criticality is balanced.
 
-    :param temperature: int
-    :param neutrons_emitted: int
+    :param temperature: temperature
+    :param neutrons_emitted: number of neutrons emitted per second
     :return:  boolean True if conditions met, False if not
 
     A reactor is said to be critical if it satisfies the following conditions:
-    - The temperature less than 800.
-    - The number of neutrons emitted per second greater than 500.
-    - The product of temperature and neutrons emitted per second less than 500000.
+    - The temperature is less than 800.
+    - The number of neutrons emitted per second is greater than 500.
+    - The product of temperature and neutrons emitted per second is less than 500000.
     """
     output = temperature * neutrons_emitted
     balanced = False
@@ -22,19 +22,19 @@ def is_criticality_balanced(temperature, neutrons_emitted):
 def reactor_efficiency(voltage, current, theoretical_max_power):
     """Assess reactor efficiency zone.
 
-    :param voltage: int
-    :param current: int
-    :param theoretical_max_power: int
+    :param voltage: voltage
+    :param current: current
+    :param theoretical_max_power: power that correspods to a 100% efficiency
     :return: str one of 'green', 'orange', 'red', or 'black'
 
     Efficiency can be grouped into 4 bands:
 
-    1. green  ->   80-100% efficiency
-    2. orange ->   60-79% efficiency
-    3. red    ->   30-59% efficiency
-    4. black  ->   <30% efficient
+    1. green -> efficiency of 80% or more,
+    2. orange -> efficiency of less than 80% but at least 60%,
+    3. red -> efficiency below 60%, but still 30% or more,
+    4. black ->  less than 30% efficient.
 
-    These percentage ranges are calculated as
+    The percentage value is calculated as
     (generated power/ theoretical max power)*100
     where generated power = voltage * current
     """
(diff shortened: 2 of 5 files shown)


exercises/concept/meltdown-mitigation/.docs/instructions.md
```diff
@@ -30,13 +30,15 @@ True
...
-where `generated_power = voltage * current`
+1. green -> efficiency of 80% or more,
+2. orange -> efficiency of less than 80% but at least 60%,
+3. red -> efficiency below 60%, but still 30% or more,
+4. black ->  less than 30% efficient.
+
+The percentage value can be calculated as `(generated_power/theoretical_max_power)*100`
+where `generated_power = voltage * current`.
+Note that the percentage value is usually not an integer number and take care of a
+proper use of the `<` and `<=` comparisons.
```
BethanyG, line 41:
```suggestion
Note that the percentage value is usually not an integer number, so make sure to consider the
proper use of the `<` and `<=` comparisons.
```

exercises/concept/meltdown-mitigation/conditionals_test.py
```diff
@@ -7,164 +7,70 @@
...
+            (625, 800, False), (625.99, 800, False), (625.01, 799.99, False),
+            (799.99, 500.01, True), (624.99, 799.99, True),
+            (500, 1000, False), (500.01, 1000, False), (499.99, 1000, True),
+            ):
+            got = is_criticality_balanced(temperature=temperature, neutrons_emitted=neutrons_emitted)
+            self.assertEqual(
+                got,
+                expected,
+                msg=f"Expected {expected} but returned {got} with T={temperature} and neutrinos={neutrons_emitted}"
+            )
```
BethanyG, line 27: I love that you parameterized these, but would prefer that you use `unittest.subtest` as opposed to a plain loop.  

That way, individual points can count as individual tests when they fail, and we can count/label each as a variant.  See [this exercise](https://github.com/exercism/python/blob/main/exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py) as an example. And a screengrab of a failure state in the UI with the variants listed:

BethanyG (changes requested):
@kbuc  -- These are excellent changes.  Thank you.  😄   I have one language suggestion (which is optional) and a request about tests/variations.  Let me know if you have any questions or issues.

exercises/concept/meltdown-mitigation/conditionals_test.py
BethanyG: Yes -- that peculiarity you mentioned is .. really a major pain for subtests.  😅 

For the runner/UI, we apprehend the "success" and re-write the parent outcome as a failure if any `subtest` is a failure.  But if someone is running Pytest locally, it can be confusing -- even with `pytest-subtest`.  

Still thinking on how to clarify that for CLI students without packaging up our own PyTest plugin for download and distribution (_which might be an option_).  But I am also of the opinion that anyone who has figured out the CLI will only be stopped by the subtest mis-match once or twice.  And I thought having shorter, cleaner, and clearer test files was worth it.

BethanyG:
Oh!  And I almost forgot (_which is terrible_) -- please add a `contributors` key to the config for this exercise, and then add yourself to the array.  😄  [exercise `config.json`](https://github.com/exercism/python/blob/main/exercises/concept/meltdown-mitigation/.meta/config.json), so that you get credit for all the work you've done on this re-write.  [Here](https://github.com/exercism/docs/blob/main/building/tracks/concept-exercises.md#file-metaconfigjson) is the spec for that `config.json` file.

exercises/concept/meltdown-mitigation/conditionals_test.py
kbuc: Yes, I got quickly used to the funny output. 
It's just annoying that running `pytest` on a default empty stub (where all functions just `pass`), one gets: 
```
=== 41 failed, 3 passed in 0.76s ====
```
3 passed tests before even writing a single chracter of code. 🙄

BethanyG: I know!  It makes me _insane_.  Evidently, it's not an "easy" problem to fix?  Once I saw this ongoing [debate/discussion set](https://github.com/pytest-dev/pytest-subtests/issues/11), I .... _backed away slowly_.  LOL

BethanyG: This next review will take me a few - I need to run the new test file through the runner and make sure the failures get reported correctly.  I _may_ need to change `point` to `variant`, but I need to double check the runner code first.  😄

exercises/concept/meltdown-mitigation/conditionals.py
```diff
@@ -1,8 +1,11 @@
+""" Meltdown Mitigation exercise """
+
+
 def is_criticality_balanced(temperature, neutrons_emitted):
     """Verify criticality is balanced.
 
-    :param temperature: int
```
IsaacG, line 4: Format:  `:param <name>: <type> [- optional text]`

Don't lose the type here.

exercises/concept/meltdown-mitigation/.docs/instructions.md
```diff
@@ -30,13 +30,15 @@ True
...
 
-1. green -> 80-100% efficiency
-2. orange -> 60-79% efficiency
-3. red -> 30-59% efficiency
-4. black -> <30% efficient
-
-These percentage ranges are calculated as `(generated_power/theoretical_max_power)*100`
-where `generated_power = voltage * current`
+1. green -> efficiency of 80% or more,
+2. orange -> efficiency of less than 80% but at least 60%,
```
IsaacG, line 34: It's generally easier for people to parse ranges from low to high. "efficiency of at least 60% but less than 80%"

exercises/concept/meltdown-mitigation/.meta/exemplar.py
```diff
@@ -67,8 +67,9 @@ def fail_safe(temperature, neutrons_produced_per_second, threshold):
     - `temperature * neutron per second` is not in the above-stated ranges ==  'DANGER'
     """
     output = temperature * neutrons_produced_per_second
-    operational_percentage = int((output / threshold) * 100)
+    operational_percentage = (output / threshold) * 100
     safety_range = 'UNKNOWN'
+    print(operational_percentage)
```
IsaacG, line 72: Generally non-CLI code should not contain any `print()` statements.

exercises/concept/meltdown-mitigation/conditionals.py
```diff
@@ -1,8 +1,11 @@
+""" Meltdown Mitigation exercise """
```
IsaacG, line 1: Docstrings should not be space padded. The docstring summary line should terminate in a period.

exercises/concept/meltdown-mitigation/conditionals.py
```diff
@@ -17,19 +20,19 @@ def is_criticality_balanced(temperature, neutrons_emitted):
 def reactor_efficiency(voltage, current, theoretical_max_power):
     """Assess reactor efficiency zone.
 
-    :param voltage: int
-    :param current: int
-    :param theoretical_max_power: int
+    :param voltage: voltage
+    :param current: current
+    :param theoretical_max_power: power that correspods to a 100% efficiency
```
IsaacG, line 25: s/correspods/corresponds/

exercises/concept/meltdown-mitigation/conditionals_test.py
```diff
@@ -7,164 +7,81 @@
...
-    def test_is_criticality_balanced_set1(self):
-        self.assertTrue(is_criticality_balanced(
-            temperature=750, neutrons_emitted=650), msg="Expected True but returned False")
-
-
-    @pytest.mark.task(taskno=1)
-    def test_is_criticality_balanced_set2(self):
-        self.assertTrue(is_criticality_balanced(
-            temperature=799, neutrons_emitted=501), msg="Expected True but returned False")
+    """ Test cases for Meltown mitigation exercise. """
```
IsaacG, line 10: Docstrings should not be space padded.

exercises/concept/meltdown-mitigation/conditionals_test.py
```diff
@@ -7,164 +7,81 @@
...
 
+    # Test cases for is_criticality_balanced()
 
+    # Testing border cases around typical points T, n == (800, 500), (625, 800), (500, 1000)
```
IsaacG, line 14: This comment would be even better as a docstring.

exercises/concept/meltdown-mitigation/conditionals_test.py
```diff
@@ -7,164 +7,81 @@
...
-    # Need to verify if f-string based errors allowed
+    def test_is_criticality_balanced(self):
+        point_id = 0
+        for temperature, neutrons_emitted, expected in (
+            (750, 650, True), (799, 501, True), (500, 600, True), (1000, 800, False),
+            (800, 500, False), (800, 500.01, False), (799.99, 500, False), (500.01, 999.99, False),
+            (625, 800, False), (625.99, 800, False), (625.01, 799.99, False),
+            (799.99, 500.01, True), (624.99, 799.99, True),
+            (500, 1000, False), (500.01, 1000, False), (499.99, 1000, True),
+            ):
```
IsaacG, line 24: This line should be indented one level less.

exercises/concept/meltdown-mitigation/conditionals_test.py
```diff
@@ -7,164 +7,81 @@
...
-            temperature=1000, neutrons_produced_per_second=25, threshold=5000)
-        self.assertEqual(
-            test_return, 'DANGER', msg=f"Expected DANGER but returned {test_return}"
-        )
+    def test_fail_safe(self):
+        point_id = 0
+        temperature = 10
+        threshold = 10000
+        for neutrons_produced_per_second, expected in (
+            (399, 'LOW'), (300, 'LOW'), (1, 'LOW'), (0, 'LOW'),
```
IsaacG, line 72: A `dict[str, list[int]]` might be a better data structure for this data. That allows you to specify the level once vs having to repeat it multiple times.

IsaacG (changes requested):

exercises/concept/meltdown-mitigation/conditionals.py
kbuc: Should we stay with `int` or generalize to `float` ?

exercises/concept/meltdown-mitigation/.meta/exemplar.py
kbuc: This was a debugging leftover... Thanks,

exercises/concept/meltdown-mitigation/conditionals.py
kbuc: Keep the `int` or go to `float`?

exercises/concept/meltdown-mitigation/conditionals.py
IsaacG: It doesn't matter much. I think `int` is cleaner and simpler and that'd be my preference. Plus, if it isn't broken, don't fix it. `int` works just fine.

If you want to get fancy, you can go with [numbers.Real](https://docs.python.org/3/library/numbers.html#numbers.Real) :)

exercises/concept/meltdown-mitigation/conditionals_test.py
kbuc: That would include one more `for` loop (and one more indent level). I'm not sure if it improves redability. 
Optionally, we could make a separate test for each level.

exercises/concept/meltdown-mitigation/conditionals.py
kbuc: Let's stay with `int` then, also tests feed ints. It's just conceptually strange for me to have quantities like temperature, voltage, current,... integers.

IsaacG: Given that digital sensors produce discrete levels, `int` makes a lot of sense. Getting a float from a sensor involves converting that discrete level.

exercises/concept/meltdown-mitigation/conditionals_test.py
IsaacG: What I tell students on Protein Transition is that you can generate your data, too. 

```
vals = {
  "LOW": [1, 2, ...],
  "MED": [...],
}
cases = [(level, num) for level nums in vals.items() for num in nums]
```

BethanyG: These test cases need to be re-written anyways, as they aren't passing what's needed to the test runner.  I just haven't gotten to figuring out how, so I would leave this alone right now.

BethanyG: @kbuc -  My sincere apologies.  This took a while, and now the review interface won't let me do a direct suggest.  I guess there are just too many comments already.  So I am going to have to do an ugly code post.  I tested these cases, and they do work with the runner, and seem to be doing the job.  

Generally,  I moved `test data` for the functions to a variable (_seemed easier to read and maintain that way_), and used `enumerate()` for variation counting.  I also moved the messages into their own variable for readability. YMMV.  Yes - my runner code is a bit brittle -- it doesn't know to summarize subtests unless you use the word "variant".  Long-term that won't be the best strategy, but it works for now.

Let me know if you have any questions or issues.  You can also close this PR and re-open the changes in a different (linked) PR if all the comments have gotten to be too much.  Up to you.  Here are my test case changes:


```python
import unittest
import pytest
from conditionals import (is_criticality_balanced,
                          reactor_efficiency,
                          fail_safe)


class TestConditionals(unittest.TestCase):
    """Test cases for Meltdown mitigation exercise.
    """

    @pytest.mark.task(taskno=1)
    def test_is_criticality_balanced(self):
        """Testing border cases around typical points.

        T, n == (800, 500), (625, 800), (500, 1000), etc.

        """

        test_data = ((750, 650, True), (799, 501, True), (500, 600, True),
                     (1000, 800, False), (800, 500, False), (800, 500.01, False),
                     (799.99, 500, False), (500.01, 999.99, False), (625, 800, False),
                     (625.99, 800, False), (625.01, 799.99, False), (799.99, 500.01, True),
                     (624.99, 799.99, True), (500, 1000, False), (500.01, 1000, False),
                     (499.99, 1000, True))

        for variant, data in enumerate(test_data, start=1):
            temperature, neutrons_emitted, expected = data
            with self.subTest(f"variation #{variant}", temperature=temperature, neutrons_emitted=neutrons_emitted, expected=expected):
                got = is_criticality_balanced(temperature, neutrons_emitted)
                msg=f"Expected {expected} but returned {got} with T={temperature} and neutrinos={neutrons_emitted}"
                self.assertEqual(got, expected, msg)


    @pytest.mark.task(taskno=2)
    def test_reactor_efficiency(self):
        voltage = 10
        theoretical_max_power = 10000

        # The numbers are chosen so that current == 10 x percentage
        test_data = ((1000, 'green'), (999, 'green'), (800, 'green'),
                    (799, 'orange'), (700, 'orange'), (600, 'orange'),
                    (599, 'red'), (560, 'red'), (400, 'red'), (300, 'red'),
                    (299, 'black'), (200, 'black'), (0, 'black'))

        for variant, data in enumerate(test_data, start=1):
            current, expected = data
            with self.subTest(f"variation #{variant}", voltage=voltage, current=current,
                              theoretical_max_power=theoretical_max_power, expected=expected):
                got = reactor_efficiency(voltage, current, theoretical_max_power)
                msg=f"Expected {expected} but returned {got} with voltage={voltage}, current={current}, " \
                    f"max_pow={theoretical_max_power}"
                self.assertEqual(got, expected, msg)



    @pytest.mark.task(taskno=3)
    def test_fail_safe(self):
        temperature = 10
        threshold = 10000
        test_data = ((399, 'LOW'), (300, 'LOW'),(1, 'LOW'),
                    (0, 'LOW'), (901, 'NORMAL'), (1000, 'NORMAL'),
                    (1099, 'NORMAL'), (899, 'DANGER'), (700, 'DANGER'),
                    (400, 'DANGER'), (1101, 'DANGER'), (1200, 'DANGER'))

        for variant, data in enumerate(test_data, start=1):
            neutrons_produced_per_second, expected = data

            with self.subTest(f"variation #{variant}", temperature=temperature,
                              neutrons_produced_per_second=neutrons_produced_per_second,
                              threshold=threshold, expected=expected):

                got = fail_safe(temperature, neutrons_produced_per_second, threshold)
                msg = f"Expected {expected} but returned {got} with T={temperature}, " \
                      f"neutrons={neutrons_produced_per_second}, threshold={threshold}"
                self.assertEqual(got, expected, msg) 
```

kbuc: @BethanyG Thank you very much for all your engagement. I have copied your code to the tests and also changed the "type hinting" in the docstring of the stub and exemplar.

exercises/concept/meltdown-mitigation/.meta/exemplar.py
```diff
@@ -1,14 +1,14 @@
 def is_criticality_balanced(temperature, neutrons_emitted):
     """Verify criticality is balanced.
 
-    :param temperature: int
-    :param neutrons_emitted: int
+    :param temperature: Union[int,float] temperature
```
BethanyG, line 4: Please don't put type hints or items from `typing` in concept exercises.  Since they are an optional module and much further down the concept tree, they shouldn't be referred to or used at this time.

BethanyG:
@IsaacG  -- Please don't recommend anything from `typing` in concept exercises or concept exercise docstrings that don't have `typing` as a prerequisite. We don't want to confuse students or require an optional module for a concept exercise.  Thanks!

exercises/concept/meltdown-mitigation/.meta/exemplar.py
kbuc: It's inside the string, so it's not part of the code at all, no module needed. It was just one of (to me) more readable ways to say that the input could be int or float and keep it compact.

kbuc: We could also write `:param temperature: int or float value of the temperature`. Does that look better?

BethanyG: Yes.  I don't want MyPy or `typing` referrals -- even in docstrings.  I just think it implies to students that they somehow need to use those things.  They don't, and it creates confusion.  It is ok to have verbose docstrings for learning.

BethanyG: I appreciate wanting to be less verbose, but we also need to be really really explicit.

BethanyG:
Bumping this to `large` size given the back and forth with test cases and language.  @kbuc -- many thanks for your patience with the process!

BethanyG (approved):

Issue #2438, opened 2021-05-28
[Pretty-Leaflet] Rewrite Exercise to Omit the Use of Classes


BethanyG:
The `Pretty-Leaflet` concept exercise (`string-formatting` concept) needs to be re-written to not require the creation of/ working with `classes` as part of its solution.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [Contributing to Exercism](https://exercism.org/contributing) |  [Help](https://exercism.org/docs/building/product)
- [What are those Weird Task Tags about?](https://exercism.org/docs/building/tasks)
- [Building Language Tracks: An Overview](https://exercism.org/docs/building/tracks) 
- [What are Concepts?](https://exercism.org/docs/building/tracks/concepts)
- [Concept Exercise Specifications](https://exercism.org/docs/building/product/concept-exercises)
- [Exercism Formatting and Style Guide](https://exercism.org/docs/building/markdown/style-guide)
- [Exercism Markdown Specification](https://exercism.org/docs/building/markdown/markdown)

-------

`Pretty-Leaflet`  was intended to appear "early" in the Python concept exercise tree and not require a student to know how to create and work with `classes` or with `higher-order functions`.  It should appear right after students learn the core basics like `strings`, `string-methods`, `loops`, and `lists`.  Requiring `classes` as a prerequisite means that students miss learning about `f-strings`, `str.format()`, and other options for strings that are pretty important early on, and are used (_or can be used_) in the exercises for `dicts`, `sets`, and `tuples`. 

Because a `concept exercise` can only use the techniques introduced in the exercise itself, or in its prerequisites, the reliance on a `class` for `pretty leaflet` needs to go. 

This can be an entire rewrite (_with new tasks or even a new story_), or just a rewrite of the stub, tests and the `examplar.py` code.  Contributors choice -- although, if the story/instructions/hints can be preserved, that would be ideal (and less work!).

[Pretty Leaflet Exercise Directory](https://github.com/exercism/python/tree/main/exercises/concept/pretty-leaflet)
[String-Formatting Concept Directory](https://github.com/exercism/python/tree/main/concepts/string-formatting)

From first glance, the following areas will need to be looked at or changed:

- [ ] exercise instructions
- [ ] exercise hints
- [ ] exercise `exemplar.py`
- [ ] exercise stub and test files

[...]

PR #2571, opened 2021-09-12
Pretty leaflet fixes

tigerBoss101:
From issue #2438, I removed the use of `classes` and replaced them with 4 functions instead. I kept the same tasks with some minor changes.

### DIFF
--- a/exercises/concept/pretty-leaflet/.docs/instructions.md
+++ b/exercises/concept/pretty-leaflet/.docs/instructions.md
@@ -11,7 +11,7 @@ The `capitalize_header` function should take in an event name and return the cap
 "Fan meetup"
 ```
 
-## 2. Add an optional date
+## 2. Format the date
 
 Create a new method `format_date` which takes in a list which contains a date, month, and a year. It displays the formatted date with the format `<month_name> <date>, <year>`
 
--- a/exercises/concept/pretty-leaflet/.docs/instructions.md
+++ b/exercises/concept/pretty-leaflet/.docs/instructions.md
@@ -36,7 +36,7 @@ When the method `display_icons` is called, the list of unicode code points passe
 ## 4. Display the finished leaflet
 
 Now you will use all the functions above, and combine them in the `print_leaflet` function.
-It should take an `event_name`, a list of `icons`, and an `event_date` list.
+It should take an `event_name`, a list of `icons`, a list of `authors`, and an `event_date` list.
 
 The poster should abide by the layout below.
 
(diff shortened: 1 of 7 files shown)


exercises/concept/pretty-leaflet/.docs/instructions.md
```diff
@@ -4,54 +4,49 @@ Your acquaintance Erin needs to print thousands of handbills for multiple events
 
 ## 1. Create the class with a capitalized header
```
BethanyG, line 5:
```suggestion
## 1. Create a capitalized header
```
```suggestion
## 1. Create the class with a capitalized header
```

exercises/concept/pretty-leaflet/.docs/instructions.md
```diff
@@ -4,54 +4,49 @@ Your acquaintance Erin needs to print thousands of handbills for multiple events
...
 
-The `Leaflet` class wraps all the instance variables and methods needed to create the leaflet. Implement that class, the header instance variable correspond to the first argument of the constructor.
-
-The constructor should take three parameters: the header of type `str`, one `list` of artists and one `list` of unicode code points.
-
-The header should be capitalized as illustrated in this example.
+The `capitalize_header` function should take in an event name and return the capitalized version of the event name.
```
BethanyG, line 7:
```suggestion
The `capitalize_header(<event_name>)` function should take in an event name `str` and return a `str` of the capitalized version.
```
```suggestion
The `capitalize_header` function should take in an event name and return the capitalized version of the event name.
```

exercises/concept/pretty-leaflet/.docs/instructions.md
BethanyG: Not sure why the interface is now "double suggesting".  Looks like some sort of bug.

BethanyG:
@tigerBoss101  -- I think this is a good first pass.  Thank you very much! 🌟   

I think this exercise has quite a bit more that needs to be done before it can go live on the site (_type hinting needs to be removed, docstrings need to be added, and other  "spec" related changes because of when this exercise was written_), but in the spirit of optimistic merging, I am going to accept this as-is, and outline next steps in a follow-on issue.  

Additionally, there appears to be a bug in the review UI for GH that is causing double comments for this PR, which is both confusing an annoying...so I think it will be better to do this in stages.

BethanyG (approved):
Approving.  See comments.  Best to handle additional detail in a follow-on, given the amount that needs changing.

PR #2574, opened 2021-09-13
tisbury cleanup: make instructions match unit tests

IsaacG:


BethanyG (approved):
Thanks for catching those extra quotes in the `clean_up()` example!

BethanyG (approved):
Thanks for the cleanup!

Issue #2572, opened 2021-09-12
Linked list tests - independent iterators


samfrances:
I have an idea for an additional test for the "Linked List" task.

I would love to contribute by creating a PR for this change, but I note you are not accepting unsolicited PRs at the moment, so I thought it best to create an issue first.

The task has the following test.

```
    @unittest.skip("extra-credit")
    def test_iterator(self):
        lst = LinkedList()
        lst.push(10)
        lst.push(20)
        iterator = iter(lst)
        self.assertEqual(next(iterator), 10)
        self.assertEqual(next(iterator), 20)
```
It would be good to have a test that `iter` returns an independent iterator each time, so that different "clients" don't step on each other's toes. Something like:

```
    @unittest.skip("extra-credit")
    def test_multiple_iterator(self):
        lst = LinkedList()
        lst.push(10)
        lst.push(20)

        iterator1 = iter(lst)
        iterator2 = iter(lst)
        self.assertEqual(next(iterator1), 10)
        self.assertEqual(next(iterator2), 10)
        self.assertEqual(next(iterator1), 20)
        self.assertEqual(next(iterator2), 20)
```

BethanyG:
Hi @samfrances!  👋🏽  

Thank you so much for writing up this issue and for asking before PR-ing!  🌟 

I don't see any harm in adding this to the test suite as extra credit.  Feel free to go ahead and PR it.  

It _may_ fail CI - I cannot remember if I need to regenerate anything when the tests change for this exercise .. but we will cross that bridge when we get to it.  😄

BethanyG:
Closed by #2580

PR #2580, opened 2021-09-14
Add iterator independence test to linked-list exercise.

samfrances:
This is a PR for https://github.com/exercism/python/issues/2572.

This PR adds a test to the linked-list exercise, which checks that different calls to `iter()` on the same `LinkedList` instance return independent iterators, (i.e. which change state independently of each other).


BethanyG (approved):
Awesome!  Thank you for this.  🌟

PR #2583, opened 2021-09-15
Update tuples.py

toutios:
Changing docstring in line 47 from "tuple of tuples" to "string of tuples separated by newlines", which is what tuple_test expects and the website describes.


BethanyG (approved):
Good catch!  Thanks for submitting this.  🌟

PR #2584, opened 2021-09-15
Fix a typo in `meltdown-mitigation` instructions

pranasziaukas:
idea -> ideal


BethanyG (approved):
Now that the previous PR and it's changes are merged, I think we can go ahead.  Many thanks for the catch!

PR #2590, opened 2021-09-20
Improve tests of `card-games` exercise

pranasziaukas:
- Fix unmatched visual indentation in `TestConcatenateRounds`
- Add a test case for `TestApproxAverageIsAverage` to also check if median equals average

The latter was also not implemented in the example solution, so added it there as well.

### DIFF
--- a/exercises/concept/card-games/.meta/exemplar.py
+++ b/exercises/concept/card-games/.meta/exemplar.py
@@ -52,7 +52,14 @@ def approx_average_is_average(hand):
     :return: bool - is approximate average the same as true average?
     """
 
-    return card_average([hand[0], hand[-1]]) == card_average(hand)
+    real_average = card_average(hand)
+    if card_average([hand[0], hand[-1]]) == real_average:
+        is_same = True
+    elif hand[len(hand) // 2] == real_average:
+        is_same = True
+    else:
+        is_same = False
+    return is_same
 
 
 def average_even_is_average_odd(hand):
--- a/exercises/concept/card-games/lists_test.py
+++ b/exercises/concept/card-games/lists_test.py
@@ -60,9 +60,9 @@ class TestConcatenateRounds(unittest.TestCase):
         want = []
 
         self.assertEqual(concatenate_rounds(rounds_1, rounds_2),
-            want,
-            msg=f'Expected {want} but got an incorrect result.'
-            )
+                         want,
+                         msg=f'Expected {want} but got an incorrect result.'
+                         )
 
     @pytest.mark.task(taskno=2)
     def test_other(self):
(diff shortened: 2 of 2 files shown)


pranasziaukas:
Also hoping to do a follow-up PR refactoring the tests to make use of `subTest`.

exercises/concept/card-games/.meta/exemplar.py
```diff
@@ -52,7 +52,13 @@ def approx_average_is_average(hand):
     :return: bool - is approximate average the same as true average?
     """
 
-    return card_average([hand[0], hand[-1]]) == card_average(hand)
+    real_average = card_average(hand)
+    is_same = False
+    if card_average([hand[0], hand[-1]]) == real_average:
+        is_same = True
+    elif hand[len(hand) // 2] == real_average:
+        is_same = True
```
IsaacG, line 60:
```suggestion
    if card_average([hand[0], hand[-1]]) == real_average:
        is_same = True
    elif hand[len(hand) // 2] == real_average:
        is_same = True
    else:
        is_same = False
```

exercises/concept/card-games/lists_test.py
```diff
@@ -60,9 +60,9 @@ def test_empty(self):
         want = []
 
         self.assertEqual(concatenate_rounds(rounds_1, rounds_2),
```
IsaacG, line 62: There should be a newline after `assertEqual(` and indentation should be one level (4 spaces)

exercises/concept/card-games/lists_test.py
```diff
@@ -209,6 +209,18 @@ def test_instructions_example_3(self):
             msg=f'Expected {want} but got an incorrect result: {got!r}'
         )
 
+    @pytest.mark.task(taskno=5)
+    def test_median_true(self):
+        hand = [1, 2, 4, 5, 8]
+        want = True
+        got = approx_average_is_average(hand)
+
+        self.assertEqual(
```
IsaacG, line 218: If `want` is not parameterized, you can just use `self.assertTrue()`

BethanyG:
I cannot believe I missed this the first several rounds.  The `exemplar.py` for this exercise includes `loops`.  Loops is not a prerequisite, and we can't make it a prerequisite, because `loops` already requires `lists`.  🤦🏽‍♀️   

So we need to revise the `introduction.md` and `instructions.md` to cover the built-ins `sum()` and `len()`.  Then `card_average` in `exemplar.py` will need to be re-written to use `return sum(hand)/len(hand)`.  We can't loop there.  😞 

So - this will now need further revision.  Rather than do that here, let's put the above change into the PR with the test file re-writes.

BethanyG (approved):
@pranasziaukas  -- Thank you for your work on this!  🌟   

I am going to approve and merge this as-is.  But then we will need to re-write the tests and also change the `instructions`, `exemplar.py`,  and `introduction` per my comments earlier.  I'll file and issue and assign it to you, if you don't mind?

pranasziaukas:
> I am going to approve and merge this as-is. But then we will need to re-write the tests and also change the `instructions`, `exemplar.py`, and `introduction` per my comments earlier. I'll file and issue and assign it to you, if you don't mind?

Sounds good @BethanyG 👍

PR #2606, opened 2021-09-27
Update tuples_test.py

squeakyboots:
Adding __name__ == "__main__" line for consistency with other exercises' test files.


BethanyG (approved):

BethanyG:
We've switched to using `PyTest` for our test runner, so this change is not strictly necessary.  But it doesn't harm anything, so we can keep it.

PR #2607, opened 2021-09-28
Update instructions.md

ioload:
It looks like the function in Step 5 'total_time_in_minute' should be 'elapsed_time_in_minutes' based on Step 4.


BethanyG (approved):
Good catch.  Thanks for submitting this! 🌟

PR #2612, opened 2021-09-29
Correct fail_safe docstring

cyborgsphinx:
I noticed this while implementing a solution for meltdown-mitigation. The README and tests agreed that it should be 90%, but the docstring showed 40%.


BethanyG (approved):
Thanks for the catch!  🌟

PR #2613, opened 2021-09-29
cater-waiter: Minor code example fixes

squeakyboots:
The example code for instruction 3 has a tuple as input but per the test cases and pre-fab docstring it should take a "str, list" (as opposed to "tuple(str, list)"). In short there are extra parenthesis.

In instruction 4 the second code example returns `{'blue cheese', 'pinenuts', 'pork tenderloin'}` but with the current data in sets_categories_data.py should return `{'pork tenderloin', 'blue cheese', 'pine nuts', 'onions'}` after correcting the missing space in pine nuts in the input. I'm guessing onions were maybe added later and pine nuts was spaced after the instruction was written.


BethanyG (approved):
Thanks for catching this!  Much appreciated.  🌟

PR #2619, opened 2021-10-01
[Guido's Gorgeous Lasagna] Added Try-except Around Imports in Test File

BethanyG:
`Guido's Gorgeous Lasagna` (basics concept) 
`lasagna_test.py` -- Added a try around the names imported from `lasagna.py` and re-worded the raised error messaging.

### DIFF
--- a/exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
+++ b/exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
@@ -13,14 +13,15 @@ try:
 
 except ImportError as import_fail:
     message = import_fail.args[0].split('(')[0]
-    function_name = import_fail.args[0].split()[3][:-1] + "()'"
+    item_name = import_fail.args[0].split()[3]
 
     if 'EXPECTED_BAKE_TIME' in message:
-        raise ImportError(f"We can't the find the constant 'EXPECTED_BAKE_TIME' in your 'lasagna.py'"
-                          f" file. Did you mis-name or forget to define it?")
+        raise ImportError(f"We can't the find the constant {item_name} in your 'lasagna.py'"
+                          " file. Did you mis-name or forget to define it?")
     else:
-        raise ImportError(f"In your 'lasagna.py' file, we can't find the function named {function_name}."
-                          f" Did you mis-name or forget to define it?")
+        item_name = item_name[:-1] + "()'"
+        raise ImportError(f"In your 'lasagna.py' file, we can't find the function named {item_name}."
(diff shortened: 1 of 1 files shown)


exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
```diff
@@ -1,11 +1,26 @@
...
+                        elapsed_time_in_minutes
+                         )
+
+
+except ImportError as import_fail:
+    message = import_fail.args[0].split('(')[0]
+    function_name = import_fail.args[0].split()[3][:-1] + "()'"
+
+    if 'EXPECTED_BAKE_TIME' in message:
+        raise ImportError(f"We can't the find the constant 'EXPECTED_BAKE_TIME' in your 'lasagna.py'"
```
IsaacG, line 19: `pylint` ought to catch this line for being an f-string without any formatting.

exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
```diff
@@ -1,11 +1,26 @@
...
+                          f" file. Did you mis-name or forget to define it?")
```
IsaacG, line 20: Usually I've seen trailing spaces at the end. I'm not sure if there's a convention around that.

exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
```diff
@@ -1,11 +1,26 @@
...
+    else:
+        raise ImportError(f"In your 'lasagna.py' file, we can't find the function named {function_name}."
+                          f" Did you mis-name or forget to define it?")
```
IsaacG, line 23: Ditto on an `f-string` without any formatting.

exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
J08K: It does, `pylint-testing.py:4:25: W1309: Using an f-string that does not have any interpolated variables (f-string-without-interpolation)`, but I do not think is a big issue. It is not in PEP8, which is the code-style we really do adhere to. I don't think it is an issue and I'll leave it up to @BethanyG to decide.

exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
J08K: I think it is better to keep the consistency with these types. I suggest not to change this.

J08K (approved):

Issue #2353, opened 2021-02-16
OLD functions concept exercise


BethanyG:
This issue describes how to implement the `functions` concept and concept exercise for the Python track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [Contributing to Exercism](https://exercism.org/docs/building) | [Exercism and GitHub](https://exercism.org/docs/building/github) | [Contributor Pull Request Guide](https://exercism.org/docs/building/github/contributors-pull-request-guide)
- [What are those Weird Task Tags about?](https://exercism.org/docs/building/product/tasks)
- [Building Language Tracks: An Overview](https://exercism.org/docs/building/tracks) 
- [What are Concepts?](https://exercism.org/docs/building/tracks/concepts)
- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Specifications](https://exercism.org/docs/building/tracks/concepts)
- [Concept Exercise Specifications](https://exercism.org/docs/building/tracks/concept-exercises) 
- [Exercism Formatting and Style Guide](https://exercism.org/docs/building/markdown/style-guide)
- [Exercism Markdown Specification](https://exercism.org/docs/building/markdown/markdown)
- [Reputation](https://exercism.org/docs/using/product/reputation)  

## Goal

This `functions` concept + concept exercise is meant to teach a deeper understanding and use of `functions`  in Python. It should also explain how Python treats/views functions (_as callable objects_).

## Learning Objectives

<details>
<summary><b>In Scope</b></summary>

- Understand more about  Python `scopes` and `namespaces`
  - understand the difference between the `global` and `nonlocal` keywords and when to use them
- Get familiar with the [special attributes](https://docs.python.org/3/reference/datamodel.html#types) of Python functions
- Get familiar with best practices when using `return`, and the difference between _explicit_ and _implicit_ `return`
   - functions without an explicit return will return the singleton object `None`
- Understand what is meant by _"functions are first class **objects** in Python"_.
  - understand that `functions` **are** `objects` in Python, and that they have `types`
  - understand that `functions` can be assigned to variables, used in expressions,  and stored in various data structures such as `dicts` or `lists`
  - create `functons` that are assigned to variables, used in expressions, and stored in different data structures.
  - understand and create` functions` that are/can be nested inside one another
[...]

PR #2623, opened 2021-10-02
[New Concept] : functions

mukeshgurpude:
Adding functions `concept`, as discussed in #2353

### DIFF
--- a/concepts/functions/.meta/config.json
+++ b/concepts/functions/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "A function is a block of code which only runs when it is called. It can pass data, known as parameters, into a function. A function can return data as a result for the further use.",
+  "blurb": "A function is a reusable code block that achieves a specific purpose.",
   "authors": ["bethanyg", "cmccandless"],
   "contributors": []
 }
--- a/concepts/functions/.meta/config.json
+++ b/concepts/functions/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "A function is a reusable code block that achieves a specific purpose.",
-  "authors": ["bethanyg", "cmccandless"],
+  "blurb": "A function is any object to which the *function call* operation can be applied. The body contains one or more statements which return some value to the caller. It can receive zero or more arguments (including other functions). Function definitions use the keyword def, which binds the function name to the object definition in the local namespace.",
+  "authors": ["bethanyg", "mukeshgurpud"],
   "contributors": []
 }
--- a/concepts/functions/.meta/config.json
+++ b/concepts/functions/.meta/config.json
@@ -1,5 +1,5 @@
 {
   "blurb": "A function is any object to which the *function call* operation can be applied. The body contains one or more statements which return some value to the caller. It can receive zero or more arguments (including other functions). Function definitions use the keyword def, which binds the function name to the object definition in the local namespace.",
-  "authors": ["bethanyg", "mukeshgurpud"],
+  "authors": ["bethanyg", "mukeshgurpude"],
   "contributors": []
 }
--- a/concepts/functions/.meta/config.json
+++ b/concepts/functions/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb for this concept",
+  "blurb": "A function is a block of code which only runs when it is called. It can pass data, known as parameters, into a function. A function can return data as a result for the further use.",
   "authors": ["bethanyg", "cmccandless"],
   "contributors": []
 }
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -24,23 +24,25 @@ def function_name(parameter1, parameter2, ...):
 ```
 
 We can also define a function without any parameters or return value.
+
 ```python
 def function_name():
     # function body
 ```
-
 
 ## Calling a Function
 
 To call a function, use the function name followed by parenthesis [`()`]. Parameters passed to the function are placed inside the parenthesis, separated by commas.
 
 Consider the following function:
+
 ```python
 def wish():
     print("Hello")
 ```
 
 The above function can be called by using the following syntax:
+
 ```python
 >>> wish()
 Hello
@@ -51,12 +53,14 @@ Hello
 Parameters are values that are passed to the function when it is called. They can be of any data type.
 
 Let's define a function `add` which adds two numbers together:
+
 ```python
 def add(x, y):
     print(x + y)
 ```
 
 When the function is called, the parameters are passed to the function. We need to pass values for both the parameters, otherwise a [`TypeError`][type-error] will be raised.
+
 ```python
 >>> add(2, 3)
 5
@@ -84,12 +88,14 @@ TypeError: add() takes 2 positional arguments but 3 were given
 The return value is a value that is returned to the caller of the function. Return value can be of any data type. It can be used by caller of the function to perform further operations. If the function does not explicitly return a value, the value `None` is returned.
 
 Let's define a function `add`:
+
 ```python
 def add(x, y):
     return x + y
 ```
 
 We can store the return value in a variable and then print it:
+
 ```python
 >>> result = add(2, 3)
 >>> print(result)
@@ -114,6 +120,7 @@ None
 ```
 
 Use of `return` immediately exits the function and returns the value to the caller.
+
 ```python
 >>> def show(x, y):
     print(x)
@@ -130,9 +137,10 @@ Use of `return` immediately exits the function and returns the value to the call
 Complex programs can be broken down into smaller parts. Different functions can be used to perform different specific tasks.
 
 Assume a program has to perform the following tasks:
-  * Calculate the area of a circle
-  * Calculate the area of a rectangle
-  * Calculate the area of a triangle
+
+* Calculate the area of a circle
+* Calculate the area of a rectangle
+* Calculate the area of a triangle
 
 We can break down the program into smaller parts.
 
@@ -152,10 +160,12 @@ Now, we can call the functions in the order we want.
 ```python
 >>> circle_area(2)
 12.56
+>>> triangle_area(2, 3)
+3.0
 >>> rectangle_area(2, 3)
 6
->>> triangle_area(2, 3)
-1.5
+>>> rectangle_area(1, 2) + circle_area(2) + triangle_area(1, 2)
+15.56
 ```
 
 ## Scope of Variables
@@ -182,6 +192,7 @@ Inside function: 10
 ```
 
 We can access the variable inside the outer function using the `global` keyword.
+
 ```python
 >>> x = 30
 >>> def random_function():
@@ -199,11 +210,9 @@ Inside function: 10
 10
 ```
 
-
 ## Functions as first class objects
 
 In python, functions can be assigned to variables and passed as arguments to other functions. They can be used as return values. Functions can also be placed into a sequence([`list`][list], [`tuple`][tuple] etc) or as value in a [`dict`][dict]. Functions can be used anywhere than any other object can be used. This is because _functions are [`first class objects`][first class objects]_.
-
 
 ```python
 # print is a function
@@ -239,6 +248,7 @@ Hello
 ```
 
 Functions can also be nested inside other functions.
+
 ```python
 def outer():
     x = 10
@@ -257,11 +267,12 @@ The inner function can access the variable `x` defined in the outer function.
 ## Special Attributes
 
 Functions in python have special attributes. Some of them are:
-  * `__name__`: Name of the function
-  * `__doc__`: Documentation string of the function
-  * `__module__`: Module in which the function is defined
-  * `__globals__`: Dictionary of global variables in the function
-  * `__code__`: Code object containing the instructions of the function
+
+* `__name__`: Name of the function
+* `__doc__`: Documentation string of the function
+* `__module__`: Module in which the function is defined
+* `__globals__`: Dictionary of global variables in the function
+* `__code__`: Code object containing the instructions of the function
 
 ```python
 >>> def add(x, y):
@@ -288,12 +299,11 @@ Functions in python have special attributes. Some of them are:
 
 The full list of function attributes can be found at [Python DataModel][attributes].
 
-
 [attributes]: https://docs.python.org/3/reference/datamodel.html#index-33
 [build-in functions]: https://docs.python.org/3/library/functions.html
 [dict]: https://docs.python.org/3/tutorial/datastructures.html#dictionaries
 [first class objects]: https://en.wikipedia.org/wiki/First-class_object
-[function]: https://en.wikipedia.org/wiki/Function_(computer_science)
+[function]: https://docs.python.org/3/glossary.html#term-function
 [global]: https://docs.python.org/3/reference/compound_stmts.html#global
 [list]: https://docs.python.org/3/tutorial/datastructures.html#list-objects
 [map]: https://docs.python.org/3/library/functions.html#map
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -1,8 +1,8 @@
 # About
 
-A [`function`][function] is a block of organized, reusable code that is used to perform a single, related action.
+A [`function`][function] is a block of organized, reusable code that is used to perform a specific task.
 `Functions` provide better modularity for your application and a high degree of code reuse.
-Like other programming languages, python provides _pre-defined functions_ ([`print`][print], [`map`][map] etc) that are readily available in the Python language. 
+Python, like other programming languages, has [_built-in functions_][build-in functions] ([`print`][print], [`map`][map], and so on) that are readily available.
 You can also define your own functions. Those are called [`user-defined functions`][user defined functions].
 Functions can run something as simple as _printing a message to the console_ or they can be quite complex.
 
@@ -48,7 +48,7 @@ Hello
 
 ## Parameters
 
-Parameters are values that are passed to the function when it is called. They are known as `arguments`.
+Parameters are values that are passed to the function when it is called. They can be of any data type.
 
 Let's define a function `add` which adds two numbers together:
 ```python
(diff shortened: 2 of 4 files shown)


concepts/functions/.meta/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb for this concept",
+  "blurb": "A function is a block of code which only runs when it is called. It can pass data, known as parameters, into a function. A function can return data as a result for the further use.",
```
IsaacG, line 2: > It can pass data, known as parameters, into a function. 

"It" here would be "functions" so this reads "functions can pass data into a function". Which is confusing and circular.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
 # About
 
-TODO: add information on functions concept
+A [`function`][function] is a block of organized, reusable code that is used to perform a single, related action.
```
IsaacG, line 3: "single" implies one. "Related" implies multiple. Related to what?

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+`Functions` provide better modularity for your application and a high degree of code reusing.
```
IsaacG, line 4:
```suggestion
`Functions` provide better modularity for your application and a high degree of code reuse.
```

Better than what?

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+Like other programming languages, python provides _pre-defined functions_ ([`print`][print], [`map`][map] etc) that are readily available in the Python language.
```
IsaacG, line 5: These have a technical name: built-in functions. It would be good to use the proper name when they are first introduced. 
It might be good to link to them, too: `https://docs.python.org/3/library/functions.html`

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
 # About
 
-TODO: add information on functions concept
+A [`function`][function] is a block of organized, reusable code that is used to perform a single, related action.
+`Functions` provide better modularity for your application and a high degree of code reusing.
+Like other programming languages, python provides _pre-defined functions_ ([`print`][print], [`map`][map] etc) that are readily available in the Python language. 
+However, you can define your own functions as well which are called [`user-defined functions`][user defined functions].
```
IsaacG, line 6:
```suggestion
You can also define your own functions. Those are called [`user-defined functions`][user defined functions].
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+Functions can be as simple as _printing a message to the console_ or as complex as _performing a calculation_.
```
IsaacG, line 7:
```suggestion
Functions can run something as simple as _printing a message to the console_ or they can be quite complex.
```

`a + b` is a calculation and is not very complex.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+To execute the code inside a function, you need to call the function, which is done by using the function name followed by parenthesis [`()`].
+Data known as `parameters` can be passed to the function inside the parenthesis. Broader terms for parameters are `arguments`. Functions can perform different tasks depending on the value of the parameters.
```
IsaacG, line 10:
```suggestion
Data. known as `parameters`, can be passed to the function by placing them inside the parenthesis. A broader term for parameters is `arguments`. Functions can perform different tasks depending on the value of the parameters.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+The function can also return a value using the `return` keyword. The value returned by the function is known as the `return value`. The return value is returned to the caller of the function.
```
IsaacG, line 12:
```suggestion
A function can also return a value using the `return` keyword. The value returned by the function is known as the `return value`. The return value is returned to the caller of the function.
```

IsaacG: Has "caller" been defined?

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+## Creation
+
+In python, functions are created using the `def` keyword. The function definition is followed by the function name and parenthesis [`()`]. Inside the parenthesis, the parameters are specified, and are separated by commas. After the parameters, the colon (`:`) is used to separate the function header from the function body.
```
IsaacG, line 16:
```suggestion
In python, functions are created using the `def` keyword. The function definition is followed by the function name and parenthesis [`()`]. Inside the parenthesis, the parameters are specified, separated by commas. After the close parenthesis, the colon (`:`) is used to separate the function signature from the function body.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+The function body is a block of code that is executed when the function is called. The body of the function is indented. The indentation is important because the function body is a block of code. A value can be returned from the function by using the `return` keyword, which will be used by the caller of the function.
```
IsaacG, line 18:
```suggestion
The function body is a block of code that is executed when the function is called. The body of the function is indented. The indentation is important because Python relies on it to know where that block of code ends. A value can be returned from the function by using the `return` keyword, which can then be used by the caller of the function.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+We can also define a function without any parameters or return value.
+```python
+def function_name():
+    # function body
+```
+
+
+## Calling a Function
+
+To call a function, use the function name followed by parenthesis [`()`]. The parameters passed to the function are separated by commas.
```
IsaacG, line 35:
```suggestion
To call a function, use the function name followed by parenthesis [`()`]. Parameters passed to the function are placed inside the parenthesis, separated by commas.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+Consider the following function:
+```python
+def wish():
+    print("Hello")
+```
+
+Above function can be called by using the following syntax:
```
IsaacG, line 43:
```suggestion
The above function can be called by using the following syntax:
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+```python
+>>> wish()
+Hello
+```
+
+## Parameters
+
+Parameters are values that are passed to the function when it is called. They are known as `arguments`.
```
IsaacG, line 51: This document first mentions arguments, switches to parameters and is now saying they are known as arguments. One term should be used consistently.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+Let's define a function `add`:
```
IsaacG, line 53:
```suggestion
Let's define a function `add` which adds two numbers together:
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+```python
+def add(x, y):
+    print(x + y)
+```
+
+When the function is called, the parameters are passed to the function. We need to pass values for both the parameters, otherwise [`TypeError`][type-error] will be raised.
```
IsaacG, line 59:
```suggestion
When the function is called, the parameters are passed to the function. We need to pass values for both the parameters, otherwise a [`TypeError`][type-error] will be raised.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+>>> add(2, 3)
+5
+
+# Function can be called multiple times, with different parameters
+>>> add(4, 3)
+7
+>>> add(5, 6)
+11
+
+# Passing incorrect number of parameters will raise `TypeError`
```
IsaacG, line 70:
```suggestion
# Passing an incorrect number of parameters will raise a `TypeError`
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+>>> add(2, 3, 4)
+Traceback (most recent call last):
+  File "<stdin>", line 1, in <module>
+TypeError: add() takes 2 positional arguments but 3 were given
+```
+
+## Return Value
+
+The return value is a value that is returned to the caller of the function. Return value can be of any data type. It can be used by caller of the function to perform further operations. If the function does not explicitly return a value, the value `None` is returned.
```
IsaacG, line 84: You call out that return values can be of any type. What about the parameter type? That was never mentioned.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+Let's define a function `add`:
+```python
+def add(x, y):
+    return x + y
+```
+
+We'll store the return value in a variable and print it:
```
IsaacG, line 92:
```suggestion
We can store the return value in a variable and then print it:
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+5
+# Type of result is `int`
+>>> type(result)
+<class 'int'>
+
+# We can also perform operations on the return value
+>>> result * 2
+10
+
+# Function without return value returns `None`
```
IsaacG, line 105:
```suggestion
# A function without an explicit return value will return `None`
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+```
+
+Use of `return` immediately exits the function and returns the value to the caller.
+```python
+>>> def show(x, y):
+    print(x)
+    return x
+    print(y)
+
+# y never gets printed, because function exists after return statement
```
IsaacG, line 123:
```suggestion
# y never gets printed, because the function exits after the return statement
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+>>> show(1, 2)
+1
+```
+
+## Modularity
+
+Complex programs can be broken down into smaller parts. Functions can be used to perform different tasks.
```
IsaacG, line 130:
```suggestion
Complex programs can be broken down into smaller parts. Different functions can be used to perform different specific tasks.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+12.56
+>>> rectangle_area(2, 3)
+6
+>>> triangle_area(2, 3)
+1.5
+```
+
+## Scope of Variables
+
+If variable is defined inside a function, then it will be only accessible inside the function. If we want to access the variable outside the function, we need to use the [`global`][global] keyword. [`nonlocal`][nonlocal] keyword is used to access the variable inside a nested function.
```
IsaacG, line 163: `nonlocal` isn't commonly used and might be best left out of this. `global` isn't that common, either.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+```python
+def outer():
```
IsaacG, line 166: `outer` implies an `inner`. Outer of what? Where's the inner?

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+    x = 10
+    print('Inside function:', x)
+```
+
+variable `x` is defined inside the outer function. It is limited to the scope of the outer function only. It will not alter the value of the variable outside the outer function.
+
+```python
+>>> x = 30
```
IsaacG, line 174: You may want to move this line *before* the function definition so the reader knows that there's a global `x` before the function is shown.

concepts/functions/about.md
IsaacG: This line also has both `python` and `Python`. Both are used throughout this file.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+>>> outer()
+Inside function: 10
+>>> x
+10
+```
+
+
+## Functions as first class objects
+
+In python, functions can be assigned to variables and passed as arguments to other functions. They can also be used as return values in other functions. Functions can also be used as items in a sequence([`list`][list], [`tuple`][tuple] etc) or as value in [`dict`][dict]. This is called _function as [`first class object`][first class objects]_.
```
IsaacG, line 204:
```suggestion
In python, functions can be assigned to variables and passed as arguments to other functions. They can be used as return values. Functions can also be placed into a sequence([`list`][list], [`tuple`][tuple] etc) or as value in a [`dict`][dict]. Functions can be used anywhere than any other object can be used. This is because _functions are [`first class objects`][first class objects]_.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+
+
+```python
+# print is a function
+
+# function can be assigned to a variable
```
IsaacG, line 210:
```suggestion
# A function can be assigned to a variable.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+>>> fake_print = print
```
IsaacG, line 211: This looks like the real print but an alias.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+>>> fake_print("Hello")
+Hello
+>>> type(fake_print)
+<class 'builtin_function_or_method'>
+
+# function can be passed as an argument to another function
```
IsaacG, line 217:
```suggestion
# Functions can be passed as an argument to another function.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+>>> def check_print(func):
+    func("Hello")
+>>> check_print(print)
+Hello
+
+# function can be used as an item in a sequence
```
IsaacG, line 223:
```suggestion
# Function can be used as an item in a sequence.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+>>> my_list = [print, "Hello"]
+>>> my_list[0]("Hello")
+Hello
+
+# function can be used as a value in a dictionary
```
IsaacG, line 228:
```suggestion
# Functions can be used as a value in a dictionary.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+>>> my_dict = {"key": print}
+>>> my_dict["key"]("Hello")
+Hello
+
+# function can be used as a return value in another function
```
IsaacG, line 233:
```suggestion
# Functions can be returned from a function.
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+    return print
+>>> return_func()("Hello")
+Hello
+```
+
+Functions can also be nested inside other functions.
+```python
+def outer():
+    x = 10
+    def inner():
```
IsaacG, line 244: This is not used often and is often not good practice. Maybe it should be skipped.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+The inner function can access the variable `x` defined in the outer function.
+
+```python
+>>> outer()
+10
+```
+
+## Special Attributes
+
+Functions in python have some special attributes. Some of them are:
```
IsaacG, line 258:
```suggestion
Functions in python have special attributes. Some of them are:
```

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+```
+
+The inner function can access the variable `x` defined in the outer function.
+
+```python
+>>> outer()
+10
+```
+
+## Special Attributes
```
IsaacG, line 256: This is another section that might not be well suited for teaching the concept of functions.

concepts/functions/about.md
```diff
@@ -1,3 +1,302 @@
...
+'Return the absolute value of the argument.'
+
+# Module
+>>> len.__module__
+'builtins'
+>>> add.__module__
+'__main__'
+```
+
+Full list of attributes can be found at [Python DataModel][attributes].
```
IsaacG, line 288:
```suggestion
The full list of function attributes can be found at [Python DataModel][attributes].
```

concepts/functions/about.md
mukeshgurpude: Actually in learning Objective section of original issue #2353, these are mentioned.
Do you want me to remove these points?

mukeshgurpude:
Hello @J08K, @BethanyG, @IsaacG 
Please review the PR. 
Please accept my apologies for my sloppy grammar😥.

concepts/functions/about.md
J08K: It is in the learning objectives, so keep them. Even though they aren't "used often" that doesn't mean people shouldn't know what they are.

concepts/functions/about.md
J08K: Please read the connected issue before reviewing. `nested functions` is in the concepts of #2353. Keep nested functions in the exercise.

Nested functions are not bad practice at all. They can be used to access the enclosing namespace. If such function is only used by one other function, then you could opt in to use `nested` or `inner functions`.

J08K: If you really would like these gone. Throw up an issue on the repo. We could either then put this issue on hold, or merge this issue (if everything else is okay) and then change the docs afterwards.

concepts/functions/about.md
J08K: Please reply to the connected issue with your concerns, since it is inside of the concepts. This is not the place for such discussions.

concepts/functions/about.md
BethanyG: I think [issue 2353](https://github.com/exercism/python/issues/2353) is pretty clear. The intent of this exercise is to dig deeper into Python functions - both as a "next step" for students and as an introduction into `higher order functions` and FP.  As well as more practice before students are "hit" with all the function parameter options, and unpacking actions (function arguments exercise).

Python's object model (as opposed to Javas object model, for example) treats `function objects` as first-class citizens.  It is possible (perhaps not _advisable_) to write nested functions -- or even constructs  where one function encloses multiple other functions. Students need to know that -- even if they don't end up using it all that often.

And that also helps teach scope (along with locals and globals) -- which (we hope) leads eventually to understanding what a `closure` is, and why you'd want one, and other things like `decorators`.

BethanyG: This was not intended as a "basic" exercise.  We have `lasgana` -- and all the exercises in-between for that.  Students would not be taking on this exercise until after they finished all the exercises on data structures and sequences and more.

BethanyG:
@mukeshgurpude  -- we all have sloppy grammar (_I can be a big offender there_), so there is no need to apologize!  It's why we have reviews. 😄   🌈 

Although I tend to want to do content review passes before grammar passes -- since re-working content often leads to additional typos or potential issues.  

But don't worry - and don't get intimidated.  We're all here to work on this and make it the best it can be.  And _**thank you**_ for submitting!  I probably won't have time to give a thorough review until tomorrow evening my time (I'm in PST) - but I'll have some more comments then. 🌟

J08K:
So, what @BethanyG said. I agree, we might need to _hint_ a little towards the more _advanced_ uses of functions. It is ok (in my opinion) if you don't cover these in full detail, but we need to let students know that there is more than just the `def` keyword. 

I really like everything you've written so far, and I'd love to hear about what you have to say.

BethanyG:
@mukeshgurpude   In the spirit of "optimistic merging", I am going to approve this PR.  We can continue to discuss and change it in follow-on requests.  Apologies for letting it "fall through the cracks" -- I do think that we should discuss the "advanced" part of this and map out in more detail what we intend for functions.  But rather than drag this particular PR on, I think we're better served with discussion via an issue, and then making follow-on PRs.

BethanyG (approved):

Issue #2627, opened 2021-10-05
[Ghost Gobble Arcade Game] Lacking test for win conditions


s-torn:
The Ghost Gobble Arcade Game exercise is solvable by defining a win only as not losing, the condition of having eaten all dots is not tested. 
In the current state the exercise passes the tests if the win function returns _not lose(power_pellet_active,touching_ghost)_.

s-torn:
I'm not able to link my pull request, which I think is about permissions?
#2628

J08K:
Hi there, nice catch! I did test out what you meant and you're right, the tests do allow for a student to pass the exercise even if they don't use `has_eaten_all_dots`. Funny that this was never noticed... I wonder how many students used this loop-hole to pass the exercise.

I will be reviewing your PR #2628 about this!

PR #2628, opened 2021-10-05
[Ghost Gobble Arcade Game] : Add missing test-case.

s-torn:
Resolves: #2627


J08K:
Hey, I edited your PR description, so that it mentions the issue. I will reply to your issue about the matter.

J08K:
This PR would enforce the correct solution to the exercise, as it would otherwise be possible to pass the exercise, without using the first parameter (`has_eaten_all_dots`).

exercises/concept/ghost-gobble-arcade-game/arcade_game_test.py
```diff
@@ -100,3 +100,11 @@ def test_win_if_all_dots_eaten_and_touching_a_ghost_with_a_power_pellet_active(s
             True,
             msg="win if all dots eaten and touching a ghost with a power pellet active"
         )
+
+    @pytest.mark.task(taskno=4)
+    def test_dont_win_if_not_all_dots_eaten(self):
```
J08K, line 105: I tested this code both locally and on the `python-test-runner`, it works completely OK. Exemplar passes the tests and the previous `WRONG Example` is now failing, which is a good thing in this case.

J08K (approved):

Issue #2630, opened 2021-10-06
Grains exercise missing instruction.


BrooksAz:
Testing fails on this exercise if you don't raise a value error when the number is over 64 (the number of squares on the board). This is not mentioned in the instructions so I am going to make a pull request adding this as a step.

BrooksAz:
[https://github.com/exercism/python/pull/2631](url)

IsaacG:
Per my comment on the exercise, many of the exercises on Exercism are built with the problem description laying out the broad expectations and the unit tests encoding the details of the expected behavior. Users are expected to take a test-driven approach to solving the exercises and filling out the details. This is by design and not an oversight.

PR #2631, opened 2021-10-06
Update instructions.md

BrooksAz:
I added the missing value error instructions and a hint about bitwise ops for the bonus point section.

### DIFF
--- a/exercises/practice/grains/.docs/instructions.md
+++ b/exercises/practice/grains/.docs/instructions.md
@@ -13,14 +13,15 @@ There are 64 squares on a chessboard (where square 1 has one grain, square 2 has
 
 Write code that shows:
 - how many grains were on a given square, and
-- the total number of grains on the chessboard
+- the total number of grains on the chessboard,
+- and throws a Value Error for numbers over 64.
 
 ## For bonus points
 
 Did you get the tests passing and the code clean? If you want to, these
 are some additional things you could try:
 
-- Optimize for speed.
+- Optimize for speed. Bitwise operators anyone?
 - Optimize for readability.
 
 Then please share your thoughts in a comment on the submission. Did this


BrooksAz:
#2630

exercises/practice/grains/.docs/instructions.md
```diff
@@ -13,14 +13,15 @@ There are 64 squares on a chessboard (where square 1 has one grain, square 2 has
 
 Write code that shows:
 - how many grains were on a given square, and
-- the total number of grains on the chessboard
+- the total number of grains on the chessboard,
+- and throws a Value Error for numbers over 64.
```
IsaacG, line 17: Many of the Exercism exercises hides the expected implementation details inside the unit tests. If we want to spell out the expectations in details, that should be an Exercism-wide conversation first.

exercises/practice/grains/.docs/instructions.md
```diff
@@ -13,14 +13,15 @@ There are 64 squares on a chessboard (where square 1 has one grain, square 2 has
...
 
 ## For bonus points
 
 Did you get the tests passing and the code clean? If you want to, these
 are some additional things you could try:
 
-- Optimize for speed.
+- Optimize for speed. Bitwise operators anyone?
```
IsaacG, line 24: 1. If you want to have people look into bitwise operators, a "read up on bitwise operators here `link`" might be better than a question.
2. Out of curiosity, are those actually faster or is that something that's already optimized for you by the interpretter?

BethanyG:
Adding the same note here as I did in #2630 -- just for context/reminder.  I will be closing this PR, since this requires an exercise append file OR a change in `problem-specifications`.  

@BrooksAz  -- Thanks for doing this work. Unfortunately, we can't accept this PR as-is.  Please see [Issue 2563](https://github.com/exercism/python/issues/2563) for some added context.  

The majority of Practice exercises are auto-generated from cross-track specifications in the [problem-specfications](https://github.com/exercism/problem-specifications) repo, so changing instructions is not as straightforward as a Pull Request to the Python track.  Any change needs to be raised in `problem-specifications`, discussed, and then approved by at least three different track maintainers.

Additionally, we are still working our way through how to introduce error raising and handling in concept exercises and exercises in general.  Our stopgap are the instruction appends noted in  #2563,  which are being worked on.

PR #2636, opened 2021-10-07
Add prickle and smolder labels

J08K:
These are some new labels:
- prickle💢 - Label for issues that should be skipped by the auto-responder (to be added, _almost done_)
- smolder🍲 - Label for issues or PRs that are a long-term idea, or lower priority for the track.

Discussed with: @BethanyG

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -319,3 +319,11 @@
 - name: "❔question❔"
   description: "I have a proposal or question about things, but nothing is broken."
   color: "293028"
+
+- name: "prickle💢"
+  description: "This issue should be skipped by the auto-responder."
+  color: "FFFFFF"
+
+- name: "smolder🍲"
+  description: "Lower-priority and/or future idea"
+  color: "8B0000"


SaschaMann:
For future reference: This file is not meant to be edited manually as indicated by the big warning on top: https://github.com/exercism/python/blob/9772a9c54f854b307ce4e9562a349cbfcc760a66/.github/labels.yml#L1-L14

Doing so will lead to the labels being removed/overwritten eventually.

BethanyG:
@SaschaMann  -  Yup.  The fact that these were to the main file slipped through the cracks originally.  🤦🏽‍♀️ .  Thank you also for the PR that adds them to the addendum.  😄   I'll take a look when I have more of a chance tonight my time .... I may want to remove one of them after all.

SaschaMann:
>  The fact that these were to the main file slipped through the cracks originally

At some point these should be codeowner'd to prevent this, I think. It's super easy to miss when reviewing.

PR #2637, opened 2021-10-08
links: use https for wikipedia

imba-tjd:


imba-tjd:
I suppose the more https the better. Anyway you are the boss.

PR #2640, opened 2021-10-08
[Making-the-grade] Update instructions.md

gregoryhills:
I think there's a typo in "Calculating Letter Grades", since you're asking to return 4 numbers, you're asking for thresholds for grades A-D not A-F


J08K:
Yep, I checked it. It does make more sense this way! Great catch indeed, @gregoryhills!

PR #2646, opened 2021-10-11
[Perfect Numbers] Clarify instructions

gregoryhills:
The original instructions read: "Implement a way to determine whether a given number is perfect. Depending on your language track, you may also need to implement a way to determine whether a given number is abundant or deficient."

I've clarified that you should determine whether the number is abundant or deficient as well.


BethanyG:
Hi @gregoryhills 

Thanks for submitting this.  The introduction and instructions for `perfect numbers` in Python are auto generated from specifications found in  the [`problem specifications`](https://github.com/exercism/problem-specifications) repo, so adjustments to them should be brought up and discussed there first, before they're (potentially) rolled out across tracks.  We'd only PR here if the changes were judged to be Python-specific.

Thanks for understanding.

gregoryhills:
Makes sense, thanks for contexzt

Issue #2647, opened 2021-10-12
[Guido's Gorgeous Lasagna] instructions.md showing different var names


goatthinking:
In the instruction page for "Guido's Gorgeous Lasagna" concept exercise a variable has two different names within the document.

When instructing on how to make the function it says to use: `elapsed_bake_time`
* https://github.com/exercism/python/blob/efc213ac959a76e4f204214ffb98e185227960ff/exercises/concept/guidos-gorgeous-lasagna/.docs/instructions.md?plain=1#L38

However in the next section it where it has you add documentation it uses: `actual_bake_time`
* https://github.com/exercism/python/blob/efc213ac959a76e4f204214ffb98e185227960ff/exercises/concept/guidos-gorgeous-lasagna/.docs/instructions.md?plain=1#L51

Suggest renaming one so that they match.

BethanyG:
@goatthinking -- thanks for the catch!  We've fixed this in #2651.

PR #2651, opened 2021-10-13
Fix the code sample in the instructions to use the same var name as other references

IsaacG:
https://github.com/exercism/python/issues/2647

The sample snippet in the instructions references `actual_bake_time` while all the other references to this parameter call it `elapsed_bake_time`.


BethanyG:
@IsaacG   -- looks like you picked up a change to another exercise as well?  There appear to be merge conflicts here.....

Could you try fetching and rebasing?  Thanks!

IsaacG:
My bad. I messed up and branched from the wrong point. This should be fixed ... I hope.

PR #2664, opened 2021-10-16
Change logic in line 88

szychy:
Logic in line 88 changed as (list[0] + list[-1] / 2) != (sum(list) / len(list)) (ergo: 4.5 != 4.0)

### DIFF
--- a/exercises/concept/card-games/lists_test.py
+++ b/exercises/concept/card-games/lists_test.py
@@ -85,7 +85,7 @@ class CardGamesTest(unittest.TestCase):
             ([1, 2, 3], True),
             ([2, 3, 4], True),
             ([2, 3, 4, 8, 8], True),
-            ([1, 2, 4, 5, 8], True),
+            ([1, 2, 4, 5, 8], False),
         ]
 
         for variant, (hand, same) in enumerate(data, start=1):


BethanyG:
Hi @szchelkowski 👋🏽 

Thank you for submitting this PR.  I think you may be misunderstanding the exercise instructions.  I've copied them below. 

(_Bolded emphasis mine_)

> **5. Alternate Averages**
> In Black Joe, speed is important. Elyse thinks she can "shortcut" the traditional calculation to get an average-like number by taking the average of the first and last card values in the hand - or by using the "middle" card value (**the median**) of the hand. She is hoping **one of** these calculations might come close to the average, but be quicker to calculate during game play. She'd like you to create a function to test her theory out.
> 
> Implement a function `approx_average_is_average(<hand>)` that returns a Boolean. The function should return `True` if **ONE** (or both) of Elyse's approximations yields the same value as calculating the "full" average of a hand. The function should return `False` if **NEITHER** approximation formula yields the same value as the "full" average. For the sake of a simple median, we are going to assume the hand always has an odd number of card values.

The method should return true if  (list[0] + list[-1] / 2) == 4.0 **OR** median == 4.0.  It does not have to be both.  

The median of `[1, 2, 4, 5, 8]` is 4.0.

Please let me know if you have any questions or issues.

szchelkowski:
Hi @BethanyG

Yes, I have completely overlooked the median value part. Sorry for making you waste time on this pull request!

Have a great day!

PR #2678, opened 2021-10-18
[Tisbury Treasure Hunt]:Potential fix for bug 2644.

BethanyG:
This is a very small thing and may not work.  Hints for Task 5 are not displaying on the website UI for this exercise.  Noticed that it was the only header with an `&`.  Removed the `&` in hopes that that was the issue with the hints not displaying.  

**Note** -- this may not work, and it may be something else.


J08K:
Weird, on the website it does say it is updated, but clicking on the Hints still displays the old message. AKA still not working sadly...

Issue #2605, opened 2021-09-27
[Card Games] Refactor Tests and Add Sum() and Len() to Exercise


BethanyG:
Per notes in [PR2590](https://github.com/exercism/python/pull/2590), this exercise needs to have its tests and documents refactored.  In particular:

- [ ] Refactor tests to include additional test scenarios and variants (with `subTest`) as needed.
- [ ] Refactor `exemplar.py` to omit the use of `loops`, since `loops` is not a prerequisite of this exercise, so cannot be used.
- [ ] Edit `introduction.md` and `instructions.md` to cover the use of built-ins `len()` and `sum()` (_to be used in place of `loops` in the `exemplar.py`_)

One suggestion to omit loops in `exemplar.py` is to `return sum(hand)/len(hand)` in `card_average()`.

BethanyG:
@pranasziaukas  -- I cannot seem to assign you to this (_instead I will assign myself_), but it is all yours to work on.  And thank you very much for taking it on! 🌟

pranasziaukas:
I created a PR to bring the tests to a better shape.

When it comes to getting rid of `loops`, my understanding is the following:
- `instructions.md` looks good to me as it doesn't ask to loop over anything
- `hints.md` has broken links AFAICT, plus
  - "iterating over a `list`" part needs to be removed
- `introduction.md`
  - mentions iterations and `for` loops in multiple places, including examples
  - doesn't have anything about `sum()` or `len()` 
- `design.md`
  - has `for` loops amongst learning objectives
  - has `len()` and similar builtin functions as out of scope

How to align it all with the big picture is a bit of a brain twister and I don't think I will be tackling it in the immediate future -- it can be taken up by anyone as a separate PR.

BethanyG:
@pranasziaukas -- understandable.  It is indeed a mess -- and you don't have to clean it up if you don't want to.  😄   I can add the rewrite parts to my to do list.

PR #2682, opened 2021-10-20
[Card Games]: Omitting `loops` concept from exemplar.

J08K:
This PR would closes issue #2605.


J08K:
@BethanyG Well, I wanted to have a last look over it before actually submitting the PR. I think it looks good as well, I'll change it to a regular PR!

PR #2713, opened 2021-10-31
Fix little sisters essay parameter order

### LINKED ISSUE #2712: [My Little Sisters Essay] Parameter order
On my little sisters essay, the parameter naming order for `replace_word_choice()` is inconsistent.

The template has:
```py
def replace_word_choice(..., old_word, new_word):
    """
    ...
    :param new_word: str replacement word
    :param old_word: str word to replace
    ...
    """
    ...
```
While the exemplar solution is:
```py
def replace_word_choice(..., new_word, old_word):
    """
    :param new_word: str replacement word
    :param old_word: str word to replace
    """
    ...
```
The template is clearly correct in the function definition, while all of the other 3 orders are wrong.

I'd like to fix this if possible!

mathstrains21:
Resolves #2712

### DIFF
--- a/exercises/concept/little-sisters-essay/.meta/exemplar.py
+++ b/exercises/concept/little-sisters-essay/.meta/exemplar.py
@@ -29,14 +29,14 @@ def clean_up_spacing(sentence):
     return clean_sentence
 
 
-def replace_word_choice(sentence, new_word, old_word):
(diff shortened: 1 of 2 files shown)


IsaacG (approved):

BethanyG (approved):
Nice catch!  Thanks for submitting this.  🌟

PR #2724, opened 2021-11-03
Removed excess `<br>` from Greeting

BethanyG:


J08K (approved):
This should resolve @iHiD's comment in the `maintaining-python` Slack channel. Thanks Bethany!

iHiD:
Thanks. I think some of the others can go too if you check the comment I linked to in Slack. There were a few stray ones I think, as I think GH adds `<p>` for the markdown and then the `<br/>` too. :)

PR #2731, opened 2021-11-07
Make instructions of task #3 more clear

safwan-samsudeen:
When I was doing this exercise, I had a doubt about the spelling rule: "If the root word originally ended in a consonant followed by a 'y', then the 'y' was changed to to 'i'.". Does that mean I have to check that the second last letter (without the "ness") is a consonant before replacing the "i" with a "y"? 

No, said my mentor (and the tests pass when I don't check the second last letter), which is why I've added an instruction not to care about the second last letter. If we do need to consider that, we need to both add a new test and clarify this instruction page. Also, the spelling rule seems to have a repetition

This is my first contribution to any open source project... so my apologies if my edits to the doc is not up to standard. Please edit them for more clarity.

### DIFF
--- a/exercises/concept/little-sisters-vocab/.docs/instructions.md
+++ b/exercises/concept/little-sisters-vocab/.docs/instructions.md
@@ -61,7 +61,7 @@ Implement the `make_word_groups(<vocab_words>)` function that takes a `vocab_wor
 `ness` is a common suffix that means _'state of being'_.
  In this activity, your sister needs to find the original root word by removing the `ness` suffix.
   But of course there are pesky spelling rules: If the root word originally ended in a consonant followed by a 'y', then the 'y' was changed to to 'i'.
- Removing 'ness' needs to restore the 'y' in those root words. e.g. `happiness` --> `happi` --> `happy`.
+ Removing 'ness' needs to restore the 'y' in those root words. e.g. `happiness` --> `happi` --> `happy`. In this exercise, you can ignore the second last letter point, though - just replace last letter 'i' with 'y' when removing the suffix.
 
 Implement the `remove_suffix_ness(<word>)` function that takes in a word `str`, and returns the root word without the `ness` suffix.
 


BethanyG:
@safwan-samsudeen - Hi!  👋🏽   Welcome to the Python track!

Thank you for submitting this PR.  🌟 

However, I am not sure I am following your explanation/confusion. 

@J08K  - can you look this over for me?  I am really not following what needs to be clarified here, and I'm worried I am missing something.  _Thanks._

safwan-samsudeen:
I'm sorry. Like, I said my first contribution ever, so I was unclear.

What I mean is, the exercise currently seems to state that, you should change the "i" to a "y" _only if_ "the root word originally ended in a consonant". In which case the submitted solution needs to check if the second last letter is a consonant. 

To explain with the given example, when the function is passed `"happiness"`, it should not return `"happi"`, it should replace `"happy"`. _Because_ the second last letter, "p", is a consonant. But in my solution, which does not check that the second last letter is a consonant, the tests pass.

So what I'm saying is, we either need to 

1. Add a new test to check for this, passing in a word whose second-last letter is a vowel, and the last letter is a "y", or
2. Remove the if clause in the "pesky" spelling rule, or
3. Explicitly state that you can ignore the second-last letter. 

I would suggest solution 2 - removing the if clause, but I felt that instead of immediately removing what's been written, I can add something, and make further changes if needed. So I implemented solution 3.

J08K:
Hi both,
Thank you for your first Pull Request!
I looked through this PR a few minutes this weekend, but I wasn't sure what was meant either, sorry that I didn't reply earlier. With "second last letter", I assume you mean "second to last letter"? I found [this PDF](https://www.redwoods.edu/Portals/43/Handouts/What%20is%20a%20suffix.pdf) which explains the language rules of adding the `-ness` suffix. 

Do you mean that we should check if the second to last letter is a consonant? I think your proposed `solution 1` is probably better than adding more instructions. 

I do want to note, this concept exercise is about `strings` not about the English language. If this issue still lets the solutions that are `correct` pass, and it still allows students to pass and learn the concept, I do not think of this as a big issue. 

So, I, personally, do not think we should add more instructions. I try to keep as little lines in the instructions as possible, but still keep it clear and concise. These instructions might confuse the students as it did me at first. It might make them worry about the wrong thing, instead of focusing on the concept `strings`.

Anyway, I wonder if many other students have/had problems with this exercise, this is my first time hearing of this. Great work on your first PR, though! 

What are your thoughts on this @BethanyG, I hope I explained it a bit?

safwan-samsudeen:
Hi,

Don't you feel solution 2 is better? I agree with you - the exercise is about strings, and not about English. So, we just remove the bit about the second to last letter (sorry for the unclear version in the above posts) from the instructions.

I feel adding a test would not contribute much to the students' understanding, and perhaps time-consuming - is there any word in the language that has a second-to-last letter of a vowel _and_ the last letter as "y"?

BethanyG:
`fey`, `coy`, and `gay` are words that follow the no-i conversion when it comes to `ness`.  
`fey` --> `feyness`
`coy` --> `coyness`
`gay` --> `gayness`

But there is a problem, because English has many many irregularities.  So we also have:

`wry` --> `wryness`
`dry` --> `dryness`
`sly` --> `slyness`

and then there is `day`, which becomes `dailyness` ... but let us not talk about that.  😆 

@J08K  -- I agree, adding more ... words ... doesn't feel like it would help this situation. This is an exercise about string manipulation -- not the English language and all its various warts.  

But given that we do have three `vowel + y` examples, we could add some additional tests, if we want to have counter-examples to the rule.  Thoughts?

J08K:
I suggest that we make an issue about this to talk it over. If we do edit the test files, we will almost definitely be breaking some of the 939 (as of this moment) submissions that have not implemented this feature. Also, it might be handy to add some text with `"Lookout for the exceptions!"`, I bet not a lot of people will think of them right out of the gate.

But, we might want to keep what we have now, if people do not check for the consonant, and it does not break the solutions for the students that do check for the second-to-last letter. It might just be better to have the students focused on the `concept` itself? 

@BethanyG @safwan-samsudeen what are your thoughts?

safwan-samsudeen:
Hey,

I think I'm expressing myself ambiguously. I do not mean to add a test. This would both break a lot the current solutions AND distract the student from the main purpose of string manipulation.

What I'm saying is, add an instruction to make it clear that the second to last letter rule doesn't matter. Currently, there is ambiguity. The wide majority of solutions do not check for the second-to-last letter as a consonant. But there are a few solutions which _do_ - please see [this solution](https://exercism.org/tracks/python/exercises/little-sisters-vocab/solutions/mburazin) which checks that the second to last letter is not a vowel. I had the same doubt, too.

So it seems like the confusion, while not common, does exist. Which is why I suggest we add an instruction like the current commit has - though we definitely need to improve the wording so that it's more clear.

J08K:
Hi there @safwan-samsudeen,
I do not think it is a good idea that we change anything about this exercise. Adding more instructions would likely confuse the student, and thus make them focus on the wrong thing. Even if someone would follow the exclusions for `-ness` they would still pass the test.
If we were to add new tests to this exercise, that would also mean we would have to add something to the instructions about those exclusions.
And all of these are supported by the fact that this exercise is not about the English Language, but it is about `string-methods` in Python. I therefore will be closing this PR. 

I do want to thank you for the time you put into this, even though we can't do anything about it. Sadly this is a compromise we have to make, between accuracy of the context and accuracy of the concept itself. 
Have a nice day!

PR #2732, opened 2021-11-08
Remove duplicate instructions header

safwan-samsudeen:
There is both a first level and a second-level header with the text "Instructions". This is not a big problem in the exercise description, but in the editor, the first task is shown as "Instructions", which is not right. I've removed the duplicate header so that Task 1 will be shown correctly as "Tracking Poker Rounds".

### DIFF
--- a/exercises/concept/card-games/.docs/instructions.md
+++ b/exercises/concept/card-games/.docs/instructions.md
@@ -1,7 +1,5 @@
 # Instructions
 
-## Instructions
-
 Elyse is really looking forward to playing some poker (and other card games) during her upcoming trip to Vegas.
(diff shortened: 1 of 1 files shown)


safwan-samsudeen:
Here's a photo of the editor:

image

Clicking on "Stuck? Reveal Hints" pops up an empty dialog.

image

J08K:
Hey there, 
Yes I doubt that this was intended. I will ask @BethanyG to check this over first though, maybe it was intended??
Anyways, I'll let this PR run the GH Actions checks.
Thanks for this!

BethanyG (approved):
@safwan-samsudeen  - Thank you very much for this PR!  And @J08K is correct -- that wasn't intentional, and I am not sure how it slipped by.  Thanks for catching it.  🌟

PR #2736, opened 2021-11-09
[Currency-Exchange]: In-line links to  reference links

BethanyG:
Changed in-line links to reference links.  Hope that this also fixes hint problem.  But either way, it needed to be done to conform to the markdown spec.


J08K (approved):
Nice!

PR #2737, opened 2021-11-09
[Tisbury Treasure Hunt]:  In-line links to Reference Links

BethanyG:
Changed in-line links to reference links.  With some luck , this will fix the hints problem.  But even if it doesn't, the was needed to bring the doc up to the markdown spec.


J08K (approved):
Sounds like the treasure was hidden in `problem-specifications`!

PR #2739, opened 2021-11-09
[Anagram] : Update from Problem Specifications

BethanyG:
Pulled down updates from problem-specifications.  

Regenerated `tests.toml` and test file.


J08K (approved):
Surely this will win an ana-_grammy_! 😉

PR #2744, opened 2021-11-09
[Forth]: Update from Problem Specifications

BethanyG:
Pulled down updates from problem-specifications.
Regenerated `tests.toml` and test file.


J08K (approved):
You are really **forth**coming, Bethany!

PR #2762, opened 2021-11-11
removed redundant test

xle263:
A test with similar test data was removed because it has no practical use


J08K:
Hi there!
This is a practice exercise, which means that these tests are autogenerated. What they should auto-generate is stated in the [Problem-Specification Repo](https://github.com/exercism/problem-specifications), so feel free to post your issue in there, since it is not up to us.
Thank you, and have a nice day!

Issue #2644, opened 2021-10-10
[Tisbury Treasure Hunt] Hints aren't showing up on web


gregoryhills:
When I look at the hint file on GitHub for Tisbury Treasure Hunt, I see a bunch of detailed hint information.
However, when I try to pull up the hints in the web interface, the informative hints present in GitHub do not appear.
I'm running Chrome on Mac OS BigSur.  

[Link to hints on Github](https://github.com/exercism/python/blob/main/exercises/concept/tisbury-treasure-hunt/.docs/hints.md)
Github screenshot:
![2021-10-10_14-30-28](https://user-images.githubusercontent.com/205414/136709061-4867c636-80bc-4376-9ff4-8d6149b0d055.png)

[Link to exercise on exercism](https://exercism.org/tracks/python/exercises/tisbury-treasure-hunt/edit)
On exercism.org
![2021-10-10_14-30-10](https://user-images.githubusercontent.com/205414/136709066-e57044b3-11c7-44e1-8e06-f2ef5ac1fe7b.png)

BethanyG:
@J08K  -- left you a note on Slack.  Can you take a look in your account?  *Thank you so much!*  

I can't see anything obviously wrong with the markdown, that doesn't mean that there isn't a problem.

J08K:
Yep, I am getting the same issue on task 5! This is ONLY on task 5, which is weird because nothing about the raw `markdown` in the task 5 header is different than the rest of the file.

This is the same for `currency exchange` although on that exercise NONE of the hints are working (To be fair, this exercise is pretty cursed).

I think it might be a website error? But I am not sure, will investigate further.

BethanyG:
The issues with `currency exchange` hints is most likely resolved by PR #2677.  
There was a missing top-level Markdown header.

The Tisbury hints (I hope!) might be fixed by #2678 -- but we won't know until it's merged and pushed to the site.

BethanyG:
Fixed (FINALLY) by #2763

PR #2763, opened 2021-11-11
[Currency Exchange & Tisbury]:  Fixed Hint Files

BethanyG:
Changed the hint titles to match the instruction task titles so they would show on the website.

### DIFF
--- a/exercises/concept/currency-exchange/.docs/hints.md
+++ b/exercises/concept/currency-exchange/.docs/hints.md
@@ -4,11 +4,11 @@
 
 - [The Python Numbers Tutorial][python-numbers-tutorial] and [Python numeric types][python-numeric-types] can be a great introduction.
 
-## 1. Estimating exchangeable value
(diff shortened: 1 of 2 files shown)


BethanyG:
**Should** fix #2644 for reals this time.  

Turns out, the website is matching hints to tasks on the website via **`titles`** -- so when a `hint` **title** doesn't match its associated task, it gets omitted.  😱   We'll need to check for this in the future.  

There is an issue logged under `exercism/exercism` to change the behavior of the website to match hints by index ([#6105](https://github.com/exercism/exercism/issues/6105)) but we should still check for a match between hint titles and task titles for the time being.  (_I would argue we should probably always check, as it is most likely confusing to have the hint titles NOT match the tasks they belong to._)

J08K (approved):
Looks good! 
But yes we will need to check for this in the future. Might want to check if there are some other exercises where hints don't show up? I'll look into it.

PR #2771, opened 2021-11-13
[Queen-Attack]: Add instructions for error checking that the two queens are not the same

safwan-samsudeen:
The tests check that the two queens are not in the same place - but the instructions don't specify that. So I'm adding that.

### DIFF
--- a/exercises/practice/queen-attack/.docs/instructions.append.md
+++ b/exercises/practice/queen-attack/.docs/instructions.append.md
@@ -4,7 +4,7 @@
 
 Sometimes it is necessary to [raise an exception](https://docs.python.org/3/tutorial/errors.html#raising-exceptions). When you do this, you should always include a **meaningful error message** to indicate what the source of the error is. This makes your code more readable and helps significantly with debugging. For situations where you know that the error source will be a certain type, you can choose to raise one of the [built in error types](https://docs.python.org/3/library/exceptions.html#base-classes), but should still include a meaningful message.
 
-This particular exercise requires that you use the [raise statement](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement) to "throw" `ValueErrors` if the `Queen` constructor is passed numbers that are negative, or numbers that are not a valid row or column.  The tests will only pass if you both `raise` the `exception` and include a message with it. You also should `raise` an `ValueError` if both the queens passed to check whether they can attack are on the same location.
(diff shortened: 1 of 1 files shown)


IsaacG:
Much of Exercism is set up with the requirements not being spelled out in the prose and the tests being used to inform the requirements. Exercism takes a test-driven development approach to the exercises. This is intentional.

safwan-samsudeen:
OK, thanks!

BethanyG:
@IsaacG - Your comment was not a helpful (and also misleading), since there is indeed a track-specific test for the scenario @safwan-samsudeen mentions **with** a specific error handling message that was missed as an example in the `instructions-append.md`.

@safwan-samsudeen  -- thank you very much for catching this error!  I am re-opening this PR.  😄

safwan-samsudeen:
OK, I've reopened it. Thanks!

exercises/practice/queen-attack/.docs/instructions.append.md
```diff
@@ -4,7 +4,7 @@
...
 
-This particular exercise requires that you use the [raise statement](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement) to "throw" `ValueErrors` if the `Queen` constructor is passed numbers that are negative, or numbers that are not a valid row or column.  The tests will only pass if you both `raise` the `exception` and include a message with it.
+This particular exercise requires that you use the [raise statement](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement) to "throw" `ValueErrors` if the `Queen` constructor is passed numbers that are negative, or numbers that are not a valid row or column.  The tests will only pass if you both `raise` the `exception` and include a message with it. You also should `raise` an `ValueError` if both the queens passed to check whether they can attack are on the same location.
```
BethanyG, line 7:
```suggestion
This particular exercise requires that you use the [raise statement](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement) to "throw" `ValueErrors` if the `Queen` constructor is passed numbers that are negative, or numbers that are not a valid row or column.  The tests will only pass if you both `raise` the `exception` and include a message with it. You also should `raise` a `ValueError` if both the queens passed to the `can_attack()` method are on the same location.
```

BethanyG (approved):
@safwan-samsudeen  - Thank you for reopening this!  I did make one suggestion for rephrase below.  Accept or reject as you see fit.  Otherwise, I think this is good to go.  🌟

safwan-samsudeen:
Hi! Thanks, please merge it now if possible.

Issue #2726, opened 2021-11-03
Local MD doc links break on the website


IsaacG:
The MD files link locally using names like `[TOOLS](./TOOLS.md)` which works when renders on Github but generates invalid links when displayed on the website. On the web, the tools link ought to be `./tools`

Related/blocks https://github.com/exercism/python/issues/2721

Issue #2721, opened 2021-11-03
Tools.md doc is not published to the website


IsaacG:
@BethanyG did some major rework of the https://github.com/exercism/python/blob/main/docs/TOOLS.md file in https://github.com/exercism/python/pull/2611/files

However, while the [tests](https://exercism.org/docs/tracks/python/tests) page shows up on the website, the tools page does not. I believe that's due to the tools page lacking an entry in https://github.com/exercism/python/blob/main/docs/config.json

The docs repo has 7 markdown files. The config has 4 entries. At least one, if not three, of those files should probably be added.

IsaacG:
Side note: the MD files link locally using names like `[TOOLS](./TOOLS.md)` which works when renders on Github but generates invalid links when displayed on the website.

BethanyG:
It was actually @J08K who did that massive amount of work.  I did the CONTRIBUTING and READMEs.  😄 
Let's get the links logged in a separate issue and corrected before we PR anything to show on the site.

PR #2776, opened 2021-11-15
[Docs]: Cleanup, Push to site, and Spell check

J08K:
This would fix both #2721 and #2726. Also I cleaned up the docs a little bit and fixed some weird stuff and did a spell check.


BethanyG (approved):
Nice work!  Thank you for this.  🌟  🌈

PR #2784, opened 2021-11-16
[Making the Grade]: Clarified Grading Intervals & Added Detail to Examples for  Task 4

### LINKED ISSUE #2773: [Making the Grade]: Confusing Grading Intervals - American Cultural Bias
The Making the Grade Exercise: https://exercism.org/tracks/python/exercises/making-the-grade/

Task 4:
> The teacher you're assisting likes to assign letter grades as well as numeric scores. Since students rarely score 100 on an exam, the "letter grade" lower thresholds are calculated based on the highest score achieved, and increment evenly between the high score and the failing threshold of <= 40.

> Create the function letter_grades() that takes the "highest" score on the exam as a parameter, and returns a list of lower score thresholds for each letter grade from "D" to "A".

```
>>> letter_grades(highest=100)
[41, 56, 71, 86]

>>> letter_grades(highest=88)
[41, 53, 65, 77]
```

The code itself has this comment:
```
    """
    :param highest: integer of highest exam score.
    :return: list of integer score thresholds for each F-A letter grades.
    """
```

This is extremely unclear and I have zero idea what it's asking for. The problem is that it's presupposing knowledge of a certain (American?) culture's grading system.

* Why does the comment say F-A but the thing says A-D.
* What order is correct?
* Where is E going?
* What's a failing grade?
* What threshold is for what letter?

I can infer some of these things, but this is meant to be a concepts exercise teaching something. It shouldn't require anything complex like that.

Note: This is NOT a request for support; I can easily code a solution; the issue is that the exercise is extremely unclear such that I have no idea what is being requested in the first place (I got past it by simply returning the test's expected data). And if I don't I wouldn't be surprised if others don't either.

BethanyG:
Closes #2773.

### DIFF
--- a/exercises/concept/making-the-grade/.docs/instructions.md
+++ b/exercises/concept/making-the-grade/.docs/instructions.md
@@ -55,11 +55,29 @@ This function should return a `list` of all scores that are `>=` to `threshold`.
 The teacher you're assisting likes to assign letter grades as well as numeric scores.
 Since students rarely score 100 on an exam, the "letter grade" lower thresholds are calculated based on the highest score achieved, and increment evenly between the high score and the failing threshold of **<= 40**.
 
-Create the function `letter_grades()` that takes the "highest" score on the exam as a parameter, and returns a `list` of lower score thresholds for each letter grade from "D" to "A".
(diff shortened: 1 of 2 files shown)


IsaacG (approved):

J08K (approved):
I think that will clear something up for the students that really do not understand. 🤞

Issue #2582, opened 2021-09-14
Stubs and examples can use linting


IsaacG:
This is a superset of https://github.com/exercism/python/issues/2579 (long lines)

`pylint` flags all sorts of issues that ought to be cleaned up.

```
» find exercises -name \*.py ! -name \*_test.py -exec pylint -d C0116,W0613,C0114,C0115,C0103,R0801,W0707,W0107,C0301,R0903,W0511 {} +
```

Logging this is a "on hold, revisit later" type issue for now.

BethanyG:
..and as a superset of #2579, it has a hold for a similar reason.  We need to have the track a bit more stable and a bit more filled out with concepts and concept exercises before we loop back for a discussion about `practice` stub files, line length, or any PyLint whining.  We may also choose to not enforce some PyLint rules.  That is still TBD.

Additionally - `example` files are more or less exempt from formal linting rules.   It isn't that "anything goes" -- but  [the spec](https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md#file-example-implementation) says nothing about them being any sort of **exemplar**, and they are at most a "minimum implementation" for use in CI and testing.  It is a very low priority that we look through them all.

J08K:
Whoops, closed the wrong thing 💙

J08K:
Closed by 8 PRs, linted all stubs and exemplars.

PR #2787, opened 2021-11-17
[Linting]: Lint practice stub files (1 / ?) 

J08K:
According to issue #2582.

### DIFF
--- a/exercises/practice/affine-cipher/.meta/example.py
+++ b/exercises/practice/affine-cipher/.meta/example.py
@@ -1,4 +1,4 @@
-BLKSZ = 5
+BLOCK_SIZE = 5
 ALPHABET = 26
 
 
@@ -33,8 +33,8 @@ def translate(text, a, b, mode):
 
 def encode(plain, a, b):
     cipher = translate(plain, a, b, 0)
-    return " ".join([cipher[idx:idx + BLKSZ]
-                     for idx in range(0, len(cipher), BLKSZ)])
+    return " ".join([cipher[idx:idx + BLOCK_SIZE]
+                     for idx in range(0, len(cipher), BLOCK_SIZE)])
 
 
 def decode(ciphered, a, b):
--- a/exercises/practice/affine-cipher/.meta/example.py
+++ b/exercises/practice/affine-cipher/.meta/example.py
@@ -2,42 +2,40 @@ BLOCK_SIZE = 5
 ALPHABET = 26
 
 
-def mod_inverse(a, alphabet):
-    a = a % alphabet
+def mod_inverse(a_key, alphabet):
+    a_key = a_key % alphabet
     for idx in range(1, alphabet):
-        if (a * idx) % alphabet == 1:
+        if (a_key * idx) % alphabet == 1:
             return idx
     return 1
 
 
-def translate(text, a, b, mode):
-    inv = mod_inverse(a, ALPHABET)
-    if inv == 1:
-        raise ValueError("a and m must be coprime.")
+def translate(text, a_key, b_key, mode):
+    inverse = mod_inverse(a_key, ALPHABET)
+    if inverse == 1:
+        raise ValueError('a and m must be coprime.')
 
     chars = []
-    for c in text:
-        if c.isalnum():
-            orig = ord(c.lower()) - 97
-            if orig < 0:
-                chars.append(c)
+    for character in text:
+        if character.isalnum():
+            origin = ord(character.lower()) - 97
+            if origin < 0:
+                chars.append(character)
                 continue
             if mode == 0:
-                new = (a * orig + b) % ALPHABET
+                new = (a_key * origin + b_key) % ALPHABET
             elif mode == 1:
-                new = (inv * (orig - b)) % ALPHABET
+                new = (inverse * (origin - b_key)) % ALPHABET
             chars.append(chr(new + 97))
 
-    return "".join(chars)
+    return ''.join(chars)
 
 
 def encode(plain, a, b):
     cipher = translate(plain, a, b, 0)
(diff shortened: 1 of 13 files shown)


exercises/practice/alphametics/.meta/example.py
```diff
@@ -129,14 +130,14 @@ def solve(an):
...
-    nzchars = set([w[-1] for s in fullexp for w in s])
+    nzchars = {w[-1] for s in fullexp for w in s}
     # initialize the lists for digit ranks
     unzchars = []  # non-zero letters unique at level
     uokzchars = []  # zero-allowed letters unique at level
     uchars = []  # all letters unique at level
     tchars = []  # all letter with multipliers per level
-    for i in range(maxp):
-        tchars.append(dict())
+    for _ in range(maxp):
```
BethanyG, line 139:
```suggestion
    for number in range(maxp):
```

exercises/practice/atbash-cipher/.meta/example.py
```diff
@@ -1,24 +1,18 @@
...
 
 
 BLKSZ = 5
-trtbl = maketrans(ascii_lowercase, ascii_lowercase[::-1])
+trtbl = str.maketrans(ascii_lowercase, ascii_lowercase[::-1])
 
 
 def base_trans(text):
-    return ''.join([c for c in text if c.isalnum()]).lower().translate(trtbl)
+    return "".join([c for c in text if c.isalnum()]).lower().translate(trtbl)
```
BethanyG, line 9:
```suggestion
    return "".join(character for character in text if character.isalnum()]).lower().translate(trtbl)
```

exercises/practice/beer-song/.meta/example.py
```diff
@@ -1,19 +1,19 @@
...
+        results.extend(verse(idx))
+        if idx > start - take + 1:
+            results.append("")
     return results
 
 
 def verse(number):
     return [
-        ''.join([
+        "".join([
```
BethanyG, line 12:
```suggestion
        "".join(
```

exercises/practice/beer-song/.meta/example.py
```diff
@@ -1,19 +1,19 @@
...
             "{} of beer on the wall, ".format(_bottles(number).capitalize()),
             "{} of beer.".format(_bottles(number))
         ]),
-        ''.join([
+        "".join([
             _action(number),
             _next_bottle(number)
         ])
```
BethanyG, line 19: Whenever `"".join()` is being used, it's probably better to use a generator expression over a list, since a list creates data in memory.  I can't suggest due to formatting, but I think both of these should omit the list.

exercises/practice/binary-search-tree/.meta/example.py
```diff
@@ -5,7 +5,7 @@ def __init__(self, data, left=None, right=None):
         self.right = right
 
     def __str__(self):
-        fmt = 'TreeNode(data={}, left={}, right={})'
+        fmt = "TreeNode(data={}, left={}, right={})"
```
BethanyG, line 8:
```suggestion
        fmt = 'TreeNode(data={}, left={}, right={})'
```

BethanyG: Our style is single quotes over double, except for docstrings, or where single doesn't work.

exercises/practice/beer-song/.meta/example.py
```diff
@@ -37,11 +37,11 @@ def _next_bottle(current_verse):
...
 def _bottles(number):
     if number == 0:
-        return 'no more bottles'
+        return "no more bottles"
     if number == 1:
-        return '1 bottle'
+        return "1 bottle"
     else:
-        return '{} bottles'.format(number)
+        return f"{number} bottles"
```
BethanyG, line 44:
```suggestion
        return f'{number} bottles'
```

exercises/practice/atbash-cipher/.meta/example.py
```diff
@@ -1,24 +1,18 @@
...
 
 
 def encode(plain):
     cipher = base_trans(plain)
-    return " ".join([cipher[i:i + BLKSZ]
-                     for i in range(0, len(cipher), BLKSZ)])
+    return " ".join([cipher[idx:idx + BLKSZ]
+                     for idx in range(0, len(cipher), BLKSZ)])
```
BethanyG, line 15:
```suggestion
    return " ".join(cipher[idx:idx + BLKSZ]
                     for idx in range(0, len(cipher), BLKSZ))
```

BethanyG (changes requested):
Overall, these are pretty good.  &nbsp; 😄 &nbsp;  But I do have some requested changes:

1.  We're using **single quotes** for everything except doncstrings.  We're only using double quotes when there is a single quote _inside_ the string that would need escaping otherwise.  I stopped recommending after the first few, so there are more than the above that need changing. 
2.  No single letter variable names.   I saw a few in one file that .. got missed?
3.   Strongly prefer using a generator expression inside of `join()` vs having a `list`.  The list gets unpacked anyways, and it just takes up memory.

Thanks for all your work on these.  🌟

exercises/practice/affine-cipher/.meta/example.py
```diff
@@ -1,17 +1,17 @@
...
+ALPHABET = 26
 
 
-def modInverse(a, ALPHSZ):
-    a = a % ALPHSZ
-    for x in range(1, ALPHSZ):
-        if ((a * x) % ALPHSZ == 1):
-            return x
+def mod_inverse(a, alphabet):
+    a = a % alphabet
```
BethanyG, line 6: Suggest a non-single-letter variable name here and elsewhere...

BethanyG (approved):
Thank you so much for doing these!! 🌟  🌈

PR #2790, opened 2021-11-18
[Linting]: Lint practice stub files (3 / ?)

J08K:
According to issue #2582.

### DIFF
--- a/exercises/practice/dnd-character/.meta/example.py
+++ b/exercises/practice/dnd-character/.meta/example.py
@@ -14,8 +14,7 @@ class Character:
 
     def ability(self):
         dice_rolls = sorted(
-            [random.randint(1, 6)
-                for _ in range(4)])
+            [random.randint(1, 6) for _ in range(4)])
         return sum(dice_rolls[1:])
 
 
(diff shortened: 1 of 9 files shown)


exercises/practice/dnd-character/.meta/example.py
```diff
@@ -14,8 +14,7 @@ def __init__(self):
 
     def ability(self):
         dice_rolls = sorted(
-            [random.randint(1, 6)
-                for _ in range(4)])
+            [random.randint(1, 6) for _ in range(4)])
```
BethanyG, line 17: Since this is on two lines, I can't do a suggestion, but since `sorted()` unpacks and it also returns a list, it is not necessary to put a list in as an argument -- it can be a generator expression.

BethanyG (approved):
Nice work!  🌟 
Only one suggestion - but not the end of the world if you don't change it.  😄

exercises/practice/dnd-character/.meta/example.py
J08K: Will do! 💙

PR #2802, opened 2021-11-24
Make examples in instructions.md clearer

davidculley:
For someone unfamiliar with binary, it's not immediately obvious that 19 can be viewed as 3+16. Directly presenting the solution for the exemplary input of 19 without this context might confuse the reader wondering how one arrives at the presented solution before it is explained in the following sentence.

### DIFF
--- a/exercises/practice/secret-handshake/.docs/instructions.md
+++ b/exercises/practice/secret-handshake/.docs/instructions.md
@@ -7,10 +7,10 @@ You and your fellow cohort of those in the "know" when it comes to
 binary decide to come up with a secret "handshake".
 
 ```text
-1 = wink
-10 = double blink
-100 = close your eyes
-1000 = jump
+    1 = wink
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @davidculley 👋🏽 

Thanks for submitting this PR.  

**`Please note:`** (_as stated above_), changes to existing [practice exercise](https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md) instructions/tests or the addition
of new exercises are proposed & discussed in [problem-specifications ](https://github.com/exercism/problem-specifications) first, so that all tracks can potentially benefit.
Once a change is approved there by three maintainers, it becomes available for various language tracks to then implement.

Only Python-specific addendums or tests are PR'd to this repo first.

Please propose/submit your change to [problem-specifications ](https://github.com/exercism/problem-specifications).

_Many thanks!_

PR #2806, opened 2021-11-28
Copy `pytest.ini` to `exercises/shared` for Future Download via Exercism CLI

BethanyG:


J08K (approved):
This should then fix those stoopid lil warnings

PR #2812, opened 2021-11-30
Fixed test by changing order of params

### LINKED ISSUE #2811: [Guido's Gorgeous Lasagna]: Tests or Instructions Might be Wrong
The code doesn't reflect the instructions.

formidablae:
Closes #2811

### DIFF
--- a/exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
+++ b/exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py
@@ -60,7 +60,7 @@ class LasagnaTest(unittest.TestCase):
         for variant, (layers, time, total_time) in enumerate(zip(layer_data, time_data, result_data), start=1):
             with self.subTest(f'variation #{variant}', layers=layers, time=time, total_time=total_time):
                 failure_msg = f'Expected {time} minutes elapsed, but the timing was calculated incorrectly.'
-                self.assertEqual(elapsed_time_in_minutes(layers, time), total_time, msg=failure_msg)
+                self.assertEqual(elapsed_time_in_minutes(time, layers), total_time, msg=failure_msg)
 
     @pytest.mark.task(taskno=5)
     def test_docstrings_were_written(self):


BethanyG:
Hi @formidablae  -- Thanks for submitting this PR!  Unfortunately, the change you've made causes the [`exemplar.py`](https://github.com/exercism/python/blob/main/exercises/concept/guidos-gorgeous-lasagna/.meta/exemplar.py) solution file for this exercise to fail 6 tests, because the positional parameters are now swapped.

Could you please describe in more detail the problem you are trying to fix, and provide screengrabs or pointers to the instruction problem?  This change as it currently stands invalidates the canonical solution (and likely many other solutions on the web site), and can't be merged.

Thanks!

formidablae:
The test script  [`lasagna_test.py`](https://github.com/formidablae/Exercism/blob/main/python/guidos-gorgeous-lasagna/lasagna_test.py)downloaded locally  using `exercism download --exercise=guidos-gorgeous-lasagna --track=python` could be passed only with

```
def elapsed_time_in_minutes(number_of_layers, elapsed_bake_time):
    return preparation_time_in_minutes(elapsed_bake_time) + number_of_layers
```
(see [submitted solution](https://exercism.org/tracks/python/exercises/guidos-gorgeous-lasagna/solutions/exercista)) contrary to the [`exemplar.py`](https://github.com/exercism/python/blob/main/exercises/concept/guidos-gorgeous-lasagna/.meta/exemplar.py) solution:
```
def elapsed_time_in_minutes(number_of_layers, elapsed_bake_time):
    return preparation_time_in_minutes(number_of_layers) + elapsed_bake_time
```

BethanyG:
And here is a solution [submitted in the last hour](https://exercism.org/tracks/python/exercises/guidos-gorgeous-lasagna/solutions/beckojr) that has the exact opposite from your example solution -- but has passed all the tests as they are currently written.

The difficult thing is that function _parameters_ for Python can be named just about anything (they follow the same rules as naming a variable), and only serve as placeholders for the eventual **arguments** that are passed in.  Additionally, _by default_, parameters in Python are _positional_ -- so Python cares more about which object is which position (_and its respective type and value_) than it does about what each is _named_.

So in the case of the solution you've linked to, the function is still doing the correct _math_ based on the _positional values_ -- irrespective of what those values or variables got _named_.  The tests still pass _despite_ the very confusingly named parameters and variables.

They could as well be named `booger` and `bob` -- so long as the function did the correct calculation, and returned the correct value.  Which in the case of [your example](https://exercism.org/tracks/python/exercises/guidos-gorgeous-lasagna/solutions/exercista) it does.

But if I now swap the **_position_** of those parameters in the tests, everything fails.  I've chosen different _names_ for the parameters I pass -- but I am using the same _position_ for them.

`TL;DR` -- we've asked the student to define their own function.  We've hinted at helpful parameter names they (should) be using, but we neither check nor dictate that they get used.  In our tests, we currently check for _outcome_ and not _implementation_ (we may be adding some implementation tests at some point).  

So it's a bit like "return the expected result by any means necessary".

formidablae:
My mistake. You are right. Sorry for the confusion.

PR #2816, opened 2021-12-09
Add indication of return values

PeroPonneso:
Added indication about what value should be returned by the functions.

Note/Opinion: Some sentences indicates the return value as the ending action after a bullet list, in others the information is mixed in the phrase; while this makes it more 'casual' or friendly, it could confuse the reader who's at the start of his/her programming journey

### DIFF
--- a/exercises/concept/chaitanas-colossal-coaster/.docs/instructions.md
+++ b/exercises/concept/chaitanas-colossal-coaster/.docs/instructions.md
@@ -38,7 +38,7 @@ Define the `add_me_to_the_queue()` function that takes 4 parameters `<express_qu
 One person arrived late at the park but wants to join the queue where their friends are waiting.
  But they have no idea where their friends are standing and there isn't any phone reception to call them.
 
-Define the `find_my_friend()` function that takes 2 parameters `queue` and  `friend_name`.
+Define the `find_my_friend()` function that takes 2 parameters `queue` and  `friend_name` and returns the position in the queue of the person's name.
 
 
 1. `<queue>` is the `list` of people standing in the queue.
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @PeroPonneso 👋🏽 

Thanks for submitting this PR!  I agree that the written directions could be confusing -- your change helps with that. 
 
The code examples directly below the task instructions are also intended to **_show_** what should be input/returned from a given function -- and the `docstrings` in the included exercise [stub files](https://github.com/exercism/python/blob/main/exercises/concept/chaitanas-colossal-coaster/list_methods.py) echo these requirements as well.  But not every student will see/read every example, so it doesn't hurt in this case to re-iterate things.  😄 

> _Note/Opinion: Some sentences indicates the return value as the ending action after a bullet list, in others the information is mixed in the phrase; while this makes it more 'casual' or friendly, it could confuse the reader who's at the start of his/her programming journey_

I agree that the instruction patterns for this and other exercises are not consistent  -- but I don't think they need to be -- so long a they covey what's expected. exercism is by default designed for those who already have some programming under their belt, and have a bit of practice with walking through different problem descriptions. 

_That doesn't mean that we can't be clearer or better with our instructions!_ - only that we are expecting that a student **not** be at the start of their programming journey, and that they be familiar with programming terminology and (_somewhat familiar with_) how to break down problems to code solutions. 

We also assume that students will look at both the [stub](https://github.com/exercism/python/blob/main/exercises/concept/chaitanas-colossal-coaster/list_methods.py) and/or [test](https://github.com/exercism/python/blob/main/exercises/concept/chaitanas-colossal-coaster/list_methods_test.py) files to get additional hints/clarification for how to go about solving a a given exercise.

BethanyG (approved):

PR #2831, opened 2021-12-30
Update noun_to_verb to adjective_to_verb

gabo-cs:
I noticed that the function actually extracts an adjective, not a noun (The description itself states _The function should return the extracted **adjective** as a verb._).
Also proposed a new sentence (with an adjective) in the test as it was using a noun.


BethanyG (approved):
@gabo-cs - 🤦🏽‍♀️  is all I can say here.  _How on earth was I blind to this?!!?_  So very happy that you caught this error and submitted a PR!  🌟  🌈   Thank you!  Crazy this has been in production since September - glad it's fixed now.

gabo-cs:
> @gabo-cs - 🤦🏽‍♀️ is all I can say here. _How on earth was I blind to this?!!?_ So very happy that you caught this error and submitted a PR! 🌟 🌈 Thank you! Crazy this has been in production since September - glad it's fixed now.

😅
Awesome!! Glad I could be helpful!! 😊

PR #2832, opened 2022-01-01
[Yacht] Removed Deprecated Python 2.x Stub Note

### LINKED ISSUE #2827: [Yacht] Remove legacy Python 2 Stub Comment
A number of exercises seem to have things explicitly working around Python 2 limitations. For example, anything that's using Enums.
Given Python 2 has been out of support for over 18 months, I'd suggest it's worth remove all of the Python-2-workaround stuff to follow best-practices that use a recent supported version of Python 3 (as of 3 days ago, this became 3.7).

For example: Yacht - https://github.com/exercism/python/blob/main/exercises/practice/yacht/yacht.py

BethanyG:
Fixes / closes #2827.

### DIFF
--- a/exercises/practice/yacht/yacht.py
+++ b/exercises/practice/yacht/yacht.py
@@ -1,17 +1,3 @@
-"""
-This exercise stub and the test suite contain several enumerated constants.
-
-Since Python 2 does not have the enum module, the idiomatic way to write
-enumerated constants has traditionally been a NAME assigned to an arbitrary,
-but unique value. An integer is traditionally used because it’s memory
-efficient.
-It is a common practice to export both constants and functions that work with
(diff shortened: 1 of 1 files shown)


exercises/practice/yacht/yacht.py
```diff
@@ -1,17 +1,3 @@
-"""
```
IsaacG, line 1: Should the stub still contain a module docstring?

BethanyG: No.  Since it referred to Python 2.x, and the equivalent in 3.x would be _either_ constants or `Enums` - and we aren't requiring `Enums`, it's best in this practice exercise to omit the docstring altogether.  See linked issue for discussion.

kotp (approved):

Issue #2833, opened 2022-01-01
[Introduction] Wrong output presented for __doc__attribute of number_to_the_power_of


mlistewnik:
Hi, 
I was reading the Introduction article about Python ([link](https://exercism.org/tracks/python/concepts/basics)) and found a code snippet that I tested in my IDLE Shell (Python 3.8.10).
[Markdown permalink:](https://github.com/exercism/python/blob/main/concepts/basics/introduction.md?plain=1#L181-183)

```

def number_to_the_power_of(number_one, number_two):
    """Raise a number to an arbitrary power.
    
    :param number_one: int the base number.
    :param number_two: int the power to raise the base number to.
    :return: int - number raised to power of second number
    
    Takes number_one and raises it to the power of number_two, returning the result.
    """

    return number_one ** number_two
```
And then:
`print(number_to_the_power_of.__doc__)`

What I get is:
```
Raise a number to an arbitrary power.
    
    :param number_one: int the base number.
    :param number_two: int the power to raise the base number to.
    :return: int - number raised to power of second number
    
    Takes number_one and raises it to the power of number_two, returning the result.
```

But in lines 182-184 there is a different output ([link](https://github.com/exercism/python/blob/7171bdf19d850b09b5a97655be6c2aeaec39da40/concepts/basics/introduction.md?plain=1#L182-L184)):

```
Returns float or int.
       Takes number_one and raises it to the power of number_two, returning the result.
```

Is it correct?

BethanyG:
Closing as fixed in #2837.  Thanks again, @mlistewnik! 💛

PR #2834, opened 2022-01-02
Correct output presented for __doc__ attribute of number_to_the_power_of

mlistewnik:
Fix to issue: https://github.com/exercism/python/issues/2833.


BethanyG:
@mlistewnik  - Nice catch!  Thank you for your careful reading, and for submitting this fix. 🌟   I noticed the same example shows up in https://github.com/exercism/python/blob/main/exercises/concept/guidos-gorgeous-lasagna/.docs/introduction.md.

Would you mind adding the fix to that file to this PR as well?  Thank you so much!

PR #2836, opened 2022-01-02
gigasecond: Make it clear that we pass a datetime object

neduard:
Simple PR, do let me know if it's inappropriate / need to raise an issue.

I wasn't really sure what the gigasecond problem wanted me to do.  This was because "moment" doesn't have a well-defined meaning.  As such, it was only when I ran the tests that I saw it was receiving a `datetime` object.

### DIFF
--- a/exercises/practice/gigasecond/.docs/instructions.md
+++ b/exercises/practice/gigasecond/.docs/instructions.md
@@ -1,6 +1,9 @@
 # Instructions
 
-Given a moment, determine the moment that would be after a gigasecond
+Given a `datetime` representing a moment in time, determine the moment that would be after a _gigasecond_
 has passed.
 
-A gigasecond is 10^9 (1,000,000,000) seconds.
+A _gigasecond_ is 10^9 (1,000,000,000) seconds.
+
+For example, `add(datetime.datetime.utcnow())` might return, depending on when it was called
+`datetime.datetime(2053, 9, 10, 22, 44, 38, 704371)`


kotp:
Notice that what is used in the tests do not necessarily dictate what the solution is.  For Ruby, we use Time, but only because `gigasecond` has a base word "second" and `Time` has a base unit of `seconds` and similarly, if there were a Degree class, it would presumably have a base unit of `second` and so would also work for this.

What I believe you may have demonstrated is some problem solving skill, recognizing that you can use DateTime as the solution, as that is indeed communicated in the tests.  However you may not need to.  Changing this description with the change proposed may unduly encourage a specific solution.

BethanyG:
Hi @neduard 👋🏽  

Thanks for submitting this PR!  A couple of things to note:

1.  This problem description is drawn from a central, cross-track repository, [problem-specifications](https://github.com/exercism/problem-specifications), so changes to the wording need to be proposed there and discussed among the track maintainers.  Only Python-specific _addendums_ (_see the [Clock](https://github.com/exercism/python/tree/main/exercises/practice/clock/.docs) exercise for an example of an instruction append_) to problems like these are proposed in this repository.  

2.   We expect that you will run tests and see failures and infer information from that.  Adding an addendum that points a student to `datetime` short-circuits that process and "gives away" the answer in this case.  As this is a practice exercise, and not a concept exercise, we'd rather not do that.

Given those reasons, I am going to close this PR.  Please let me know if you have questions or issues.  Please also open an issue if you'd like to discuss this further and feel an `instruction append` is needed.  Thanks!

Issue #2839, opened 2022-01-04
[Bowling] Exception-expecting tests pass themselves by accident


RNeilsen:
**(status: ~~I am currently working on learning how to fix, and then fixing, this bug :)~~ [Pull request](https://github.com/exercism/python/pull/2840) submitted!)**

The following tests are supposed to be checking that they get an exception when they ask for the score when a game is incomplete for various reasons:

- `test_an_unstarted_game_cannot_be_scored`
- `test_an_incomplete_game_cannot_be_scored`
- `test_bonus_rolls_for_a_strike_in_the_last_frame_must_be_rolled_before_score_can_be_calculated`
- `test_both_bonus_rolls_for_a_strike_in_the_last_frame_must_be_rolled_before_score_can_be_calculated`
- `test_bonus_roll_for_a_spare_in_the_last_frame_must_be_rolled_before_score_can_be_calculated`

However, all five of these tests make a call to `game.roll()` without passing any arguments. `game.roll()` is supposed to be passed a number of pins on a roll; when that argument is missing, it raises a TypeError. Those five tests, which were _expecting_ an exception to be raised, therefore pass - but the exception that was raised was nothing to do with the game being complete or incomplete (which the tests are supposed to be checking for); the exception was actually just a consequence of the malformed call to the `roll` method.

Comparing the Python test file with the [Ruby test file](https://github.com/exercism/ruby/blob/main/exercises/practice/bowling/bowling_test.rb) for the same exercise shows that the Ruby ones call the `game.score` method, _not_ the `game.roll` method, for those five tests, which indicates that the problem is actually in the [Python test file generator](https://github.com/exercism/python/blob/main/exercises/practice/bowling/.meta/template.j2) (JinJa2 template), specifically this block of code (lines 7-10:

```        {% if case is error_case -%}
        with self.assertRaisesWithMessage(Exception):
            game.roll({{ input["roll"] }})
        {% else -%}
```
In `error_case` type tests, there is no consideration of the `property` field from the canonical-data.json file; so the tests that expect an exception to be raised are always generated with `game.roll()` with no arguments, and thus they always _do_ generate exceptions, and thus they always pass, regardless of what the user has coded. 

Contrast with the corresponding code in the [Ruby test file generator](https://github.com/exercism/ruby/blob/main/exercises/practice/bowling/.meta/generator/bowling_case.rb), lines 31-35, which correctly use the `property` field to decide which function call should be used for these tests.

PR #2840, opened 2022-01-04
Fix error_case test gen to check for 'score' prpty

RNeilsen:
error_case tests were being generated with code that always raised exceptions regardless of the user's code; this was disguised as 'passing' tests since they were expecting an exception anyway.

This change means the generator now checks the 'property' field from the canonical-data.json file and correctly calls the `score` method instead of `roll` for error tests where that is specified.

**Edit to add: [issue link here](https://github.com/exercism/python/issues/2839)**

### DIFF
--- a/exercises/practice/bowling/.meta/template.j2
+++ b/exercises/practice/bowling/.meta/template.j2
@@ -5,8 +5,13 @@
         rolls = {{ input["previousRolls"] }}
         game = self.roll_new_game(rolls)
         {% if case is error_case -%}
+        {% set property = case["property"] -%}
         with self.assertRaisesWithMessage(Exception):
+            {% if property == 'score' -%}
+            game.score()
+            {% else -%}
             game.roll({{ input["roll"] }})
+            {% endif -%}
         {% else -%}
         self.assertEqual(game.score(), {{ case["expected"] }})
         {% endif %}
--- a/exercises/practice/bowling/bowling_test.py
+++ b/exercises/practice/bowling/bowling_test.py
@@ -151,13 +151,13 @@ class BowlingTest(unittest.TestCase):
         rolls = []
         game = self.roll_new_game(rolls)
         with self.assertRaisesWithMessage(Exception):
-            game.roll()
+            game.score()
 
     def test_an_incomplete_game_cannot_be_scored(self):
         rolls = [0, 0]
         game = self.roll_new_game(rolls)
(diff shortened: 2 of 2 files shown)


exercises/practice/bowling/.meta/template.j2
```diff
@@ -5,8 +5,13 @@
         rolls = {{ input["previousRolls"] }}
         game = self.roll_new_game(rolls)
         {% if case is error_case -%}
+        {% set property = case["property"] -%}
         with self.assertRaisesWithMessage(Exception):
+            {% if property == 'score' -%}
+            game.score()
+            {% else -%}
```
IsaacG, line 12: Would it make sense for this to explicitly check the property value?

BethanyG: @IsaacG - The CI for this has not run (as previously asked -- please run the CI and make sure it passes before reviewing.  If you cannot run the CI, please do not review.).

IsaacG: I do believe I cannot run the CI.

BethanyG: Please refrain from reviewing PRs that require a manually triggered CI run then.  Thanks.

BethanyG:
@RNeilsen  - Thanks for submitting this.  😄   I'm bumping the "work" label to large because (_in addition to the troubleshooting)_ you needed to sus out how to alter the JinJa template.  Nice work, you!  🌟 

@IsaacG  - to answer your question, there are only two properties that create errors for this exercise: `roll` and `score`.  `Roll` is the `else` case in the template.  See the note under [bowling](https://github.com/exercism/problem-specifications/blob/main/exercises/bowling/canonical-data.json) in problem specifications.

BethanyG (approved):

BethanyG:
Hi @RNeilsen 👋🏽 

Apologies for being late in realizing this.  Since you've updated the test generation and tests for this exercise, you should add your name to the exercise contributors list here:  https://github.com/exercism/python/blob/main/exercises/practice/bowling/.meta/config.json.  I would do it for you, but you should also get credit for submitting the PR, so.  😄 

Make sure the name you add is your github username, since that's the name that is used to award reputation on exercism.org.  Thanks again for submitting this PR, and for your attention to detail in the `bowling` exercise.  🌈

PR #2844, opened 2022-01-06
[Black Jack]: Comparisons Concept and Concept Exercise Finalization

BethanyG:
Added/Changed/Edited files:

- [x]   config.json
- [x]   exercises/concept/black-jack/.meta/config.json
----
- [x]   concepts/comparisons/introduction.md
- [x]   concepts/comparisons/about.md
____
- [x]   exercises/concept/black-jack/.docs/introduction.md
- [x]   exercises/concept/black-jack/.docs/instructions.md
- [x]   exercises/concept/black-jack/.docs/hints.md
____
- [x]  exercises/concept/black-jack/.meta/exemplar.py
- [x]  exercises/concept/black-jack/black_jack.py
- [x]   exercises/concept/black-jack/black_jack_test.py

### DIFF
--- a/concepts/comparisons/about.md
+++ b/concepts/comparisons/about.md
@@ -32,7 +32,7 @@ For the other numeric types ([complex][complex numbers], [decimal][decimal numbe
 For more information on the rules that python uses for numeric conversion, see [arithmetic conversions][arithmetic conversions] in the Python documentation.
 
 ```python
-import fractions
+>>> import fractions
 
 # A string cannot be converted to an int.
 >>> 17 == '17'
@@ -118,7 +118,6 @@ False
 Container data types (_`lists`, `tuples`, `sets`, `dicts`, etc._) also compare [_lexicographically_][lexographic order] - they are equal if both containers have the same data **and** the same data types (_in the case of `lists` and `tuples`, they must also have the same **ordering**_), unequal otherwise.
 
 ```python
-# Comparing lists
 >>> [1, 2] == [1, 2]
 True
 
@@ -180,6 +179,8 @@ Due to their singleton status, `None` and `NotImplemented` should always be comp
 See the Python reference docs on [value comparisons][value comparisons none] and [PEP8][PEP8 programming recommendations] for more details on this convention.
 
 ```python
+>>> 
+# A list of favorite numbers.
 >>> my_fav_numbers = [1, 2, 3]
 
 >>> your_fav_numbers = my_fav_numbers
@@ -216,6 +217,7 @@ The operators `in` and `not in` test for _membership_.
 For string and bytes types, `<name> in <fullname>` is `True` _**if and only if**_ `<name>` is a substring of `<fullname>`.
 
 ```python
+>>> 
 # A set of lucky numbers.
 >>> lucky_numbers = {11, 22, 33}
 >>> 22 in lucky_numbers
@@ -224,13 +226,8 @@ True
 >>> 44 in lucky_numbers
 False
 
-# A dictionary of employees.
->>> employee = {
-    'name': 'John Doe',
-    'id': 67826,
-    'age': 33,
-    'title': 'ceo'
-    }
+# A dictionary of employee information.
+>>> employee = {'name': 'John Doe', 'id': 67826, 'age': 33, 'title': 'ceo'}
 
 # Checking for the membership of certain keys.
 >>> 'age' in employee
--- a/concepts/comparisons/about.md
+++ b/concepts/comparisons/about.md
@@ -1,199 +1,291 @@
 # About
 
-A [comparison operator][comparisons] in Python (_also called a Python relational operator_), looks at the values of two operands and returns `True` or `False` based on whether the `comparison` condition is met.
-The most common comparison operators are `"<"`, `">"`, `"=="`, `">="`, `"<="`, and `"!="`.
-They all have the same priority (which is higher than that of the Boolean operations).
+A [comparison operator][comparisons] in Python (_also called a Python relational operator_), looks at the _values_ of two [operands][operand] and returns a boolean `True` or `False` if the `comparison` condition is or is not met.
+
+The table below shows the most common Python comparison operators:
+
+
+| Operator | Operation                | Description                                                    |
+| -------- | ------------------------ | -------------------------------------------------------------- |
+| `>`      | "greater than"             | `a > b` is `True` if `a` is **strictly** greater in value than `b` |
+| `<`      | "less than"                | `a < b` is `True` if `a` is **strictly** less in value than `b`    |
+| `==`     | "equal to"                 | `a == b` is `True` if `a` is **strictly** equal to `b` in value    |
+| `>=`     | "greater than or equal to" | `a >= b` is `True` if `a > b` OR `a == b` in value             |
+| `<=`     | "less than or equal to"    | `a <= b` is `True` if `a < b` or `a == b` in value             |
+| `!=`     | "not equal to"             | `a != b` is `True` if `a == b` is `False`                      |
+| `is`     | "identity"                 | `a is b` is `True` if **_and only if_** `a` and `b` are the same _object_  |
+| `is not` | "negated identity"                 | `a is not b` is `True` if `a` and `b` are **not** the same _object_ |
+| `in`     | "containment test"         | `a in b` is `True` if `a` is member, subset, or element of `b`               |
+| `not in` | "negated containment test" | `a not in b` is `True` if `a` is not a member, subset, or element of `b`     |
+
+
+They all have the same priority (_which is higher than that of [Boolean operations][boolean operations], but lower than that of arithmetic or bitwise operations_).
+
+
+## Comparison between different data types
+
+Objects that are different types (_except numeric types_) never compare equal by default.
+Non-identical instances of a `class` will also _**not**_ compare as equal unless the `class` defines special [rich comparison][rich comparisons] methods that customize the default `object` comparison behavior.
+For (much) more detail, see [Value comparisons][value comparisons] in the Python documentation.
+
+Numeric types are (mostly) an exception to this type matching rule.
+An `integer` **can** be considered equal to a `float` (_or an [`octal`][octal] equal to a [`hexadecimal`][hex]_), as long as the types can be implicitly converted for comparison.
+
+For the other numeric types ([complex][complex numbers], [decimal][decimal numbers], [fractions][rational numbers]), comparison operators are defined where they "make sense" (_where implicit conversion does not change the outcome_), but throw a `TypeError` if the underlying objects cannot be accurately converted for comparison.
+For more information on the rules that python uses for numeric conversion, see [arithmetic conversions][arithmetic conversions] in the Python documentation.
+
 
 ```python
->>> 7 > 5
-True
->>> 99 < 100
-True
->>> 4 < 4
+import fractions
+
+# A string cannot be converted to an int.
+>>> 17 == '17'
 False
->>> 4 <= 4
+
+# An int can be converted to float for comparison.
+>>> 17 == 17.0
 True
->>> 1 >= 1
+
+# The fraction 6/3 can be converted to the int 2
+# The int 2 can be converted to 0b10 in binary.
+>>> 6/3 == 0b10 
 True
->>> 5 == 5
+
+# An int can be converted to a complex number with a 0 imaginary part.
+>>> 17 == complex(17)
 True
->>> 6 != 6  # not equal to
+
+# The fraction 2/5 can be converted to the float 0.4
+>>> 0.4 == 2/5
+True
+
+>>> complex(2/5, 1/2) == complex(0.4, 0.5)
+True
+```
+
+Any ordered comparison of a number to a `NaN` (_not a number_) type is `False`.
+A confusing side-effect of Python's `NaN` definition is that `NaN` never compares equal to `NaN`.
+If you are curious as to why `Nan` was defined this way in Python, this [Stack Overflow Post on NaN][SO NaN Post] around the setting of the international standard is an interesting read.
+
+
+```python
+>>> x = float('NaN')
+
+>>> 3 < x
 False
->>> 5 != 6
-True
->>> not 3 == 3  # interpreted as not(3 == 3)
+
+>>> x < 3
+False
+
+# NaN never compares equal to NaN
+>>> x == x
 False
 ```
 
+
+## Comparing Strings
+
+Strings (`str`) are compared [_lexicographically_][lexographic order], using their individual Unicode code points (_the result of passing each code point in the `str` to the built-in function [`ord()`][ord], which returns an `int`_).
+If all code points in both strings match and are _**in the same order**_, the two strings are considered equal.
+This comparison is done in a 'pair-wise' fashion - first-to-first, second-to-second, etc.
+Unlike in Python 2.x, in Python 3.x, `str` and `bytes` cannot be directly coerced/compared.
+
+
+```python
+>>> 'Python' > 'Rust'
+False
+
+>>> 'Python' > 'JavaScript'
+True
+
+# Examples with Mandarin.
+# hello < goodbye
+>>> '你好' < '再见'  
+True
+
+# ord() of first characters
+>>> ord('你'), ord('再')
+(20320, 20877)
+
+# ord() of second characters
+>>> ord('好'), ord('见')
+(22909, 35265)
+
+# And with Korean words.
+# Pretty < beautiful.
+>>> '예쁜' < '아름다운'  
+False
+
+>>> ord('예'), ord('아')
+(50696, 50500)
+```
+
+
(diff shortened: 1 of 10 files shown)


concepts/comparisons/about.md
```diff
@@ -1,199 +1,291 @@
...
+| Operator | Operation                | Description                                                    |
+| -------- | ------------------------ | -------------------------------------------------------------- |
+| `>`      | "greater than"             | `a > b` is `True` if `a` is **strictly** greater in value than `b` |
+| `<`      | "less than"                | `a < b` is `True` if `a` is **strictly** less in value than `b`    |
+| `==`     | "equal to"                 | `a == b` is `True` if `a` is **strictly** equal to `b` in value    |
+| `>=`     | "greater than or equal to" | `a >= b` is `True` if `a > b` OR `a == b` in value             |
+| `<=`     | "less than or equal to"    | `a <= b` is `True` if `a < b` or `a == b` in value             |
+| `!=`     | "not equal to"             | `a != b` is `True` if `a == b` is `False`                      |
+| `is`     | "identity"                 | `a is b` is `True` if **_and only if_** `a` and `b` are the same _object_  |
+| `is not` | "negated identity"                 | `a is not b` is `True` if `a` and `b` are **not** the same _object_ |
```
kotp, line 17: If we are alligning for the table we should do so.

concepts/comparisons/about.md
```diff
@@ -1,199 +1,291 @@
...
+
+
 ## Identity comparisons
 
-The operators `is` and `is not` test for an object's _identity_.
-An object's identity is determined using the `id()` function.
+The operators `is` and `is not` test for object [_identity_][object identity], as opposed to object _value_.
+An object's identity never changes after creation and can be found by using the [`id()`][id function] function.
+
+\<apple\> `is` \<orange\> evaluates to `True` if _**and only if**_ `id(<apple>)` == `id(<orange>)`.
```
kotp, line 188: If the `<>` for this is meant to be code, we should do something similar as done in the "and only if" statement where it is wrapped in code delimiters inline.

concepts/comparisons/about.md
```diff
@@ -1,199 +1,291 @@
...
+apple `is not` orange yields the inverse.
```
kotp, line 189: It looks like the code delimiter changed here as well, perhaps unintended.

concepts/comparisons/introduction.md
```diff
@@ -1,23 +1,65 @@
...
+| Operator | Operation                | Description                                                    |
+| -------- | ------------------------ | -------------------------------------------------------------- |
+| `>`      | "greater than"             | `a > b` is `True` if `a` is **strictly** greater in value than `b` |
+| `<`      | "less than"                | `a < b` is `True` if `a` is **strictly** less in value than `b`    |
+| `==`     | "equal to"                 | `a == b` is `True` if `a` is **strictly** equal to `b` in value    |
+| `>=`     | "greater than or equal to" | `a >= b` is `True` if `a > b` OR `a == b` in value             |
+| `<=`     | "less than or equal to"    | `a <= b` is `True` if `a < b` or `a == b` in value             |
+| `!=`     | "not equal to"             | `a != b` is `True` if `a == b` is `False`                      |
+| `is`     | "identity"                 | `a is b` is `True` if **_and only if_** `a` and `b` are the same _object_  |
+| `is not` | "negated identity"                 | `a is not b` is `True` if `a` and `b` are **not** the same _object_ |
```
kotp, line 17: Same here for alignment of column delimiter.

concepts/comparisons/introduction.md
```diff
@@ -1,23 +1,65 @@
...
-For example, the `str` value of a number is considered completely different from the `integer` or `floating-point` value.
-However, an `integer` **can** be considered equal to a `float`, as they are both numeric types that Python can implicitly convert to compare.
-For other numeric types, comparison operators are defined where they "make sense", but throw a `TypeError` if the underlying objects cannot be converted for comparison.
-For more information on the rules that python uses for numeric conversion, see [arithmetic conversions][arithmetic conversions] in the Python documentation.
 
-[comparisons]: https://docs.python.org/3/library/stdtypes.html?
-[arithmetic conversions]: https://docs.python.org/3/reference/expressions.html?highlight=number%20conversion#arithmetic-conversions
+They all have the same priority (_which is higher than that of [Boolean operations][boolean operations], but lower than that of arithmetic or bitwise operations_).
+
+
```
kotp, line 24: Two blank spaces which is usually ignored in HTML rendering, and not really helpful in the plain text viewing.  This happens in other places as well.  Single blank line between parenthesis is likely intended.

exercises/concept/black-jack/.docs/introduction.md
```diff
@@ -1,185 +1,249 @@
...
+| Operator | Operation                | Description                                                    |
+| -------- | ------------------------ | -------------------------------------------------------------- |
+| `>`      | "greater than"             | `a > b` is `True` if `a` is **strictly** greater in value than `b` |
+| `<`      | "less than"                | `a < b` is `True` if `a` is **strictly** less in value than `b`    |
+| `==`     | "equal to"                 | `a == b` is `True` if `a` is **strictly** equal to `b` in value    |
+| `>=`     | "greater than or equal to" | `a >= b` is `True` if `a > b` OR `a == b` in value             |
+| `<=`     | "less than or equal to"    | `a <= b` is `True` if `a < b` or `a == b` in value             |
+| `!=`     | "not equal to"             | `a != b` is `True` if `a == b` is `False`                      |
+| `is`     | "identity"                 | `a is b` is `True` if **_and only if_** `a` and `b` are the same _object_  |
+| `is not` | "negated identity"                 | `a is not b` is `True` if `a` and `b` are **not** the same _object_ |
```
kotp, line 16: Another column delimiter off kilter.

exercises/concept/black-jack/.docs/introduction.md
```diff
@@ -1,185 +1,249 @@
...
 ```
 
-Operator `is not` tests if the first and second operand is not the same object.
+
+## Comparing object identity
+
+The operators `is` and `is not` test for object [_identity_][object identity], as opposed to object _value_.
+An object's identity never changes after creation and can be found by using the [`id()`][id function] function.
+
+\<apple\> `is` \<orange\> evaluates to `True` if _**and only if**_ `id(<apple>)` == `id(<orange>)`.
```
kotp, line 157: Starting to feel like this is coming from some place higher up the chain maybe, as this seems like it should be

    apple is orange

We can compare apples to oranges, and those would be valid names, but not sure what the `<>` does to them, though I think the communication that I get from reading about apples and oranges are a generation, kind of like `foo` or `bar`, and maybe the `<foo>` and `<bar>` does something different in Python specifically.

So maybe take the feedback like you might with apple slices, you know, with a little salt.

exercises/concept/black-jack/.docs/introduction.md
```diff
@@ -1,185 +1,249 @@
...
+apple `is not` orange yields the inverse.
```
kotp, line 158: I think this should be 

```markdown
`apple is not orange` yields the inverse.
```

instead.

exercises/concept/black-jack/.docs/introduction.md
```diff
@@ -1,185 +1,249 @@
...
+
+Due to their singleton status, `None` and `NotImplemented` should always be compared to items using `is` and `is not`.
```
kotp, line 160: Maybe expand on why, other than only "because of their singleton status".  I am sure there is something more implied here.  And I can not think of why I would ever do other comparisons, anyway.  So the fact that it is mentioned and specifically mentioned that they are singletons, raises an eyebrow, wanting more information.

exercises/concept/black-jack/.meta/exemplar.py
```diff
@@ -1,68 +1,93 @@
+"""Functions to help play and score a game of blackjack.
+
+How to play blackjack:    https://bicyclecards.com/how-to-play/blackjack/
+"Standard" playing cards: https://en.wikipedia.org/wiki/Standard_52-card_deck
+
+"""
```
kotp, line 6: Why a blank line around this string, and not the following?

exercises/concept/black-jack/black_jack.py
```diff
@@ -1,50 +1,64 @@
+"""Functions to help play and score a game of blackjack.
+
+How to play blackjack:    https://bicyclecards.com/how-to-play/blackjack/
+"Standard" playing cards: https://en.wikipedia.org/wiki/Standard_52-card_deck
+
+"""
```
kotp, line 6: Also blank line around the string content unlike some other `"""` strings.

kotp (changes requested):
Looking pretty good, but I think there may have been some unintended changes, plus what appears to be style deviance instances in the documents themselves.

And a question about `<apples>` and `<oranges>`.

concepts/comparisons/about.md
BethanyG: There's a big difference in the way this appears in code vs render.  When rendered, this is aligned: 

| Operator | Operation                | Description                                                    |
| -------- | ------------------------ | -------------------------------------------------------------- |
| `>`      | "greater than"             | `a > b` is `True` if `a` is **strictly** greater in value than `b` |
| `<`      | "less than"                | `a < b` is `True` if `a` is **strictly** less in value than `b`    |
| `==`     | "equal to"                 | `a == b` is `True` if `a` is **strictly** equal to `b` in value    |
| `>=`     | "greater than or equal to" | `a >= b` is `True` if `a > b` OR `a == b` in value             |
| `<=`     | "less than or equal to"    | `a <= b` is `True` if `a < b` or `a == b` in value             |
| `!=`     | "not equal to"             | `a != b` is `True` if `a == b` is `False`                      |
| `is`     | "identity"                 | `a is b` is `True` if **_and only if_** `a` and `b` are the same _object_  |
| `is not` | "negated identity"                 | `a is not b` is `True` if `a` and `b` are **not** the same _object_ |
| `in`     | "containment test"         | `a in b` is `True` if `a` is member, subset, or element of `b`               |
| `not in` | "negated containment test" | `a not in b` is `True` if `a` is not a member, subset, or element of `b`     |

concepts/comparisons/about.md
BethanyG: Yeah. I debated that.  This way, it emphasizes the `is` operator, but requires I escape the `<`.  Not sure I like how `<apple> is <orange>` reads as one block.  We _could_ do `<apple>` `is` `<orange>` ...dunno if that is better.  Thoughts?

concepts/comparisons/about.md
BethanyG: Same as above -- not sure what reads better, but a whole block doesn't read well to me.

concepts/comparisons/introduction.md
BethanyG: Same comment as above.  The render is more important than the code -- and the render is in alignment.

concepts/comparisons/introduction.md
BethanyG: I am not sure I follow here.  Which two spaces are we talking about?

exercises/concept/black-jack/.docs/introduction.md
BethanyG: 😆   So..same comment here as I had earlier.  We just need to decide which reads better.  The _intent_ of the `<` is to denote something similar to `<foo>` and `<bar>` - placeholders -- just not `foo` and `bar` placeholders.

exercises/concept/black-jack/.docs/introduction.md
BethanyG: yeah..or whatever we end up deciding.  Although it should be `<apple> is not <orange>` -- if we want to keep to the `<>` signifying a variable or object.

exercises/concept/black-jack/.docs/introduction.md
BethanyG: I can put a link into the python docs.  We're practically quoting them here.  The `singleton status` -- is exactly that -- there is only ever ONE instance of either `None` or `NotImplemented` in any running Python program.  So equivalence is never the right metaphor.....it is either **THE** `None` or **THE** `NotImplemented` - nothing is ever equal to them, because nothing can ever BE them .... except themselves.

BethanyG: Here is the relevant snip:

BethanyG: ..I guess we could also link to PEP8 - but the length of that document might confuse someone who is trying to just learn about comparisons.

exercises/concept/black-jack/.meta/exemplar.py
BethanyG: [PEP257](https://www.python.org/dev/peps/pep-0257/) puts the closing quotes on their own lines.  PyCharm defaults to inserting and extra line above - and since I think it also reads better that way, I left it.

concepts/comparisons/about.md
BethanyG: Neverimind.  Prettier to the rescue!

kotp: Yep, I knew that it would look good rendered (markdown is awesome, tables are cool that way), but having it look good in source is nice too.

The other choice is "less white space" and not align any of it, but have it naturally end, which I do not _not_ like choice as well.

But having it inconsistent, well, that is distracting. ;)

exercises/concept/black-jack/.meta/exemplar.py
BethanyG: I'll normalize it.

BethanyG:
Alrighty.  Ran everything through priettier, and cleaned up the `<apples>` and `<oranges>` 😉   I also removed the excess lines in the module docstrings.  Let me know if there's anything I've missed.  😄   Thanks for reviewing this!

exercises/concept/black-jack/.docs/introduction.md
kotp: I like that, I think the link to that section is good.  I like the addition of "never the equality operators".  A little more prodding to the right idea here.  That we check for identity not equality when asking if `ourThing is None`, or if `ourBehavior is not NotI'mplemented` if that is correct.

exercises/concept/black-jack/.docs/introduction.md
kotp: If that `<>` indicates variable or object in Python speak then that is what should be used.  I think in the context, perhaps it is clear to focus on the `is not` rather than the variable name used (and is why we use foo and bar as well known "can be whatever, as allowed by the syntax of the language in that case sensitive interpretation".

concepts/comparisons/introduction.md
kotp: Two blank vertical spaces (new lines) is what I really mean here.

concepts/comparisons/about.md
kotp: I think that I notice the delimiters  and generally ignore the name of the variables being used anyway.  So to me the `<>` is distracting from the communication, while if I saw `apple is orange` where the reading is talking about `is` and `is not` I would see the example as intended, that the names on either side do not really matter, the "how to use the identity operator `is` or `is not`.  It is why we see `foo` and `bar` as nonsense names in code documentation, we learn that those are not the important part, only that something is there in each place.

exercises/concept/black-jack/.docs/introduction.md
kotp: Right, but in this case, I see the `<>` and that yells pay attention to this, because that is what delimiters do for me in code.

BethanyG:
@kotp - are there additional changes needed here for approval?  Let me know - thanks!

kotp (approved):

kotp:
I approve, as long as `<fish>` is valid Python.  I kind of think it may be confusing if the student plus that in and it does not work.  And then has to interpret it more as Unix documentation style for the meanings of those brackets.

BethanyG:
I have no bead on if `<fish>` is going to be confusing.  The convention is used in many Python concept exercises to denote a placeholder variable or parameter, to signal to the student that the _name_ can really be anything they want.

My thought is to have the exercise go live in `beta` mode, and if it is confusing to students they will (loudly) let us know - and we can discuss re-working it at that point.

kotp:
On 22/01/08 10:27AM, BethanyG wrote:
> I have no bead on if `<fish>` is going to be confusing.  The
> convention is used in many Python concept exercises to denote a
> placeholder variable or parameter, to signal to the student that the
> _name_ can really be anything they want.

Also, if it has been being used, and understood well,
already, then I do not have a problem with it.

> My thought is to have the exercise go live in `beta` mode, and if it
> is confusing to students they will (loudly) let us know - and we can
> discuss re-working it at that point.

Definitely.  I do not know if this is common outside of the Python
track, and used widely in the Python greater community.  As an
"outsider" I definitely that as a perspective.  But if it is convention
in the Python community, then it should be used here as well, as the
student will learn that this is not something unique in how this is
communicated on Exercism.

PR #2847, opened 2022-01-09
Update to the test_is_blackjack function

PaulT89:
Solution/Modification: Added two extra test cases to the data array in the test_is_blackjack function.

Problem: One can conceivably pass the test_is_blackjack test by calling the value_of_card function for each card, add the results together, and then test if the value was 11 (as an Ace in the value_of_card function is fixed as 1).  However, this alone does not necessarily indicate that blackjack has been achieved, as any two cards that sum to 11 (such as 4 and 7, or 9 and 2) would also pass the test, even though they shouldn't.  For blackjack to occur, one must also check that one of the cards is an Ace (or, alternatively, that one of the cards has a value of 10)

### DIFF
--- a/exercises/concept/black-jack/black_jack_test.py
+++ b/exercises/concept/black-jack/black_jack_test.py
@@ -64,7 +64,8 @@ class BlackJackTest(unittest.TestCase):
     def test_is_blackjack(self):
         data = [
                 (('A', 'K'), True), (('10', 'A'), True),
-                (('10', '9'), False), (('A', 'A'), False)]
+                (('10', '9'), False), (('A', 'A'), False),
+                (('4', '7'), False), (('9', '2'), False)]
 
         for variant, (hand, blackjack) in enumerate(data, 1):
             with self.subTest(f'variation #{variant}', input=hand, output=blackjack):


BethanyG (approved):
Hi @PaulT89  👋🏽  

This is an excellent catch!  Thank you so much for submitting this PR.  You've made this exercise better -- and its only been "out in the world" for 24 hours!  😄   🎉

PaulT89:
You're welcome.  Literally created a GitHub account just for this.  I've also noticed a bunch of spelling errors in the instructions of a few of the Common Lisp exercises.  Do I just do the same thing over there?

BethanyG:
@PaulT89  -- yes!  😄   I would expect that the maintainer of the `lisp` track would be delighted to have some PRs that cleaned up spelling and other errors.

BethanyG:
@PaulT89 😱   _**I am so so sorry!**_  I forgot that you need to add your GitHub user name to the contributors list for this exercise! [Here](https://github.com/exercism/python/blob/main/exercises/concept/making-the-grade/.meta/config.json) is an example of what that looks like.  You need to add the key, list, and your Git Hub user name to [this](https://github.com/exercism/python/blob/main/exercises/concept/black-jack/.meta/config.json) file. 

I would do it for you -- but I want you to also get credit for the PR.  😄 

 Let me know if you have any questions or issues.  Again, apologies for not remembering to do that earlier.

PaulT89:
Actually, it looks like I just needed to add my name - the key and list were already there.  Thanks for telling me - I wouldn't have known to do so otherwise. 😊

PR #2851, opened 2022-01-12
Update sets.py

zepam:
Docstring in categorize_dish() and tag_special_ingredients() references categories.py, which doesn't exist.
The correct reference is sets_categories_data.py.
This is an additional update from the 1/9 commit making similar changes in the code.

### DIFF
--- a/exercises/concept/cater-waiter/sets.py
+++ b/exercises/concept/cater-waiter/sets.py
@@ -43,7 +43,7 @@ def categorize_dish(dish_name, dish_ingredients):
     :return: str "dish name: CATEGORY"
 
     This function should return a string with the `dish name: <CATEGORY>` (which meal category the dish belongs to).
-    All dishes will "fit" into one of the categories imported from `categories.py`
+    All dishes will "fit" into one of the categories imported from `sets_categories_data.py`
     (VEGAN, VEGETARIAN, PALEO, KETO, or OMNIVORE).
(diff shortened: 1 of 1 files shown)


BethanyG (changes requested):
Thank you for submitting this!  You know, I did look over the stub and `.meta/exemplar.py` files when you submitted your first PR -- but my brain skipped right over those docstrings!  😱  .  Glad you caught the references.  Could you also change the docstrings on the `.meta/exemplar.py` file as well?  That would be much appreciated.  😄

BethanyG (approved):
Excellent!  Thank you for your work.  😄

BethanyG:
@zepam -  😱   _I am so so sorry!_  I forgot that you need to add your GitHub user name to the contributors list for this exercise! [Here](https://github.com/exercism/python/blob/main/exercises/concept/making-the-grade/.meta/config.json) is an example of what that looks like.  You need to add the key, list, and your name to [this](https://github.com/exercism/python/blob/main/exercises/concept/cater-waiter/.meta/config.json) file. 

I would do it for you -- but I want you to also get credit for the PR.  😄 

 Let me know if you have any questions or issues.  Again, apologies for not remembering to do that earlier.  😞

zepam:
Thanks for thinking of it! I didn't even know I could do that... and now I get another PR. 😄

PR #2852, opened 2022-01-12
Update exemplar.py

zepam:
Docstring in categorize_dish() and tag_special_ingredients() references categories.py, which doesn't exist.
The correct reference is sets_categories_data.py.
This is an additional update from the 1/9 commit making similar changes in the code.


BethanyG (approved):
@zepam  -- in future, if files are related to a single exercise, they can go in the same PR.  😄   Rather than have you re-do, I'll just go ahead and merge both.  Thanks again for your work!  🌟

PR #2865, opened 2022-01-21
Update instructions.md of the python Black-jack exercice

achrafattary:
I was working on this exercise, and I found it confusing that it is not mentioned whether it is possible to already have an ace in hand, and if that was the case, what should be it's value. So I checked the test file and i saw that one example has an Ace in it and that it behaves as an 11,so,for clarification, I added a hint that initial Aces in hand should be considered as 11.


BethanyG (approved):
@achrafattary Thanks for submitting this!  Nice clarification.  😄   🌟

PR #2866, opened 2022-01-21
Update black_jack.py

achrafattary:
Added that Aces takes 11 if already in hand in part 3 for coherence with changes made in instruction.md


achrafattary:
Added in the docstring of the method*

BethanyG (approved):
Awesome.  Thanks!  😄   In future, its fine to combine things like this into one PR if it relates to one exercise or concept.

PR #2867, opened 2022-01-21
Update instructions.md

achrafattary:
the paragraph in the third part was  too dense, made the hint in a new line and made the "hint" bold for coherence with other exercises


BethanyG (approved):
I like this.  I agree:  it was too dense the way it was.  **Thanks!**  😄

PR #2870, opened 2022-01-23
ghost-gobble-arcade_game: Fixed typos in hints.md

PaulT89:
I was just looking over this and found that only the first sentence had 'operators' (plural) while the other 3 had 'operator' (singular).  Judging by the text in the indirect link, I figured that the latter 3 were incorrect.


BethanyG (approved):
Nice catch.  Thanks for submitting this!  🌟

PR #2880, opened 2022-01-27
I added one more column ('2', 'A', 1) for the test of the "value_of_ace"

EZFRICA:
I added one more column ('2', 'A', 1) for the test of the "value_of_ace" function to check if the value of the ace we already have in hand is 11.
"Hint: if we already have an ace in hand then it's value would be 11."

### DIFF
--- a/exercises/concept/black-jack/.meta/exemplar.py
+++ b/exercises/concept/black-jack/.meta/exemplar.py
@@ -53,11 +53,10 @@ def value_of_ace(card_one, card_two):
     :return: int - value of the upcoming ace card (either 1 or 11).
     """
 
-    if 11 + (value_of_card(card_one) + value_of_card(card_two)) > 21:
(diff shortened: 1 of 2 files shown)


BethanyG:
Hi @EZFRICA - Thank you so much for submitting this PR!  🌟  

The new test case caused the existing [exemplar.py](https://github.com/exercism/python/blob/main/exercises/concept/black-jack/.meta/exemplar.py) code for this exercise to fail.  I've updated the exemplar code so that the CI checks pass.  

In future, when proposing new test cases or test data, please double-check that the `exemplar.py` code for the exercise still passes.  If there is a test failure during that check, please either submit changes to the exemplar where they make sense -- or double-check that the test case proposed is a correct one.   _Thanks so much for your attention!_  😄

BethanyG (approved):

PR #2885, opened 2022-01-29
Correct typo in about.md

bhagerty:
Parenthetical phrase "(Its turtles all the way down)" should be "(it's turtles" etc. - the parenthetical phrase is part of the sentence, so "Its"should not be capitalized, and here, the word should be "it's" because the meaning is "it is," represented by the contraction "it's", not the possessive "Its."


BethanyG (approved):
Nice catch!  Thank you for the PR.  😄  🌟

Issue #2871, opened 2022-01-23
[Kindergarten Garden] — Test suite contradicts problem instructions


jaywritescode:
In the "Kindergarten Garden" problem, the instructions specifically say:

> Given a diagram, determine which plants each child in the kindergarten class is responsible for....

> There are 12 children in the class: Alice, Bob, Charlie, David, Eve, Fred, Ginny, Harriet, Ileana, Joseph, Kincaid, and Larry.

The first seventeen of nineteen tests in the problem's test suite conform to the given instructions; however, the final two unit tests introduce students Samantha, Patricia, Roger, and Xander. These students, according to a reasonable interpretation of the instructions, are _not_ in the class. We are told that the class has twelve children and then given the names of those twelve children. 

Related to #2856, this contradiction could be resolved by removing the childrens' names from the instructions and explicitly passing them into the driver function, or by removing the last two tests in the test suite.

BethanyG:
Hi @jaywritescode 👋🏽 

Thanks for filing this issue!  I agree that the instructions for this exercise might be easily mis-interpreted or mis-read without careful reading of the test file and a bit of trial-and-error.

As outlined in my reply to issue #2856  - the instructions for this `practice` exercise (_actually all current practice exercises_) are cross-track and are maintained in a repo called [problem-specifications](https://github.com/exercism/problem-specifications).  General changes to them are proposed and discussed there between different track maintainers. Only track-specific/language specific appends are PR'd to this repo.

The reason I proposed an `instructions.append` for #2856 was that the poster was talking about `class` interfaces and other OOP details that are (somewhat) specific to Python.  For example, [this same problem in Haskell](https://github.com/exercism/haskell/tree/main/exercises/practice/kindergarten-garden) doesn't have `classes`, because the Haskell programming language doesn't implement them.

However, the issue that **you** are describing gets at the way the core description and tests for this exercise are created.  So your proposed changes should be brought up in [problem-specifications](https://github.com/exercism/problem-specifications), so everyone can talk them through and see if it does indeed make sense to change the wording of the problem, or eliminate or change some of the tests described.

Given that, I am going to close this issue -- but encourage you to take the text of what you've posted here and open an issue for discussion in the `problem-specifications` repo.

jaywritescode:
Thanks for redirecting me to the right place. Much appreciated!

PR #2887, opened 2022-01-31
Revise problem statement to reflect tests

bhagerty:
I propose revising the problem statement to reflect the tests. 

- First, I think the statement should make it clearer what kind of result to return (a list, with each plant's name capitalized). As written, it seems like you should return a string with the first letter capitalized. That's what I tried to do, and only when it failed the test did I realize what I should be returning. 
- Second, I propose clarifying what the row of plants is (a string, with a newline character separating the rows). 
- Third, I propose clarifying that a list of students will only optionally be supplied, and what to do if it is supplied. As written, the problem statement just says there are 12 students, giving no indication that our solution should be able to handle a different list of students. I think we should flag this for users, rather than having them discover it for the first time when they get failing tests.

### DIFF
--- a/exercises/practice/kindergarten-garden/.docs/instructions.md
+++ b/exercises/practice/kindergarten-garden/.docs/instructions.md
@@ -39,11 +39,13 @@ RG......................
 In the first row, nearest the windows, she has a violet and a radish.  In the
 second row she has a radish and some grass.
 
-Your program will be given the plants from left-to-right starting with
-the row nearest the windows. From this, it should be able to determine
-which plants belong to each student.
+Your program will be given a diagram of plants from left-to-right starting with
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @bhagerty 👋🏽 

Thank you for taking the time to submit this PR!  

Unfortunately, as is described in the now-closed issue #2871 -- since `Kintergarten Garden` is a practice exercise, all proposed changes to the instructions proper need to be made in the [problem specifications](https://github.com/exercism/problem-specifications) repo.  Only Python Track-specific changes would be made here.

Suggest you open this proposed change in [problem-specifications](https://github.com/exercism/problem-specifications), so it can be discussed cross-track.  Thanks!

PR #2889, opened 2022-01-31
[Grade School]: Stub out `added` and `add_student` methods that are called in the tests

bhagerty:
A bunch of tests call an "added()" method, but the method is not stubbed out. At a minimum, it should be stubbed out.

### DIFF
--- a/exercises/practice/grade-school/grade_school.py
+++ b/exercises/practice/grade-school/grade_school.py
@@ -10,3 +10,6 @@ class School:
 
     def grade(self, grade_number):
         pass
+
+    def added(self):
+        pass


BethanyG:
Hi @bhagerty 👋🏽 

Nice catch on this!  And in that same spirit, I think we should also add `add_student()` and `class School` to the stub file.  I am also going to PR an `instructions.append` to push students toward making a class, since this exercise requires one.

exercises/practice/grade-school/grade_school.py
```diff
@@ -10,3 +10,6 @@ def roster(self):
 
     def grade(self, grade_number):
         pass
+
+    def added(self):
+        pass
```
BethanyG, line 15:
```suggestion
        pass
        
        
    def add_student(self, name, grade):
        pass
```

BethanyG (changes requested):

exercises/practice/grade-school/grade_school.py
bhagerty: Thanks for responding @BethanyG ! I'm not sure I understand your suggestions. The existing stub file,  before my changes, is below, and it starts with `class School` and includes `def add_student(self, name, grade)`. So unless I'm missing something, there's no need for the changes you propose. Am I missing something?

```
class School:
    def __init__(self):
        pass

    def add_student(self, name, grade):
        pass

    def roster(self):
        pass

    def grade(self, grade_number):
        pass
```

BethanyG:
Nope.  Clearly, I was having a late-night brain fart.  😉

BethanyG (approved):

PR #2890, opened 2022-02-01
[Grade-School]: Added `instructions.append` to direct students to make a School `class`

BethanyG:


kotp (approved):
Slight nit: Add rather than added.  Commits are present tense, rather than past tense, as they effect the now, rather than the past, when applied.

PR #2892, opened 2022-02-04
refactor: reformat lists_test.py by pylint

noritakaIzumi:
## fixed:

- trim whitespaces in the end of line
- trim blanklines between pytest tasks (PEP 8)

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -66,7 +66,7 @@ jobs:
       run: pip install dataclasses
 
     - name: Install pytest
-      run: pip install pytest
+      run: pip install pytest~=6.2.5
 
     - name: Check exercises
       run: |
--- a/exercises/concept/card-games/lists_test.py
+++ b/exercises/concept/card-games/lists_test.py
@@ -18,8 +18,8 @@ class CardGamesTest(unittest.TestCase):
 
         input_vars = [0, 1, 10, 27, 99, 666]
 
-        results = [[0, 1, 2], [1, 2, 3], 
-                   [10, 11, 12], [27, 28, 29], 
+        results = [[0, 1, 2], [1, 2, 3],
+                   [10, 11, 12], [27, 28, 29],
                    [99, 100, 101], [666, 667, 668]]
 
         for variant, (number, rounds) in enumerate(zip(input_vars, results), start=1):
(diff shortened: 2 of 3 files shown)


BethanyG (approved):
@noritakaIzumi -- Thank you for this!  🌟

BethanyG:
Humm, the CI is failing for this.  Investigating.

noritakaIzumi:
Is it ok if I tried to add a row to `requirements.txt` ?

The runner fails and prints the following in "canonical_sync":
```text
ModuleNotFoundError: No module named 'toml'
```

I saw the module `toml` not in `requirements.txt`🤔

noritakaIzumi:
Pytest have released a new version yesterday.
And the module `toml` (missing module) is replaced with `tomli`

https://github.com/pytest-dev/pytest/releases/tag/7.0.0#:~:text=Switch%20TOML%20parser%20from%20toml%20to%20tomli

I think we have to specific version of pytest in `canonical_sync`.

In other words

`.github/workflows/ci-workflow.yml` line 69:

```yaml
      run: pip install pytest
```

needs to be this

```yaml
      run: pip install pytest~=6.2.2
```

BethanyG:
@noritakaIzumi   Thank you for the offer of help.  😄    

This is not _exactly_  about a missing requirement -- at least not one in `requirements.txt`.  When the CI runs, it builds our [test runner](https://github.com/exercism/python-test-runner) by pulling _it's_ `requirements.txt` and making a docker container.  Then, from inside the container, it runs a script located in `python/bin` named `test_exercises.py`.  

`test_exercises.py` imports `data.py` from the same directory.  It's the `data.py` file that is failing on the `toml` import error.  But it **_shouldn't_** be having that issue - so I think the error may be tagging off of something incorrect.  That import has been there a looong time.  It only started failing today, PST.  Additionally, the same script running on my desktop is succeeding.

What **did** change today was the release of PyTest 7.0 (_we were using 6.2.5_).  And since our `requirements.txt` only specifies a minimum version and not a max, the CI container got built with PyTest 7.0.  Since that is what has most recently changed,  I am attempting to confirm by testing on my machine.  And I would love help -- if you are familiar/comfortable with docker and running pytest, et. al. 🌟 

Basically, I am trying to figure out what changed to make the CI throw a fit over a toml import that has not changed.  My working theory is that PyTest 7.0 has caused some sort of issue, that is manifesting with that error.  But right now, I can't figure it out.  😢   I'm having issues getting the script to run manually from docker.  

If it **is** PyTest 7.0, we can fix it by changing the `requirements.txt` for the _runner_, so that it forces PyTest 6.2.5....if that makes sense?

BethanyG:
@noritakaIzumi  -- didn't see your second comment -- YES!  I think you are correct there -- we need to pin the version until we can figure out why PyTest 7.0 is making a mess of things.

noritakaIzumi:
@BethanyG I think so too.

BethanyG:
_**and just like that!**_  😄   🎉 

@noritakaIzumi  -- care to PR a change to the runner `requiremnts.txt`, so that we make sure it doesn't pull PyTest 7.0 by accident?  😉    The repo is here:  https://github.com/exercism/python-test-runner.

noritakaIzumi:
@BethanyG I got it 🥰

PR #2899, opened 2022-02-06
[Collections] Corrected typo

Timo4B:
deleted ","   #2901


BethanyG:
I disagree.  I think the comma was in the correct place, but the space in front of it needed to go.

But this blurb also needs to be re-written slightly in light of the introduction of structural pattern matching in Python 3.10, so I think we'll leave it as-is until that re-write can be logged and done.

PR #2908, opened 2022-02-09
Add additional test case to black-jack

DanielNoord:
I saw this in a submission that passed the test. The incorrect implementation was:

```python
def is_blackjack(card_one, card_two):
    if card_one != card_two:
        if card_one and card_two in ['J','Q',"K",'A']:
            return True 
    return False
```

Not sure if I should have opened an issue for this first and if patches like these are accepted but just thought I would open a PR with a possible fix. Let me know!


BethanyG:
Since the CI will not pass without updating exercises from problem specifications, I am putting this on hold until that can happen.

BethanyG (approved):
@DanielNoord  - Good catch!  Thank you for submitting this additional test case.  It really helps.

Issue #2910, opened 2022-02-09
Incorrect tests in book_store.py


tamireinhorn:
Despite the README from the exercise being the same as the C# track, for example, the tests all expect the return to be an int, and not a float. Therefore, even for the simplest test 'test_only_a_single_book', the test expectes 800, not 8! Either the tests should go have their results changed or the readme should be different on this track.

kotp:
The base unit is 1 cent, and so 800 is an appropriate price for a book.

PR #2911, opened 2022-02-09
Edited tests so they use the proper result

tamireinhorn:
As said in the issue, all the tests on this exercise for Python for some reason expected the result to be multiplied by 100, instead of the actual result from the calculation. This PR simply fixes them :)

This is related to [Issue #2910 ](https://github.com/exercism/python/issues/2910) which I just opened.

### DIFF
--- a/exercises/practice/book-store/book_store_test.py
+++ b/exercises/practice/book-store/book_store_test.py
@@ -10,11 +10,11 @@ from book_store import (
 class BookStoreTest(unittest.TestCase):
     def test_only_a_single_book(self):
         basket = [1]
-        self.assertEqual(total(basket), 800)
+        self.assertEqual(total(basket), 8.00)
 
     def test_two_of_the_same_book(self):
         basket = [2, 2]
-        self.assertEqual(total(basket), 1600)
+        self.assertEqual(total(basket), 16.00)
 
     def test_empty_basket(self):
         basket = []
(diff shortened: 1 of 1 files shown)


exercises/practice/book-store/book_store_test.py
```diff
@@ -10,91 +10,91 @@
 class BookStoreTest(unittest.TestCase):
     def test_only_a_single_book(self):
         basket = [1]
-        self.assertEqual(total(basket), 800)
+        self.assertEqual(total(basket), 8.00)
```
kotp, line 13: The price of a book is 800 cents, or 8 dollars.  By reducing the you are reducing the price of the books to 8 cents.

kotp (commented):
I believe the change is not appropriate.

exercises/practice/book-store/book_store_test.py
tamireinhorn: The base unit being one cent is never specified and is not present in the C# track, for instance. The price there is treated as 8.00, and, likewise, so it is in the README. From the example:
`5 * (8 - 2.00) == 5 * 6.00 == $30.00`
Were it as you said, it would have to all be written with 800 (cents) in mind, which it isn't.

kotp: Fair enough.  It is likely a choice in implementation as to which base unit to use.  Financials are often done with either cents or a tenth of a cent as a base unit.

Do you have an idea for improving the documentation rather than changing the tests?

kotp: We might look at the problem-specifications repository as well to determine if the documentation should be changed there as well, if it is not specified and not present in the other tracks.

This exercise has existed since 2017, so there is a bit of history with it.

BethanyG:
Hi @tamireinhorn 👋🏽 

The test results for this exercise were pulled directly from [problem-specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/book-store/canonical-data.json), and are written in cents.  The C# track decided to implement them as floats, but Python decided to leave them as written in the canonical data.  Neither interpretation is "right" or "wrong" -- just two different choices for two different tracks.

Since almost 500 students have published solutions to this exercise, and many more have completed it, I am loath to change the expected results and invalidate all of those solutions without a really good reason.

As @kotp mentions, we may want to do an instruction append or otherwise provide hints for those students who don't look at the test cases -- although I would argue that since this track attempts TDD, that the students _should_ be looking at the tests to see if an `int` or a `float` is expected. 

And all that being said, the tests for this exercise are also auto-generated from problem-specifications, based on a [JinJa template](https://github.com/exercism/python/blob/main/exercises/practice/book-store/.meta/template.j2).  In order to change the tests, you must edit the JinJa template and re-generate the test file.

tamireinhorn:
Hello! 
Fair enough, I just mentioned a problem that I myself had with the exercise, where it wasn't quite obvious or intuitive that cents was the base unit, given the README, the specifications and the idea being 'proper' pricing, which in reality is normally given in currency and not in cents. 
I think it would be reasonable to append to the instructions, if nothing else.

PR #2912, opened 2022-02-09
Bring hello-world in line with problem-specs

iHiD:
I get lots of people confused specifically on Python's hello world because it asks people to write a function, whereas in reality they just need to change the existing code. This brings Python in line with other tracks and will hopefully reduce the volume of emails I get!


kotp (approved):

kotp:
I hope that helps to reduce your incoming e-mail load.

BethanyG:
@kotp -- In the future, please ping me before merging an update like this.  Thanks!

PR #2917, opened 2022-02-10
[Phone Number] Only one problem per test input

astrieanna:
Because the area code is not allowed to start with 0 or 1, inputs designed to elicit other errors should not use area codes that start with either of those digits.

### DIFF
--- a/exercises/practice/phone-number/phone_number_test.py
+++ b/exercises/practice/phone-number/phone_number_test.py
@@ -22,7 +22,7 @@ class PhoneNumberTest(unittest.TestCase):
 
     def test_invalid_when_9_digits(self):
         with self.assertRaises(ValueError) as err:
-            PhoneNumber("123456789")
+            PhoneNumber("213456789")
         self.assertEqual(type(err.exception), ValueError)
         self.assertEqual(err.exception.args[0], "incorrect number of digits")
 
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @astrieanna 👋🏽 

This exercise is one that is auto-generated from [problem-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises/phone-number), so proposed changes/issues with test cases need to be discussed there.  We pass the test data from [cannonical data](https://github.com/exercism/problem-specifications/blob/main/exercises/phone-number/canonical-data.json) through via a [JinJa template](https://github.com/exercism/python/blob/main/exercises/practice/phone-number/.meta/template.j2) to generate the Python test code.

astrieanna:
Oh, sorry! 🤦🏻‍♀️ It's tricky to figure out what it Python-specific and what is from up-stream.

BethanyG:
I know!  We've over-complicated things a bit.  **_Apologies for that!_**  For future reference:

1.  [**Practice exercises**](https://exercism.org/docs/building/product/practice-exercises) under `exercises/practice` are largely pulled from `problem-specifications` so have all that _complication_, that needs `instruction.append` or `test.append` or PR-ing to the problem specs repo, or JinJa2 templating.

2. [**Concept exercises**](https://exercism.org/docs/building/product/concept-exercises) under `exercises/concept` are the "learning" exercises, and are specific to the Python track.  They are open for edit/suggestion/change -- to all the documents in their folders.  Although for some changes, you'll have to convince the maintainers.  😉

Issue #2920, opened 2022-02-10
[Cater Waiter] Example uses Dict instead of Set for introducing Sets in the Introduction


Timo4B:
In an example of the introduction to "Cater Waiter" a `dict ` is used to introduce "Fast Membership Testing" on sets. 

It is named `animals`:
```
>>> animals = {'chicken': 'white','sparrow': 'grey','eagle': 'brown and white',
                'albatross': 'grey and white','crow': 'black','elephant': 'grey',
                'dog': 'rust','cow': 'black and white','tiger': 'organge and black',
                'cat': 'grey','squirrel': 'black'}
```
Later this `dict `is used to show that a set is always a loose subset of itself:
```
# A set is always a loose subset of itself
>>> animals <= animals
True
```
There is also a `set` named `birds` in this example:
```
>>>  birds   = {'crow','sparrow','eagle','chicken', 'albatross'}
```
I would prefer using `birds` instead of `animals` to show that a **set** is always a loose subset of itself. The comment says...

> \# A set is always a loose subset of itself

So we should **use** a **set** as well. Otherwise, it will lead to confusion. Also, this error message comes up when you run the code from the example:

> TypeError: '<=' not supported between instances of 'dict' and 'dict'

`animals` is used later on in the exercise as an example that "Methods will take any iterable as an argument". It would therefore not be omitted.

exercises/concept/cater-waiter/.docs/introduction.md

BethanyG:
Closing as fixed.  Thanks!

PR #2921, opened 2022-02-10
[Cater Waiter] Change Dict to Set in Example of Introduction

Timo4B:
Replace **dictionary** `animals` by **set** `birds`. This should be done because the example is about sets. See #2920


BethanyG (approved):
@Timo4B   Thanks for this -- agree that the use of `animals` was either a cut-and-paste error or oversight on my part.  😄

Issue #2916, opened 2022-02-10
[Two Bucket]:  Update Test Generation Template To Accommodate New Tests


BethanyG:
Recently, we updated the `Two Bucket` exercise to incorporate new tests from the [problem-specifications](https://github.com/exercism/problem-specifications) repo.  Unfortunately, these new tests have broken test generation for the exercise, and had to be skipped in the [tests.toml](https://github.com/exercism/python/blob/main/exercises/practice/two-bucket/.meta/tests.toml) file.

We need to:
1.   Figure out why running `bin/generate_tests.py` is failing for `two-bucket` when the three tests are included.
      The three tests are:
      - [449be72d-b10a-4f4b-a959-ca741e333b72]
         description = "Not possible to reach the goal"
         include = false

      - [aac38b7a-77f4-4d62-9b91-8846d533b054]
         description = "With the same buckets but a different goal, then it is possible"
         include = false

      - [74633132-0ccf-49de-8450-af4ab2e3b299]
         description = "Goal larger than both buckets is impossible"
         include = false
2.  Make the appropriate edits to the `two-bucket` [JinJa2 template](https://github.com/exercism/python/blob/main/exercises/practice/two-bucket/.meta/template.j2)
3.  Regenerate the test file for `two-bucket` by running `bin/generate_tests.py two-bucket`
4.  Test that the code in the `two-bucket` [example.py](https://github.com/exercism/python/blob/main/exercises/practice/two-bucket/.meta/example.py) file passes all the re-generated tests.
5.   Make any necessary edits to `example.py` if there are any test failures.

This task requires knowledge of (_or interest in learning_):

[JinJa2 templating](https://realpython.com/primer-on-jinja-templating/)
[JinJa2 docs (_the version we use_)](https://jinja.palletsprojects.com/en/2.10.x/)
[The Python Test Generator](https://github.com/exercism/python/blob/main/docs/GENERATOR.md)

mhorod:
I think I have solved the issue. Is there any formatter for template files that I should use before committing my changes?

BethanyG:
Excellent, @mhorod.  Thank you so much for taking this on!  🌟 

>  Is there any formatter for template files that I should use before committing my changes?

No - we don't yet have a JinJa formatter in place.  😅    I'll have to eyeball it.  If you have experience with JinJa, and have recommendations on formatters, I'd love to hear them.  We have an issue open here on JinJa templating and linting: [issue 2698](https://github.com/exercism/python/issues/2698) 

[...]

PR #2929, opened 2022-02-13
two-bucket: Update test generation template (#2916)

mhorod:
Fixing issue #2916 

Test generator was failing because the new tests contain `error` field which wasn't handled by the old JinJa2 template.
No changes to `example.py` were required.


BethanyG (approved):
Many thanks for submitting this, @mhorod!  🌟

PR #2933, opened 2022-02-16
Update queen-attack to be clearer for non-chess players

T-O-R-U-S:
- Added examples
- Added an explanation of why the queens are labelled B and W


BethanyG:
Hi @T-O-R-U-S 👋🏽 

Thank you for taking the time to submit this PR!  

Unfortunately,  since `Queen Attack` is a practice exercise, all proposed changes to the instructions proper need to be made in the [problem specifications](https://github.com/exercism/problem-specifications) repo.  Only Python Track-specific changes would be made here.

Suggest you open this proposed change in [problem-specifications](https://github.com/exercism/problem-specifications), so it can be discussed cross-track.  Thanks!

PR #2935, opened 2022-02-17
meltdown-mitigation: Swapped expected and actual arguments in subtests

### LINKED ISSUE #2934: [Meltdown-Mitigation] Expected and actual were swapped in self.assertEqual
I'm pretty new to contributing in general, but I'm excited to have the chance to work on this awesome platform with you all and I hope I can be of some use. I remember working on the Meltdown Mitigation problem a while back while using PyCharm and the expected and actual results were reversed in the testing tool output, so the value that is returned by the FUT would be the expected value and the actual would be expected, you get the picture. 

expected-actual1

Just a quick swap around of the expected and actual_result fixes it. 

expected-actual2

I have a commit ready to go on my repo, I just didn't want to miss anything before creating a pull request. Let me know if there's anything I can do to get the ball rolling. 

Edit: I noticed after mulling over this a bit that all of the tests follow a similar pattern, so if this was intended, feel free to just close this unless you'd like someone to go through and change them one by one. Regards.

Metallifax:
When using the testing tool in PyCharm, the expected and actual results are reversed in the output, so the value that is returned by the FUT is the expected value and the actual is expected. For example: 

```
Expected True but returned None with T=750 and neutrinos=650
True != None

Expected :None
Actual   :True
```

compared to:

```
Expected True but returned None with T=750 and neutrinos=650
None != True

Expected :True
Actual   :None
```

I noticed that many tests in the concepts folder follow this similar pattern, let me know if you'd like to incrementally change each of these tests so that there is some compatibility with PyCharm/JetBrains based editors and I'd be happy to contribute.

 closes #2934

### DIFF
--- a/exercises/concept/meltdown-mitigation/conditionals_test.py
+++ b/exercises/concept/meltdown-mitigation/conditionals_test.py
@@ -32,7 +32,7 @@ class MeltdownMitigationTest(unittest.TestCase):
                 actual_result = is_criticality_balanced(temp, neutrons_emitted)
                 failure_message = (f'Expected {expected} but returned {actual_result} '
                                    f'with T={temp} and neutrinos={neutrons_emitted}')
-                self.assertEqual(actual_result, expected, failure_message)
+                self.assertEqual(expected, actual_result, failure_message)
 
     @pytest.mark.task(taskno=2)
     def test_reactor_efficiency(self):
@@ -54,7 +54,7 @@ class MeltdownMitigationTest(unittest.TestCase):
                 actual_result = reactor_efficiency(voltage, current, theoretical_max_power)
                 failure_message = (f'Expected {expected} but returned {actual_result} '
                                    f'with voltage={voltage}, current={current}, max_pow={theoretical_max_power}')
-                self.assertEqual(actual_result, expected, failure_message)
+                self.assertEqual(expected, actual_result, failure_message)
 
     @pytest.mark.task(taskno=3)
     def test_fail_safe(self):
@@ -73,4 +73,4 @@ class MeltdownMitigationTest(unittest.TestCase):
                 actual_result = fail_safe(temp, neutrons_per_second, threshold)
                 failure_message = (f'Expected {expected} but returned {actual_result} with T={temp}, '
                                    f'neutrons={neutrons_per_second}, threshold={threshold}')
-                self.assertEqual(actual_result, expected, failure_message)
+                self.assertEqual(expected, actual_result, failure_message)


exercises/concept/meltdown-mitigation/conditionals_test.py
```diff
@@ -32,7 +32,7 @@ def test_is_criticality_balanced(self):
                 actual_result = is_criticality_balanced(temp, neutrons_emitted)
                 failure_message = (f'Expected {expected} but returned {actual_result} '
                                    f'with T={temp} and neutrinos={neutrons_emitted}')
-                self.assertEqual(actual_result, expected, failure_message)
+                self.assertEqual(expected, actual_result, failure_message)
```
kotp, line 35: I am not a python testing expert, but I would expect that actual comes first and then expected, and then the failure message (optional).

Are you sure this correction is a correction?

BethanyG:
Hi @Metallifax 👋🏽 

Thanks for the offer of help and for submitting this PR.  Unfortunately, I am going to close it.  You will find both our concept and practice exercise tests use the [unittest syntax](https://docs.python.org/3/library/unittest.html?highlight=unittest#basic-example) of `self.assertEqual(actual, expected, msg)`.  The one exception is our `lists` concept exercise, which needs to be corrected.

The tests were designed  to use PyTest as a runner on the command line and as our test runner inside docker -- not PyCharms test tool.  In fact, PyCharm [has a bug open](https://youtrack.jetbrains.com/issue/PY-26471) about this "swapped" behavior with regards to `unittest`, but they have been ignoring it since 2017.

Suggest you use PyTest as the test runner to avoid the "swapped" behavior.  At some point in the future, we are planning to convert our tests to PyTest syntax (which can be configured to swap in PyCharm)- but it will be a while before we do so.

exercises/concept/meltdown-mitigation/conditionals_test.py
BethanyG: @kotp -- the documentation for the latest Python version (3.10.2) of [`unittest.TestCase`](https://docs.python.org/3/library/unittest.html?highlight=unittest#basic-example) also shows  `AssertEqual(actual, expected, msg)` -- as does the [quick example](https://docs.pytest.org/en/7.0.x/#a-quick-example) in the Pytest documentation with `assert actual == expected`.

Metallifax:
I appreciate the swift reply, and actually didn't even expect to get eyes looking at this in such a short time, so I appreciate it. Just wanted to say thanks for clearing this up as this has been a strange occurrence happening since getting on PyCharm and as you said, only happens with UnitTest, I'll be much more thorough next time.

PR #2945, opened 2022-02-19
Fix bug in test case for test_value_of_ace

natehouk:
A variant of the test_value_of_ace test case is incorrect. If you have a '2' and 'A' in your hand, and you receive a third card of an 'A', the value of the 'A' received is 11 not 1.


BethanyG:
Hi @natehouk 👋🏽 

Thanks for submitting this PR.  🙂   Per the [instructions](https://github.com/exercism/python/blob/main/exercises/concept/black-jack/.docs/instructions.md#3-calculate-the-value-of-an-ace) for this exercise, and the [notes in the stub file](https://github.com/exercism/python/blob/main/exercises/concept/black-jack/black_jack.py) (_see excerpt below_), the value of an Ace _**already in the hand**_ is 11.  Therefore, the test case as currently written is correct.



As such, I am closing this PR -- but would encourage you to submit future PRs or issues, should you see something amiss.

PR #2947, opened 2022-02-20
Making the Grade: Added newlines in code file docstrings

Metallifax:
Consistent with other code files in concepts that include a docstring

### DIFF
--- a/exercises/concept/card-games/.docs/instructions.md
+++ b/exercises/concept/card-games/.docs/instructions.md
@@ -103,7 +103,7 @@ False
 Every 11th hand in Black Joe is a bonus hand with a bonus rule: if the last card you draw is a Jack, you double its value.
 
 Implement a function `maybe_double_last(<hand>)` that takes a hand and checks if the last card is a Jack (11).
- If the the last card **is** a Jack (11), double its value before returning the hand.
+ If the last card **is** a Jack (11), double its value before returning the hand.
 
 ```python
 >>> hand = [5, 9, 11]
--- a/exercises/concept/making-the-grade/.meta/exemplar.py
+++ b/exercises/concept/making-the-grade/.meta/exemplar.py
@@ -12,7 +12,7 @@ def round_scores(student_scores):
 
 
 def count_failed_students(student_scores):
-    """Count the number of failing students in the provided list.
+    """Count the number of failing students out of the group provided.
 
     :param student_scores: list of integer student scores.
     :return: integer count of student scores at or below 40.
--- a/exercises/concept/making-the-grade/.meta/exemplar.py
+++ b/exercises/concept/making-the-grade/.meta/exemplar.py
@@ -1,5 +1,6 @@
 def round_scores(student_scores):
-    """
+    """Round all student scores in the provided list.
+
     :param student_scores: list of student exam scores as float or int.
     :return: list of student scores *rounded* to nearest integer value.
     """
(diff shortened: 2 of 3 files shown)


BethanyG:
Hi @Metallifax - Thanks for submitting this PR!

Unfortunately, I can't accept this as-is.  The "extra line"  is predicated on a summary sentence for the very first line of the docstring.  Since these docstrings are missing that summary sentences, they don't need the addition of extra white space.  See [PEP257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).

However, if you'd like to add summary first-line sentences in this stub file and the `.meta/exemplar.py` file -- I would happily accept and merge them.  😄   

LMK if you'd like to take that on -- and if you have any questions or issues around what is appropriate to put there.

Metallifax:
@BethanyG Just so I'm super clear, based on the documentation provided, you'd like something that looks like...

```python
def round_scores(student_scores):
    """Round all student scores in the provided list.

    :param student_scores: list of student exam scores as float or int.
    :return: list of student scores *rounded* to nearest integer value.
    """

    pass
```
... and for both stub and exemplar (my mistake btw). 

If so, I'm very game. I've been pretty much filling in the summary sentences while completing exercises and my formatting anxiety kicked in when I saw that missing line. Let me know if something like this is too simplistic or too close to what the return value is saying. I might be doing something similar on the Java track and would love the practice.

BethanyG:
@Metallifax  -- Your example looks perfect!  I don't think that's too close to what the return is saying.  And if on second read it is -- we can tidy it up during PR review.  😄 

I look forward to seeing your updated PR!  🌈

Metallifax:
@BethanyG Looking forward to the feedback -- very fun to make a more involved change. Just so I know, whats the etiquette on the length of commit messages if any? Thanks again.

BethanyG (commented):
@Metallifax  -- Thank you so much for these.  I suggested some rephrases.  I wanted to stay away from stating params and return type in the summary since they're also spelled out in the docstring...if that makes sense?  Feel free to accept or reject as you see fit.

BethanyG:
@Metallifax  

> _Just so I know, whats the etiquette on the length of commit messages if any? Thanks again._

Not exactly sure what you mean here.  If you mean the message that goes with the commit using `git` -- it depends on if you are using `git commit -m` or `git commit -v`/`git commit -c` (_The last two open an editor for longer comments_).  

If you are using the first option (`git commit -m`), that is for summary only, and it should be brief and to the point.  

If the second (_`git commit -v`/`git commit -c` -- or other options that open editors_), that provides for both a summary and a body -- in which case, you can be more detailed in the body portion.  See [git commit](https://git-scm.com/docs/git-commit) for an exhaustive rundown.

But if what you mean is "how many commits/changes in a PR" -- That is up to you.  Generally, we want to keep PRs to one exercise or closely related set of changes unless you've discussed some sort of cross-exercise change with maintainers.

Also -- having hundreds of changes within even one exercise does make things hard to review -- but we don't necessarily have limits beyond "common sense" -- if that clarifies?  We also frown on "bug hunting" via mass spell checking or linting.

Metallifax:
@BethanyG I meant as in the commit -m, yes like the message itself. Sorry if you got spammed there, I've never did suggestion committing before but I think I've gotten over that hump already. I implemented all your changes but one with the extra period and just committed that change myself. Should be all there now.

BethanyG (approved):
@Metallifax   Nice work, you!  Thanks again for your work.  🌟

Metallifax:
@BethanyG Cheers!

PR #2958, opened 2022-02-27
[Making the Grade]: Made summary first lines sentences in each docstr…

Metallifax:
…ing use active voicing

- For example, create -> creates, or round -> rounds, etc.
- Removed an extra quotation mark at the start of round_scores docstring

### DIFF
--- a/exercises/concept/making-the-grade/.meta/exemplar.py
+++ b/exercises/concept/making-the-grade/.meta/exemplar.py
@@ -1,5 +1,5 @@
 def round_scores(student_scores):
-    """Round all provided student scores.
+    """Rounds all provided student scores.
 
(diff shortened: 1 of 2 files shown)


IsaacG:
Note [PEP-257](https://www.kite.com/python/docs/pep257.PEP257Checker.check_imperative_mood) examples are all imperative and [pydocs checker D401](https://www.kite.com/python/docs/pep257.PEP257Checker.check_imperative_mood) explicitly enforced imperative mood. This would break with both of those.

BethanyG:
@Metallifax - Isaac is correct here.  I am getting my active and imperative voices confused. I am going to close this one.  Thanks for your patience with these!

PR #2961, opened 2022-02-27
[Ellens Alien Game] Per discussion in issue 2954, Updated Tests for Task 2 and Code Files

BethanyG:
- Changed test for task number 2 to check if health value is in a set of values, rather than one value.
- Changed Stub file docstring.
- Changed Exemplar file to include a note about two viable solutions for hit method task, and added docstrings.

### DIFF
--- a/exercises/concept/ellens-alien-game/.meta/exemplar.py
+++ b/exercises/concept/ellens-alien-game/.meta/exemplar.py
@@ -1,24 +1,39 @@
+"""Exemplar solution to Ellen's Alien Game exercise."""
+
+
 class Alien:
-    """A class to represent an Alien.
+    """Create an Alien object with location x_coordinate and y_coordinate.
 
     Attributes
     ----------
     (class)total_aliens_created: int
     x_coordinate: int - Position on the x-axis.
     y_coordinate: int - Position on the y-axis.
-    health: int - Amount of helath points.
+    health: int - Amount of health points.
 
     Methods
     -------
-    hit(): Decrements Alien health by one point.
-    is_alive(): Returns a boolena for if Alien is alive.
-    teleport(new_x_coordinate, new_y_coordinate): Moves Alien to new coordinates.
+    hit(): Decrement Alien health by one point.
+    is_alive(): Return a boolean for if Alien is alive (if health is > 0).
+    teleport(new_x_coordinate, new_y_coordinate): Move Alien object to new coordinates.
     collision_detection(other): Implementation TBD.
     """
 
     total_aliens_created = 0
 
     def __init__(self, x_coordinate, y_coordinate):
+        """Initialize a new Alien object and increment total_aliens_created by 1.
+
+        :param x_coordinate: int - Alien position on the x-axis
+        :param y_coordinate: int - Alien position on the y-axis
+
+        :attribute x_coordinate: int - Alien position on the x-axis
+        :attribute y_coordinate: int - Alien position on the y-axis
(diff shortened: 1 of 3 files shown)


exercises/concept/ellens-alien-game/.meta/exemplar.py
```diff
@@ -1,47 +1,90 @@
...
-    health: int - Amount of helath points.
+    health: int - Amount of health points.
 
     Methods
     -------
-    hit(): Decrements Alien health by one point.
-    is_alive(): Returns a boolena for if Alien is alive.
-    teleport(new_x_coordinate, new_y_coordinate): Moves Alien to new coordinates.
+    hit(): Decrement Alien health by one point.
+    is_alive(): Return a boolean for if Alien is alive (if health is > 0).
```
kotp, line 17:
```suggestion
    is_alive(): Return a boolean to indicate if Alien is alive (if health is > 0).
```

kotp (approved):

zepam:
I am working on this exercise and when I ran the tests, got this error:
```
We received the following error when we ran your code:

  ImportError while importing test module '.mnt.exercism-iteration.classes_test.py'.
Hint: make sure your test modules.packages have valid Python names.
Traceback:
.usr.local.lib.python3.9.importlib.__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
.mnt.exercism-iteration.classes_test.py:4: in <module>
    from classes import new_aliens_collection
E   ImportError: cannot import name 'new_aliens_collection' from 'classes' (.mnt.exercism-iteration.classes.py)
```

BethanyG:
Hi @zepam 👋🏽   

If you look at the last line of the stack trace, PyTest is saying that it cannot find/import a function named `new_aliens_collection` from your `classes.py` file.  Since it cannot import one of the items under test, it cannot run _any_ tests for the `classes.py` file until the problem is corrected.

Double check the name of your last (_outside the class_) function that's connected to [*task 7*](https://github.com/exercism/python/blob/main/exercises/concept/ellens-alien-game/.docs/instructions.md#7-object-creation).  We had an error in the instructions and test file around the expected function name that was just recently corrected, so you may have the wrong function name.  The function name should be `new_aliens_collection`, and it should be _outside_ the `Alien` class (_per the instructions_).

I think if you supply that function definition -- even with `pass` in the body -- PyTest will be able to import the file and run the tests and show you passes or fails accordingly.

Let me know if you have further issues.  Additionally, you can always submit your code via the [CLI](https://exercism.org/docs/using/solving-exercises/working-locally) to request mentoring even if the test aren't passing.

zepam:
Hi @BethanyG  👋🏽
Ahhh, yes... I see that now. I misunderstood "standalone function" to mean something inside the class rather than outside it. I made that change and it passes. Thanks for your patient and thorough answer! I should have known it wasn't a bug with your code. 🤦🏻

kotp:
@zepam I did want to mention (in case you did not know) that you can submit "broken code" via the command line tool `exercism`, request mentoring, and then hopefully the mentor will be able to work through things like this, if you are "stuck".

It is not currently allowed via the web-editor to do so, but we do give an "out" for the times that you are honestly stuck with errors like this by way of the command line tool.

This PR being not really the right avenue for requesting this kind of help.

But we hope you also do not feel gun-shy about opening an issue on the track, if you do find that the instructions are confusing, or there is a problem we need to work through, and hopefully resolve through code or content changes.

That said, I do welcome you to open an issue if you find that the wording could use some clarification.

PR #2964, opened 2022-03-01
[Track Planning] Exercise and Concept Tree Updates

BethanyG:
Ugly, evil markdown from hell in a magic combination with an epic mermaid chart.  😈


Metallifax:
Thank you for the column that tracks the docstrings on such short notice, and your courage for taking on that markup.

kotp (approved):
You have already (effectively) applied the patch I sent.

PR #2968, opened 2022-03-06
Update lasagna.py

jonathanjohnson8:


kotp:
This file should remain an unsolved exercise file, rather than giving a solution to the student.

Unless I am misunderstanding the intention, this seems like a solution being proposed, but that solution should instead be submitted on the platform.

If you would like feedback on it from a mentor, then request mentoring there.

jonathanjohnson8:
Hi Victor!

My apologies. I now see that we submit solutions via the editor on the platform.

At first, I thought we submitted solutions via a pull request.

Best,
Jon

Jon M. Johnson
Stanford University | open loop university | Undergraduate Student
B.S. Candidate - Symbolic Systems

________________________________

PR #2975, opened 2022-03-11
Fix Python syntax in example code

Stigjb:
Replace backticks with quotes


BethanyG (approved):
Thanks for the catch!  🌟

PR #2981, opened 2022-03-15
clarify instructions on function add_me_to_the_queue

bkc:
The current instructions for `add_me_to_the_queue()` suggest that the existing queue (list) should be updated by appending person_name, then that queue (list) should be returned from the function.

However the test for this function only checks to see if the returned list has the expected items, it doesn't check to see if the returned list is the same list as  was passed in.

That is, its not clear if the solution should be:

```python
return [normal_queue, express_queue][ticket_type] + [person_name]
```

or if it should be

```python
[normal_queue, express_queue][ticket_type].append(person_name)
return [normal_queue, express_queue][ticket_type]
```

My change to the instruction makes it clearer that a copy of the list is to be returned. I'm unsure if the concept of 'copy' should be introduced in this stage or not.

### DIFF
--- a/exercises/concept/chaitanas-colossal-coaster/.docs/instructions.md
+++ b/exercises/concept/chaitanas-colossal-coaster/.docs/instructions.md
@@ -16,7 +16,7 @@ You have been asked to write some code to better manage the guests at the park.
 
 ## 1. Add me to the queue
 
-Define the `add_me_to_the_queue()` function that takes 4 parameters `<express_queue>, <normal_queue>, <ticket_type>, <person_name>` and returns the appropriate queue updated with the person's name.
+Define the `add_me_to_the_queue()` function that takes 4 parameters `<express_queue>, <normal_queue>, <ticket_type>, <person_name>` and returns a copy of the appropriate queue updated with the person's name.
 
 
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @bkc 👋🏽 

Thanks for filing this PR.  As you noted, we don't check and do not say in the instructions that we expect a copy of the passed-in list.  

That's actually by design, since one of the [objectives of this exercise](https://github.com/exercism/python/blob/main/exercises/concept/chaitanas-colossal-coaster/.meta/design.md) was to teach the `list.append()` method, which does indeed _mutate in place_.  The [exemplar](https://github.com/exercism/python/blob/main/exercises/concept/chaitanas-colossal-coaster/.meta/exemplar.py) for this exercise uses `list.append()` for `add_me_to_the_queue`, for that reason.

Now, shame on me (_the test writer_) for not checking I get the same list back that I passed in.  But to be fair, we didn't exactly set up the exercise or the tests in that way here.  The important bit was the use of `list.append()`.  Consider some additional checks on the "to-do" list as of now. 😄  

Meanwhile, I am going to close this particular PR as "not needed".  But please feel free to submit another PR or file an issue should you see anything amiss as you continue with the exercises.  Especially if you'd like to take on writing some tests for this exercise that **do** check to see if the passed in list is the same as the returned one.  😉

PR #2987, opened 2022-03-27
fix: Typo in pytest command

EduardoTT:
It was missing a `-` in `-f`:
it was `pytest -x -ff bob_test.py` and should be `pytest -x --ff bob_test.py`

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -148,7 +148,7 @@ cd path/to/exercise
 Then, run the tests together with the previously explained arguments `-x` and`--ff`:
 
 ```bash
-pytest -x -ff bob_test.py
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @EduardoTT 👋🏽 

Thank you so much for submitting this PR!  Unfortunately, you've caught us in the middle of a big re-write of these docs (_I am about to submit the PR here in an hour or so_).  I think this change was picked up in the re-write.  I would merge this PR in, but the structure and naming of things has also changed in the doc, so this command doesn't read the same way anymore.

While I am closing this PR, I do encourage you to submit another, should you find further typos or issues.  Thanks again for your work. 💛

PR #2995, opened 2022-04-01
[TOOLS.md]: Formatting updates

BethanyG:
- [x] Reformatted disclaimer as an admonition.
- [x] Reformatted TOC with manual reflinks.
- [x] Cleaned up and shortened conda terminal examples.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -7,7 +7,7 @@ This is a collection of tools that are popular in our community.
 It is not intended to be prescriptive nor exhaustive.
 We think these tools do their job well, but there are most certainly other tools that are also good.
 If you have an editor, IDE, tool, or plugin recommendation, we encourage you to add it to this document on GitHub.
-Exercism does not have any financial affiliation with any of the tools listed below.
+Exercism does not have any affiliations with any of the tools listed below.
 
 ~~~~
 
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -1,8 +1,15 @@
 # Tools
 
-A list of tools, IDEs, and editors that can help with writing and debugging Python code.
 
-<br>
+~~~~exercism/noe
+
+This is a collection of tools that are popular in our community.
+It is not intended to be prescriptive nor exhaustive.
+We think these tools do their job well, but there are most certainly other tools that are also good.
+If you have an editor, IDE, tool, or plugin recommendation, we encourage you to add it to this document on GitHub.
+Exercism does not have any financial affiliation with any of the tools listed below.
+
+~~~~
 
 ---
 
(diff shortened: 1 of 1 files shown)


docs/TOOLS.md
```diff
@@ -1,8 +1,15 @@
...
-A list of tools, IDEs, and editors that can help with writing and debugging Python code.
 
-<br>
+~~~~exercism/noe
+
+This is a collection of tools that are popular in our community.
+It is not intended to be prescriptive nor exhaustive.
+We think these tools do their job well, but there are most certainly other tools that are also good.
+If you have an editor, IDE, tool, or plugin recommendation, we encourage you to add it to this document on GitHub.
+Exercism does not have any financial affiliation with any of the tools listed below.
```
kotp, line 10: What kind of affiliations does Exercism have with these tools?  If we are only disclaiming a financial affiliation instead of any affiliation, it suggests that we are indeed affiliated in some way.  It would be good to enumerate the ways.

kotp (approved):

docs/TOOLS.md
BethanyG: What would be your suggestion here?  Should we remove the affiliation note, or remove financial, or .... ??

BethanyG, line 10:
```suggestion
Exercism does not have any affiliations with any of the tools listed below.
```

docs/TOOLS.md
```diff
@@ -1,8 +1,15 @@
 # Tools
 
-A list of tools, IDEs, and editors that can help with writing and debugging Python code.
 
-<br>
+~~~~exercism/noe
```
glennj, line 4: @BethanyG , typo? `exercism/note`

(cannot resist, sorry) I have admonished your admonition

BethanyG: I have duly accepted you admonishment for the admonition accident.

docs/TOOLS.md
kotp: If we have no affiliation then we could state so. If we enumerate one, we should enumerate them all.  If none, then we should remove `financial`.  I am not positive if the Exercism organization does financially or by making upstream submissions/contributions to some of these tools, therefore forming some kind of affiliation.  We may even use them, so "affiliation" may be attributed just by that action.

Unless for some regulation it is imperative that we list affiliations for financial benefits to be listed either to or from something.

Also, if we are stating this, we should really make sure that we have not made any financial donations or any of these have made any financial donations or payments to Exercism.

What I am suggesting is that we be careful when we make such a claim that it is not fraudulent, that we probably need to double check with someone that is intimately aware of the financial conditions and considerations that are made.  And I am not that person, but am aware that there may be some things that I can not see, and this seems to be a claim made on behalf of the organization that I am not sure is clearly true.

BethanyG: The comment was written by Job originally, and its spirit was intended to point out that we are not pushing students toward solutions, or demanding they use one.  And we took no promotional $$ (at the time) to do so.  But if you feel any mention of affiliation is fraudulent, then the whole line can be removed.

kotp: I will do that.  Not because I feel it is fraudulent, just that it seems like we may not know for certain.  There are better ways to state that we have no motivation other than a sense of wanting to provide useful information.  (I think that the second line before this one does that just fine, without making promises that we may not be aware of either way.)

PR #2997, opened 2022-04-01
[Darts]: Fixes for reflinks and nested bullets

BethanyG:
General hints do not appear to support either reflink nor nested bullets, so I re-arranged the file to omit them.


BethanyG:
Placed this on hold, as it may be a bug with how hints display.  Will file an issue for the hint display, and depending on that resolution, will close or request review on this PR.

kotp (approved):

Issue #3003, opened 2022-04-11
[Guidos Gorgeous Lasagna] Correction to Instructions


jalriols:
The first step of the exercise consist on defining a constant. In the instructions there is an example where the the constant is called as follows:
```python
>>> lasagna.EXPECTED_BAKE_TIME
```

I believe that the constant was probably an atribute from the lasagna class in an earlier version of the problem, and now it just makes it difficult to understand what "lasagna" means in the exercise.

BethanyG:
Hi @jalriols 👋🏽 

Thanks for submitting this issue.

While it may be confusing (_and we may need to make the example clearer by explicitly showing the import_), it is not wrong here to say `lasagna. EXPECTE_BAKE_TIME`.  It wasn't left over from a `lasagna` **_class_**, rather `lasagna` is the name of the exercise **_file/module_**.

 The convention in Python is to use  **`<module name>`.`<object name>`** if you are referring to a variable, constant, or class from an imported file/module - which you would be, if you were accessing the EXPECTED_BAKE_TIME constant from the `lasagna py` file in a  Python terminal (_which is what those `>>>` signify in the code example_)

You could also do `from lasagna import EXPECTED_BAKE_TIME` to avoid having to reference `lasagna`, but we chose not to go over that here.

I also left this note on the PR -- so if you'd like to add an import statement to the code example as clarification, I'd happily review and accept it -- but the current usage of `lasagna.EXPECTED_BAKE_TIME` is correct.

jalriols:
Hi @BethanyG, thanks for your thorough explanation.

I think you expect me to modify the first example to look like this:
```python
>>> import lasagna
>>> lasagna.EXPECTED_BAKE_TIME
40
```

But then I think I should also add the _import_ for the rest of the functions to be consistent, like the example below:

```python
>>> from lasagna import bake_time_remaining
>>> bake_time_remaining(30)
10
```

Another option would be leaving only the first import, but calling all the functions with the complete name:

```python
>>> lasagna.bake_time_remaining(30)
10
```

What do you think?

BethanyG:
Hi @jalriols, thanks for your thoughtful reply!  

Now that I am looking at it, I like the idea of adding the more specific import for the functions as well.  That gives two different examples of imports -- the whole file/module vs a specific function from it.

But as a learner, is that going to be confusing?  If not, then I say go with your top two examples above - and many thanks for filing the issue and doing the work!  😄

PR #3004, opened 2022-04-11
Remove lasagna from the constant name (EXPECTED_BAKE_TIME)

jalriols:
The first part of the exercise consist on defining a constant. In the instructions there is an example where the the constant is called as follows:
```python
>>> lasagna.EXPECTED_BAKE_TIME
```
I believe that the constant was probably an atribute from the lasagna class in an earlier version of the problem, and now it just makes it difficult to understand what "lasagna" means in the exercise. 
#3003

### DIFF
--- a/exercises/concept/guidos-gorgeous-lasagna/.docs/instructions.md
+++ b/exercises/concept/guidos-gorgeous-lasagna/.docs/instructions.md
@@ -10,7 +10,7 @@ Define an `EXPECTED_BAKE_TIME` constant that returns how many minutes the lasagn
 According to your cookbook, the Lasagna should be in the oven for 40 minutes:
 
 ```python
->>> lasagna.EXPECTED_BAKE_TIME
+>>> EXPECTED_BAKE_TIME
 40
 ```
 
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @jalriols 👋🏽 

Thanks for submitting this PR.  While it may be confusing (_and we may need to make the example clearer by explicitly showing the import_), it is not wrong here to say `lasagna. EXPECTE_BAKE_TIME`.  It wasn't left over from a `lasagna` **_class_**, rather `lasagna` is the name of the exercise **_file/module_**.

 The convention in Python is to use  **`<module name>`.`<object name>`** if you are referring to a variable, constant, or class from an imported file/module - which you would be, if you were accessing the EXPECTED_BAKE_TIME constant from the `lasagna py` file in a  Python terminal (_which is what those `>>>` signify in the code example_)

You could also do `from lasagna import EXPECTED_BAKE_TIME` to avoid having to reference `lasagna`, but we chose not to go over that here.

So if you'd like to add in the import to the example, that would be great.  Otherwise, I am going to close this PR as "won't fix, since the current usage is correct, if a bit confusing.

BethanyG:
Putting this on hold pending the discussions and changes in #3003.

BethanyG:
Putting this on hold pending the discussions and changes in #3003.

BethanyG (approved):
@jalriols  - _excellent!_ Thank you so much for filing the issue, and for PR-ing these.  🌟 🌈

PR #3016, opened 2022-04-20
high-scores: handle HighScores.scores tests

IsaacG:
Support the scores property tests. Add myself as a contributor.


IsaacG:
Related to issue #2786

BethanyG:
I like it.  _Technically_, we should append new test cases in an append file, but this one is a bit on the line.....I think I am going to let it slide.  However, if we end up adding more cases, the will definitely need to be documented in `additional_tests.json`.

BethanyG (approved):
🌟

PR #3026, opened 2022-04-29
[Simple Linked List]: Added instruction append and hints

### LINKED ISSUE #3009: [Single Linked List] Head concept doesn't seem to match descriptions
Reading the descriptions for Linked Lists in general, and even in the cited documentation for the exercise give us the idea of a list that would work as [1,2,3] where 1 would be the head node and 3 would be the tail node, hence follow LIFO. However, all the tests seem to expect that the HEAD would be 3, and, for instance, any new node pushed into the list would be the new head. 
Given that even the exercise itself links to documentation which contradicts the expected results, either the explanation is not clear enough (or I'm not very good at understanding it) or the tests should be somewhat changed. I've confirmed that with a mentor as well.

BethanyG:
Per discussions/questions in  #3009.

Added resources, explanations, and hints for the exercise.


exercises/practice/simple-linked-list/.docs/instructions.append.md
```diff
@@ -1,16 +1,87 @@
 # Instructions append
 
+## How this Exercise is Structured in Python
+
+While `stacks` and `queues` can be implemented using `lists`, `collections.deque`, `queue.LifoQueue`, and `multiprocessing.Queue`, this exercise expects a ["Last in, First Out" (`LIFO`) stack][Baeldung: The Stack Data Structure] (_interactive example [here][LIFO Stack]_) using a _custom-made_ [singly linked list][singly linked list].
+
+<br>
+
+_push and pop happen here_
```
kotp, line 9: This line indicates that both happen here, but it only addresses one…

exercises/practice/simple-linked-list/.docs/instructions.append.md
```diff
@@ -1,16 +1,87 @@
...
+<br>
+
+
+This should not be confused with a [`LIFO` stack using a dynamic array or list][LIFO Stack Array], which may use a `list` underneath.
+Dynamic array based `stacks` have a different `head` position and different time complexity (Big-O) and memory footprint.
+
+<br>
+
+
+&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp; &nbsp; &nbsp;&nbsp; &nbsp; &nbsp;  _push and pop happen here_
```
kotp, line 25: … and then the other.

kotp (changes requested):
Only one statement from me, otherwise I think it looks really good!

exercises/practice/simple-linked-list/.docs/instructions.append.md
BethanyG: It addresses neither.  The representation is meant to show the `next` pointer between nodes - not the push nor pop actions. 

Unfortunately, I am limited by the ASCII art here.  But if you can find a way to cleanly show push and pop actions here within how it is written, I am open to updating it.  Otherwise, we need to either embed a full diagram (_which then becomes difficult to read when the exercise is downloaded via CLI_), or remove this ascii art as an example.

The confusing point here for many is that a `stack` underpinned by a Linked List has the `head` on the "left" -- not that push and pop happen _at_ the head.  See discussions under #3009 for why this was included in the first place. 

For a diagram, we could custom-make a variant on the below, but I'd rather not.

exercises/practice/simple-linked-list/.docs/instructions.append.md
BethanyG: This was only included as an example of what the opposite looks like.  I don't want to overly confuse students with this one, since it is not something that will pass the tests for the exercise.  It is only shown to give students a warning of sorts that `stacks` and `queues` can be implemented from either direction.  But same note as above applies here as well.  We can include a diagram, but that has implications for how students who download the exercise will read things.

BethanyG: For either this one or the other - please let me know.  It will take about an hour of my time to make custom diagrams and figure out how and where we upload them to include them.  AFIK, we don't support mermaid in our markdown, so doing a mermaid diagram here is out.

exercises/practice/simple-linked-list/.docs/instructions.append.md
kotp: I saw the pop and shift happens here, and the vertical line which I thought was indicating the exact position.  So that was the confusing part, and then only the one vertical line, which I thought the "and" part was missing from either, or the "and" was simply extraneous.

That that I know it shows neither... a bit more confusing to read, I think.

I do agree with the hard to read in the terminal if we have a graphic representation.  I do see it as a mermaid friendly problem, but how many people can read mermaid charts in the terminal as plain text and see what it is meant to depict?  Not sure.

Thinking about what would look good in markdown as rendered and as plain text…

I will discard my review until I can think of a solution for that.

exercises/practice/simple-linked-list/.docs/instructions.append.md
```diff
@@ -1,16 +1,87 @@
...
+While `stacks` and `queues` can be implemented using `lists`, `collections.deque`, `queue.LifoQueue`, and `multiprocessing.Queue`, this exercise expects a ["Last in, First Out" (`LIFO`) stack][Baeldung: The Stack Data Structure] (_interactive example [here][LIFO Stack]_) using a _custom-made_ [singly linked list][singly linked list].
+
+<br>
+
+_push and pop happen here_
+
+&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;|
+
+
+(_head_) **`[Node_3]` -> `[Node_2]` -> `[Node_1]`** (_tail_) -> None
```
kotp, line 14:
```suggestion
    _push and pop happens here_
                           |
                           |
                           |+----------------------------------\
                                                                |
                                                                v
    +------+    +--------+     +--------+     +--------+    +------+
    | head | ** | Node_3 |---->| Node_2 |---->| Node_1 | ** | tail |
    +------+    +--------+     +--------+     +--------+    +------+

```

Not the best, I am not an artist.  But something like this?

And now I understand what was trying to be communicated, it is the point at which push and pop happen, as opposed to differentiating with push vs shift, etc.

exercises/practice/simple-linked-list/.docs/instructions.append.md
kotp: No great confidence in the ascii art, and the trade off would be the italics that are wanted in the rendering.   But the non-breaking spaces would not be needed as well.

exercises/practice/simple-linked-list/.docs/instructions.append.md
BethanyG: I ... don't think that is very readable, and having `head` and `tail` in their own boxes implies that they are nodes, when really `Node_1` is the tail node, and `Node_3` is the head node.

exercises/practice/simple-linked-list/.docs/instructions.append.md
BethanyG: I've inserted SVG diagrams.  I've added descriptions, so hopefully that makes them somewhat accessible.  Students who download will either have to follow the links, or open the images to see.

BethanyG:
@ErikSchierboom  -- a few questions for you (_because there is nothing I love more than being a problem child_ 😆  ):

1.  These `.svg` files were under 1MB in size, so I decided to **not** upload them to `exercism/v3-files` -- should I have uploaded them anyway?
2.  Not at all sure how I should organize or reference images in things like exercise appends and hints files.  Can I ref-link images?  Should the files live in an `img` folder in the exercise .. or be loose files?
3.  Was it OK that I referenced the `.doc` folder for the image used in the `.meta/hints` file?
4.  The images won't render in the UI right now because they aren't (yet) at the URL.  Is there another way I can embed or reference them...or should I leave the links as-is? 
5.   Are these better as in-line SVG? (_I an guessing no_)

exercises/practice/simple-linked-list/.docs/instructions.append.md
kotp: I think that is reasonable.

If the location for the image could be `/.exercism/images/linked-list-stack.svg` this would even render in (I think most) local markdown preview solutions for people.  I confirmed for some locally for me.  If it were delivered to the already existing directory of `.exercism` I think that would work pretty well.  It is where we put our meta data anyway.

ErikSchierboom (commented):
> These .svg files were under 1MB in size, so I decided to not upload them to exercism/v3-files -- should I have uploaded them anyway?

I'm not sure we have a policy, but the existing files there are of a similar size to the one you're uploading (see e.g. https://github.com/exercism/v3-files/blob/main/python/string-formatting/decision_tree.png), so I think uploading to `v3-files` makes sense.

> Not at all sure how I should organize or reference images in things like exercise appends and hints files. Can I ref-link images? Should the files live in an img folder in the exercise .. or be loose files?


> Was it OK that I referenced the .doc folder for the image used in the .meta/hints file?
No, because the `.docs` and `.meta` folders are treated specially and their files won't be downloaded via the CLI (well, they're rendered as README.md, HELP.md and HINTS.md files).

> The images won't render in the UI right now because they aren't (yet) at the URL. Is there another way I can embed or reference them...or should I leave the links as-is?
If you'll upload them to v3-files, you can reference them with a full URL: https://github.com/exercism/julia/blob/4f06b8f7ab0a67de5d817d04732ed5db1c3033bd/exercises/practice/killer-sudoku-helper/.docs/instructions.md

> Are these better as in-line SVG? (I an guessing no)
No I don't think so :)

ErikSchierboom:
And sorry for the delay in replying!

BethanyG:
@ErikSchierboom   I reopened the PR to `v3_files`.  I can neither request review nor get at a URL to link to.  What is the process there for review/approval, and how do I get a link for the docs here?  Do I need to wait for the v3 PR to be approved?

BethanyG:
Ok.  I think things are finally sorted (_both here, and in my brain_) 😉 .  Re-requesting review.  Links to images should resolve now.

kotp (approved):

bobahop (approved):
Double-checked graphic descriptions and verified all links correctly work.

PR #3028, opened 2022-04-29
[Card Games] Additional test for Task 6.

### LINKED ISSUE #3025: [Card Games] Test coverage for Task 6.
The current test cases for Task 6 (for `def average_even_is_average_odd(hand)`) allows for `return card_average(hand) % 2 == 0` to pass all tests, however this is probably not the intention of this particular Task, and as written a counter example that should return `True` but returns `False` for this given solution would be `[1, 3, 5]`.

tsigma6:
Adding in a test for `[1, 3, 5, 7, 9]` to prevent an invalid alternate solution for Task 6.

Alternate solution that passes all current tests: `return card_average(hand) % 2 == 0`.

Fixes #3025 .

### DIFF
--- a/exercises/concept/card-games/lists_test.py
+++ b/exercises/concept/card-games/lists_test.py
@@ -85,9 +85,9 @@ class CardGamesTest(unittest.TestCase):
     @pytest.mark.task(taskno=6)
     def test_average_even_is_average_odd(self):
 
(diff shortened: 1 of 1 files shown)


BethanyG (approved):
@tsigma6 - Nice work!  Many thanks for submitting this (and for your original issue).  🌟

Issue #2992, opened 2022-03-31
[Ellen's Alien Game] Tests are Blocked from Running in Editor & Locally Due to Function Import Error


lotruheawea:
Dear community,

working on Ellen's Alien Game, the tests do not run and I get this message:

> We received the following error when we ran your code:
> 
>   ImportError while importing test module '.mnt.exercism-iteration.classes_test.py'.
> Hint: make sure your test modules.packages have valid Python names.
> Traceback:
> .usr.local.lib.python3.9.importlib.__init__.py:127: in import_module
>     return _bootstrap._gcd_import(name[level:], package, level)
> .mnt.exercism-iteration.classes_test.py:4: in <module>
>     from classes import new_aliens_collection
> E   ImportError: cannot import name 'new_aliens_collection' from 'classes' (.mnt.exercism-iteration.classes.py)

There is `from classes import new_aliens_collection` in line 4 in `classes_test.py`. Why is it there?

lotruheawea:
Ah, the last instruction calls for the creation of function `new_aliens_collection`. Well, there are two ways, I guess. Either

- create a dummy implementation of `new_aliens_collection` in the default `classes.py` or
- use a `try`-`except` clause like in lines 6-10.

Having just done the exercise, I think it is good to have the tests running step by step. So the second option would mean they fail with an `ImportError`...

If you want, I can create a PR implementing the first option:

```
def new_aliens_collection():
     """

    :param coordinate_pairs: tuple of (x_coordinate: int, y_coordinate: int)
    :return: list of (Alien instances)

    This function should return a `list` of Alien instances,
    each initialized with a single coordinate pair
    """
    pass
```

BethanyG:
Hi @lotruheawea 👋🏽 

Thanks for filing this issue!  There is a bit of a dilemma  here, since this exercise was intended to teach both class creation, and some basics around using or calling the class. We also wanted to show that Python supports both defining a `class` and defining stand-alone functions within the same module. Pre-defining the stand-alone function stub short-circuits that learning somewhat.

But students are getting confused with the wording of task 7:

<img width="1116" alt="2022-03-31_08-12-19" src="https://user-images.githubusercontent.com/5923094/161089569-89be1dc8-200b-437c-8845-c9cbe1d3f06f.png">

And, as you pointed out, the import error for `new_alien_collection` makes it look as though the tests are "broken", unless you are used to decoding Python stack traces.

[...]

PR #3030, opened 2022-04-30
[Ellens Alien Game] Split files

PaulT89:
See Issue #2992

As expected, the tests choked.  Thoughts?


PaulT89:
Actually, new idea - keep the exemplar as-is (it's there for mentors anyway), and split the actual exercise files.

BethanyG:
Yeah. I don't think this strategy works.  When I said "fence", I meant a `try/except` in the test file, not splitting into multiple code and test files.  We'll need to do something else.  Apologies that I was not clearer with that in my issue comment.

BethanyG:
Splitting the files (stub or exemplar) is going to lead to the same failure.  Our test runner is not expecting to have to test sequentially.  It can cope with importing something from somewhere else...but the file under test needs to be ONE file.  Maybe someday it will..but not today.  But additionally, I don't think two files improves understanding - I think that adds student confusion.

My thoughts are wrapping the imports in a `try/except` and also putting a TODO in the stub.  Maybe also being pointed in the directions.  See the tests for [Guidos Gorgeous Lasagna](https://github.com/exercism/python/blob/main/exercises/concept/guidos-gorgeous-lasagna/lasagna_test.py) for an example of what I meant by wrapping, or "fencing".

PaulT89:
Oh well, I guess I'm out of ideas then.  Sorry.

BethanyG:
Ok.  Thank you for trying tho.  🙂   Apologies I can't alter the test runner at this time.  When I have a chance, I will PR what I am thinking, and ask you for feedback, if you'd like?

PaulT89:
I'm not sure I'd be able to contribute much - I don't know much of anything about test-runners

BethanyG:
It would be a change to the test file for the exercise, and possibly the task descriptions and intro.  I am not going to change the test runner to be multi-solution file at this time. 🙂

PaulT89:
Okay, well, I guess I'll try to help in whatever way I can.

PR #3055, opened 2022-05-17
Logic Error in odds average == evens average?

eddieurfaust:
In the last test for `test_average_even_is_average_odd(self)` ,[1, 3, 5, 7, 9] no even numbers exist, so hand == []. Here `False` should be expected instead of `True`.

### DIFF
--- a/exercises/concept/card-games/lists_test.py
+++ b/exercises/concept/card-games/lists_test.py
@@ -87,7 +87,7 @@ class CardGamesTest(unittest.TestCase):
 
         input_vars = [[5, 6, 8], [1, 2, 3, 4], [1, 2, 3], [5, 6, 7], [1, 3, 5, 7, 9]]
 
-        results = [False, False, True, True, True]
+        results = [False, False, True, True, False]
 
         for variant, (hand, same) in enumerate(zip(input_vars, results), start=1):
             error_message = f'Hand {hand} {"does" if same else "does not"} yield the same odd-even average.'


exercises/concept/card-games/lists_test.py
```diff
@@ -87,7 +87,7 @@ def test_average_even_is_average_odd(self):
 
         input_vars = [[5, 6, 8], [1, 2, 3, 4], [1, 2, 3], [5, 6, 7], [1, 3, 5, 7, 9]]
 
-        results = [False, False, True, True, True]
+        results = [False, False, True, True, False]
```
SleeplessByte, line 90: Why do you expect `False` ?

It seems that the instructions talk about off and even _positions_. In the array `[1, 3, 5, 7, 9]` there are both odd-numbered positions and even-numbered positions.

eddieurfaust: You are absolutely right! I didn't read the instructions carefully enough.

Still, separating the odds and evens satisfies all test cases. 

```python
evens = [card for card in hand if card % 2 == 0]  
odds = [card for card in hand if card % 2 != 0]  
  
if not evens or not odds:  
    return False  
  
return card_average(evens) == card_average(odds)
```

Should I open another PR for this or are the exercise test cases ok as they are?

SleeplessByte: Is there a test case input that you can craft that would break currently valid solutions?

If you can, it _may_ be worth opening another PR.

For this specific case though, the average of 0 items isn't really defined in the instructions and thus i would not expect a test case for it (to keep the exercise simple and focused towards the learning goal(s)).

PR #3064, opened 2022-05-24
[Conditionals]: Update introduction.md & about.md

JAButenegro:
When explaining elif (changes in lines 48, 52 and 54):
x > y doesn't necessarily guarantee x > z directly, therefore first check-in (if) should be x > y and y > z, while second check-in (elif) could be y > z, therefore else would return z > x and z > y
Counterexample: x = 15, y = 10, z = 20. The current program (since x > y) would return "x is greater than y and z", when x >z is not true. The revised code would also evaluate x > z, therefore returning "z is greater than x and y"

### DIFF
--- a/concepts/conditionals/introduction.md
+++ b/concepts/conditionals/introduction.md
@@ -44,9 +44,11 @@ y = 10
 z = 20
 
 # The elif statement allows for the checking of more conditions.
-if x > y and x > z:
+if x > y > z:
+
     print("x is greater than y and z")
-elif y > z:
+elif y > x > z:
+
     print("y is greater than x and z")
 else:
     print("z is greater than x and y")
--- a/concepts/conditionals/introduction.md
+++ b/concepts/conditionals/introduction.md
@@ -44,14 +44,14 @@ y = 10
 z = 20
 
 # The elif statement allows for the checking of more conditions.
-if x > y:
+if x > y and x > z:
     print("x is greater than y and z")
 elif y > z:
     print("y is greater than x and z")
 else:
-    print("z is great than x and y")
+    print("z is greater than x and y")
 ...
->>> z is great than x and y
+>>> z is greater than x and y
 ```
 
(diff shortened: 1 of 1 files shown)


concepts/conditionals/introduction.md
```diff
@@ -44,14 +44,14 @@ y = 10
 z = 20
 
 # The elif statement allows for the checking of more conditions.
-if x > y:
+if x > y and x > z:
```
BethanyG, line 47:
```suggestion
if x > y > z:

```

concepts/conditionals/introduction.md
```diff
@@ -44,14 +44,14 @@ y = 10
 z = 20
 
 # The elif statement allows for the checking of more conditions.
-if x > y:
+if x > y and x > z:
     print("x is greater than y and z")
 elif y > z:
```
BethanyG, line 49:
```suggestion
elif y > x > z:

```

BethanyG (changes requested):
@JAButenegro - Thank you for catching this error and for submitting a PR.  Two requested changes:
1.  Python allows three-way comparison, so we should probably be using that here.
2.  This introduction document displays before the student finishes an exercise, but we also have an `about.md` that displays _after_.  could you also make these changes in that document as well?

Many thanks!  🌈

JAButenegro:
Hi @BethanyG, thanks for the feedback! It's actually my first day on GitHub, but I guess the changes you asked me to do to the about.md file has been completed (you might want to check that anyway). 
Thanks also for the amazing work in Exercism! ❤️

BethanyG (approved):
@JAButenegro I went ahead and accepted the changes to this PR, so you don't have to dig through the help docs (_its a confusing interface_).  🙂   Thanks for submitting this PR! 🌟

JAButenegro:
Great! Highly appreciated! 💯

PR #3066, opened 2022-05-24
[List Methods]: Fix Markdown Heading in Hints File

BTowersCoding:
I was going through this exercise and noticed that one of the hints was not being displayed due to a missing `#`.


BethanyG:
@BTowersCoding  🤦🏽‍♀️  _Thank you so much for catching that & submitting a PR!_  I think Erik is talking about setting up an org-wide linting action to catch things like this, since omitting a `#` means something disappears from the site.

BethanyG (approved):

PR #3067, opened 2022-05-26
[Update Concept] functions, and complete function arguments

bobahop:
Updating the functions concept to prepare for the functions concept exercise. This includes completing the function arguments concept, which is currently TODO. Function arguments will be lightly covered in functions, but a link to the function arguments concept will be provided, so something should be in there, I think.

### DIFF
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -149,7 +149,7 @@ def my_function(*args, **kwargs):
 
 `*args` is a two-part name that represents a `tuple` with an indefinite number of separate positional arguments, also known as a [`variadic argument`][variadic argument].
 
-`args` is the name of the group of arguments and could be any other name, such as `my_args`, `arguments`, etc.
+`args` is the name given to the `tuple` of arguments,  but it could be any other valid Python name, such as `my_args`, `arguments`, etc.
 The `*` is the operator which transforms the group of separate arguments into a [`tuple`][tuple].
 Since a tuple can be iterated, `args` can be passed to any other function which takes an iterable.
 Although `*args` is commonly juxtaposed with `**kwargs`, it doesn't have to be.
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -194,7 +194,8 @@ It did not reassemble the elements into a tuple.
 
 `**kwargs` is a two-part name that represents an indefinite number of separate [key-value pair][key-value] arguments.
 `kwargs` is the name of the group of arguments and could be any other name, such as `my_args`, `arguments`, etc.
-The `**` is the operator which transforms the group of separate arguments into a [`dictionary`][dictionary].
+The `**` transforms the group of named arguments into a [`dictionary`][dictionary] of `{argument name: argument value}` pairs.
+
 Since a dictionary can be iterated, `kwargs` can be passed to any other function which takes an iterable.
 Although `**kwargs` is commonly juxtaposed with `*args`, it doesn't have to be.
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -93,7 +93,8 @@ Hello, you
 
 ```
 
-Requiring keyword-only arguments for function calls can be done through the use of the `*` operator in the parameter list:
+Requiring keyword-only arguments for function calls can be done through the use of the `*` operator in the parameter list.
+
 
 Following is an example of keyword-only arguments:
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -88,8 +88,6 @@ Hello, Bob
 # Function call with positional data as arguments.
 >>> print(concat("Hello, ", "Bob"))
 Hello, Bob
->>> print(concat())
-Hello, you
 
 ```
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -198,9 +198,6 @@ TypeError: can only concatenate list (not "int") to list
 
 ```
 
-What happened is that the `*` in `*args` only unpacked the list into its separate elements, and it was done.
-It did not reassemble the elements into a tuple.
-
 ## `**kwargs`
 
 `**kwargs` is a two-part name that represents an indefinite number of separate [key-value pair][key-value] arguments.
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -181,7 +181,35 @@ For instance, `*` is used for multiplication, it is used for unpacking, and it i
 Since a tuple can be iterated, `args` can be passed to any other function which takes an iterable.
 Although `*args` is commonly juxtaposed with `**kwargs`, it doesn't have to be.
 
-Following is an example of an arbitrary amount of values being passed to a function between a positonal argument and a keyword argument:
+Following is an example of an arbitrary amount of values being passed to a function:
+
+```python
+
+>>> def add(*args):
+# args is passed to the sum function, which takes an iterable
+...         return sum(args)
+... 
+>>> print(add(1, 2, 3))
+6
+```
+
+If `*args` follows one or more positional arguments, then `*args` will be what is left over after the positional arguments.
+
+Following is an example of an arbitrary amount of values being passed to a function after a positional argument:
+
+```python
+
+>>> def add(first, *args):
+# first will be 1, leaving the values 2 and 3 in *args
+...         return first + sum(args)
+... 
+>>> print(add(1, 2, 3))
+6
+```
+
+If one or more default arguments are defined after `*args` they are separate from the `*args` values.
+
+To put it all together is an example of an arbitrary amount of values being passed to a function that also has a positional argument and a default argument:
 
 ```python
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -149,9 +149,14 @@ def my_function(*args, **kwargs):
 ```
 
 `*args` is a two-part name that represents a `tuple` with an indefinite number of separate positional arguments, also known as a [`variadic argument`][variadic argument].
-
 `args` is the name given to the `tuple` of arguments,  but it could be any other valid Python name, such as `my_args`, `arguments`, etc.
 The `*` is the operator which transforms the group of separate arguments into a [`tuple`][tuple].
+
+If you have ever [unpacked a tuple][unpack a tuple] you may find the `*` to be confusing.
+The `*` in a parameter definition, instead of unpacking a tuple, converts one or more positional arguments _into_ a tuple.
+We say that the `*` operator is [overloaded]. as it has different behavior in different contexts.
+For instance, `*` is used for multiplication, and it is also used for unpacking.
+    
 Since a tuple can be iterated, `args` can be passed to any other function which takes an iterable.
 Although `*args` is commonly juxtaposed with `**kwargs`, it doesn't have to be.
 
@@ -166,6 +171,8 @@ Following is an example of an arbitrary amount of values being passed to a funct
 6
 >>> print(add(1, 2, 3, last=4))
 10
+# This uses the unpacking operator * to separate the list elements into positional arguments.
+# It is not the same operator as the * in *args.
 >>> print(add(*[1, 2, 3]))
 6
 
@@ -231,6 +238,8 @@ one two three
 [dictionary]: https://www.w3schools.com/python/python_dictionaries.asp
 [function concept]: ../functions/about.md
 [key-value]: https://www.pythontutorial.net/python-basics/python-dictionary/
+[overloaded]: https://www.geeksforgeeks.org/operator-overloading-in-python/
 [tuple]: https://www.w3schools.com/python/python_tuples.asp
+[unpack a tuple]: https://www.geeksforgeeks.org/unpacking-a-tuple-in-python/
 [unpacking operator]: https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists
 [variadic argument]: https://en.wikipedia.org/wiki/Variadic_function
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -10,9 +10,9 @@ Parameter names should not contain spaces or punctuation.
 ## Positional Arguments
 
 Positional arguments are for parameters defined without a [default argument][default arguments].
-Positional arguments can optionally be called by their keyword.
+Positional arguments can optionally be called by their name.
 
-Following is an example of positional arguments being called by position and by their keyword:
+Following is an example of positional arguments being called by position and by their name:
 
 ```python
 >>> def concat(x, y):
@@ -27,7 +27,7 @@ Hello, Bob
 ```
 
 The first call to `concat` passes the arguments by position.
-The second call to `concat` passes the arguments by keyword, allowing their positions to be changed.
+The second call to `concat` passes the arguments by name, allowing their positions to be changed.
 
 Note that positional arguments cannot follow keyword arguments.
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -10,7 +10,7 @@ Parameter names should not contain spaces or punctuation.
 ## Positional Arguments
 
 Positional arguments are values passed to a function in the same order as the parameters which bind to them.
-Positional arguments can optionally be called by using their parameter name.
+Positional arguments can optionally be passed by using their parameter name.
 
 Following is an example of positional arguments being passed by position and by their parameter name:
 
@@ -64,9 +64,9 @@ TypeError: concat() got some positional-only arguments passed as keyword argumen
 ## Keyword Arguments
 
 Keyword arguments are for parameters defined with a [default argument][default arguments].
-Keyword arguments can optionally be called by their position.
+Keyword arguments can optionally be passed by their position.
 
-Following is an example of keyword arguments being called by their keyword and by position:
+Following is an example of keyword arguments being passed by their keyword and by position:
 
 ```python
 >>> def concat(greeting="Hello, ", name="you"):
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -115,6 +115,23 @@ TypeError: concat() takes 0 positional arguments but 2 were given
 
 ```
 
+## Default Argument Values
+
+Default values for one or more arguments can be supplied in the parameter list.
+This allows the function to be called with _fewer_ arguments if needed.
+Default values can be overridden by calling the function with a new argument value, overriding the default:
+
+```python
+# Function with default argument values.
+>>> def concat(greeting, name="you", punctuation="!"):
+...         return f"{greeting}, {name}{punctuation}"
+... 
+>>> print(concat("Hello"))
+Hello, you!
+
+# Overriding the default values
+>>> print(concat("Hello", name="Bob", punctuation="."))
+Hello, Bob.
 ## Positional or Keyword Arguments
 
 Arguments can be positional or keyword if neither the `/` nor `*` operators are used in the parameter definitions.
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -183,7 +183,8 @@ Following is an example of an arbitrary amount of values being passed to a funct
 ```
 
 Note that when an argument is already in an iterable, such as a tuple or list, it needs to be unpacked before being passed to a function that takes an arbitrary amount of separate arguments.
-This is accomplished by `*`, which is the [unpacking operator][unpacking operator].
+This is accomplished by using `*`, which is the [unpacking operator][unpacking operator].
+
 `*` in this context _unpacks_ the container into its separate elements which are then transformed by `*args` into a tuple.
 Where there are only positional arguments, the unpacking action must result in the same number of arguments as there are formal parameters.
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -228,4 +228,5 @@ one two three
 [key-value]: https://www.pythontutorial.net/python-basics/python-dictionary/
 [tuple]: https://www.w3schools.com/python/python_tuples.asp
 [unpacking operator]: https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists
+[variadic argument]: https://en.wikipedia.org/wiki/Variadic_function
 [variadic argument]: https://medium.com/spatial-data-science/unlock-the-power-variadic-arguments-in-python-functions-a591bf572c2
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -229,4 +229,3 @@ one two three
 [tuple]: https://www.w3schools.com/python/python_tuples.asp
 [unpacking operator]: https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists
 [variadic argument]: https://en.wikipedia.org/wiki/Variadic_function
-[variadic argument]: https://medium.com/spatial-data-science/unlock-the-power-variadic-arguments-in-python-functions-a591bf572c2
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -169,14 +169,13 @@ def my_function(*args, **kwargs):
 `args` is the name given to the `tuple` of arguments,  but it could be any other valid Python name, such as `my_args`, `arguments`, etc.
 The `*` is the operator which transforms the group of separate arguments into a [`tuple`][tuple].
 
-----
+~~~~exercism/note
 If you have ever [unpacked a tuple][unpack a tuple] you may find the `*` in `*args` to be confusing.
 The `*` in a _parameter_ definition, instead of unpacking a tuple, converts one or more positional arguments _into_ a tuple.
 
 We say that the `*` operator is [overloaded], as it has different behavior in different contexts.
 For instance, `*` is used for multiplication, it is used for unpacking, and it is used to define an arbitrary number of positional parameters.
-
-----
+~~~~
     
 Since a tuple can be iterated, `args` can be passed to any other function which takes an iterable.
 Although `*args` is commonly juxtaposed with `**kwargs`, it doesn't have to be.
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -9,8 +9,8 @@ Parameter names should not contain spaces or punctuation.
 
 ## Positional Arguments
 
-Positional arguments are for parameters defined without a [default argument][default arguments].
-Positional arguments can optionally be called by their name.
+Positional arguments are values passed to a function in the same order as the parameters which bind to them.
+Positional arguments can optionally be called by using their name.
 
 Following is an example of positional arguments being called by position and by their name:
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -111,7 +111,7 @@ TypeError: concat() takes 0 positional arguments but 2 were given
 ## Positional or Keyword Arguments
 
 Arguments can be positional or keyword if neither the `/` nor `*` operators are used in the parameter definitions.
-Alternately, the positional-or-keyword arguemtns can be placed between the positional-only parameters on the left and the keyword-only parameters on the right.
+Alternately, the positional-or-keyword arguments can be placed between the positional-only parameters on the left and the keyword-only parameters on the right.
 
 Following is an example of positional-only, positional-or-keyword, and keyword-only arguments:
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -223,6 +223,8 @@ Following is an example of an arbitrary amount of key-value pairs being passed t
 >>> print(concat(one=1, two=2, three=3))
 one two three
 
+```
+
 
 [default arguments]: https://www.geeksforgeeks.org/default-arguments-in-python/
 [dictionary]: https://www.w3schools.com/python/python_dictionaries.asp
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -213,16 +213,16 @@ Note that the `dict.values()` method is called to iterate through the `kwargs` d
 
 When iterating a dictionary the default is to iterate the keys.
 
-Following is an example of an arbitrary amount of key-value pairs being passed to a function that iterates the keys:
+Following is an example of an arbitrary amount of key-value pairs being passed to a function that then iterates over `kwargs.keys()`:
 
 ```python
 >>> def concat(**kwargs):
+             # Join concatenates the key names from `kwargs.keys()`
 ...         return " ".join(kwargs)
 ... 
 >>> print(concat(one=1, two=2, three=3))
 one two three
 
-```
 
 [default arguments]: https://www.geeksforgeeks.org/default-arguments-in-python/
 [dictionary]: https://www.w3schools.com/python/python_dictionaries.asp
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -59,10 +59,13 @@ Following is an example of positional-only arguments:
 ... 
 >>> print(concat("Hello, ", "Bob"))
 Hello, Bob
+...
+# Call to the function using keyword arguments.
 >>> print(concat(name="Bob", greeting="Hello, "))
 Traceback (most recent call last):
     print(concat(name="Bob", greeting="Hello, "))
 TypeError: concat() got some positional-only arguments passed as keyword arguments: 'greeting, name'
+
 
 ```
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -209,7 +209,8 @@ Following is an example of an arbitrary amount of key-value pairs being passed t
 6
 ```
 
-Note that the `values()` method is called to iterate on the dictionary values.
+Note that the `dict.values()` method is called to iterate through the `kwargs` dictionary values.
+
 When iterating a dictionary the default is to iterate the keys.
 
 Following is an example of an arbitrary amount of key-value pairs being passed to a function that iterates the keys:
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -18,7 +18,9 @@ Following is an example of positional arguments being called by position and by 
 >>> def concat(greeting, name):
 ...         return f"{greeting}{name}"
 ... 
+# Passing data to the function by position.
 >>> print(concat("Hello, ", "Bob"))
+
 Hello, Bob
 >>> print(concat(name="Bob", greeting="Hello, "))
 Hello, Bob
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -47,7 +47,8 @@ results in this error:
 SyntaxError: positional argument follows keyword argument
 ```
 
-Requiring positional-only arguments for function calls can be done through the use of the `/` operator in the parameter list:
+Requiring positional-only arguments for function calls can be done through the use of the `/` operator in the parameter list.
+
 
 Following is an example of positional-only arguments:
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -136,6 +136,7 @@ Traceback (most recent call last):
     print(concat(greeting="Hello, ", name="Bob", ending="!"))
 TypeError: concat() got some positional-only arguments passed as keyword arguments: 'greeting'
 
+```
 
 ## `*args`
 
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -22,7 +22,10 @@ Following is an example of positional arguments being passed by position and by 
 >>> print(concat("Hello, ", "Bob"))
 
 Hello, Bob
+...
+# Passing data to the function using the parameter name.
 >>> print(concat(name="Bob", greeting="Hello, "))
+
 Hello, Bob
 
 ```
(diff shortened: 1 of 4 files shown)


concepts/functions/introduction.md
```diff
@@ -1,17 +1,105 @@
...
+
+```
+
+The parameter is defined as `name`.
+`"Bob"` is the argument which the program passes to the `hello` function.
+The program outputs `Hello, Bob`.
+
+What if someone calls `hello` without passing an argument?
+The program would error with a report that an argument is missing.
+If we don't want the program  to error with no argument, we can define a [`default argument`][default arguments].
```
BethanyG, line 75:
```suggestion
If we don't want the program to error with no argument (_but suspect that the calling code might not supply one_), we can define a [`default argument`][default arguments].
```

concepts/functions/introduction.md
```diff
@@ -1,17 +1,105 @@
...
+The default argument defines what value to use if the argument is missing when the function is called.
```
BethanyG, line 76:
```suggestion
A default argument defines what value to use if the argument is missing when the function is called.
```

concepts/functions/introduction.md
```diff
@@ -1,17 +1,105 @@
...
+
+Following is an example of a function with a default argument
```
BethanyG, line 78:
```suggestion
Following is an example of a function with a default argument:
```

concepts/functions/introduction.md
```diff
@@ -1,17 +1,105 @@
...
+
+```python
+def hello(name="you"):
+    return f"Hello, {name}"
+
+
+print(hello())
+
+```
```
BethanyG, line 87:
```suggestion
```python
>>> def hello(name="you"):
...        return f"Hello, {name}"
...
>>> print(hello())
Hello, you
```
```

File: concepts/functions/introduction.md [hunk 5]
```diff
@@ -1,17 +1,105 @@
...
+
+The `name` parameter is given the default argument of `"you"`, so the program outputs `Hello, you`.
```
BethanyG, line 89:
```suggestion
The `name` parameter is given the default argument of `"you"`, so the program outputs `Hello, you`, if not passed a `name` argument.
```

BethanyG (commented):
@bobahop   This is great!  Thank you so much for revising this and moving it forward.  💙   I've left a bunch of little nits.  The only "required" change I'd request is that we use REPL formatting for these small code examples.  I'd also add something in here somewhere about Python being an OOP language, and so _functions are objects_.  Beyond that, this looks really good.

BethanyG:
🤦🏽‍♀️   I also forgot .. we probably need to mention scoping/namespacing in here somewhere......

concepts/functions/introduction.md
```diff
@@ -1,17 +1,102 @@
 # Introduction
 
-A [`function`][function] is a named block of code that takes input and produces output when invoked. Functions are used to perform specific and repetitive tasks throughout a program.
+A [`function`][function] is a way to compartmentalize code so it can be called from one or more places in the program.
+Functions are used to perform specific and repetitive tasks throughout a program.
 
-More formally:  a function is any Python object to which the [`function call`][calls] operation can be applied. The body of the function contains one or more statements which will return some value to the calling code/caller. The function object can define/receive zero or more parameters/arguments (_including other functions_).
+More formally: a function is any Python object to which the [`function call`][calls] operation can be applied.
+A function may be used to [`return`][return] one or more values as a result of some operation(s), or it may be used for one or more [`side effects`][side effects].
+If a function does not specify a return value it will still return `None`.
```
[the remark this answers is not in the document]
bobahop: I think for ia Intro it can just be described as `None`. I don't think this is the place to introduce the concept of singleton object.

concepts/functions/introduction.md
bobahop: I like the direction of this suggestion, but I think better would be

If we don't want the program to error with no argument (_but want to allow the calling code to not supply one_), we can define a [`default argument`][default arguments].

bobahop:
Thank you for REPLing the examples. If I had seen the About.md first I might have done that.

concepts/functions/introduction.md
BethanyG: I think that the edit I made will allow you to commit the suggestion.  Let me know if it doesn't.

bobahop: Thank you for the effort, but I committed a manual edit yesterday, so this isn't needed. Sorry, I should have commented I did that!

BethanyG: No worries, as long as it's done.  😄   I'll mark resolved then.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+one two three
+
+```
+
+[default arguments]: https://www.geeksforgeeks.org/default-arguments-in-python/
+[dictionary]: https://www.w3schools.com/python/python_dictionaries.asp
+[function concept]: ../functions/about.md
+[key-value]: https://www.pythontutorial.net/python-basics/python-dictionary/
+[tuple]: https://www.w3schools.com/python/python_tuples.asp
+[unpacking operator]: https://docs.python.org/3/tutorial/controlflow.html#unpacking-argument-lists
```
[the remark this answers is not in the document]
BethanyG: So I think this needs a bit of work, and some (maybe) re-arranging.  In particular, I think we need to be very clear about what happens when you define a parameter with a default value.   

Doing so allows the function to be called with fewer arguments (_because the default argument value is already assigned to the parameter name_) BUT **default values are read in only once**, which has both positive and negative implications.  Default values can be overridden (yay), but if they are _data structures_, their value is shared between calls, which can lead to .... consequences. 

I also think we need to disentangle the use of `tuple unpacking` in function calls from the `tuple` and `dict` formation that takes place with `*args` and `**kwargs`.  While it is totally true that you can pass & unpack a list for `*args`, I think right now the text confuses the two and might lead a student to misunderstand when they can use unpacking, and when they cannot (_you can't unpack more values than there are parameters, _unless_ the function defines `*args`).

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
 You can also define your own functions. Those are called [`user-defined functions`][user defined functions].
 Functions can run something as simple as _printing a message to the console_ or they can be quite complex.
 
-To execute the code inside a function, you need to call the function, which is done by using the function name followed by parenthesis [`()`].
-Data. known as `parameters`, can be passed to the function by placing them inside the parenthesis. A broader term for parameters is `arguments`. Functions can perform different tasks depending on the value of the parameters.
+To execute the code inside a function, you need to call the function, which is done by using the function name followed by parenthesese [`()`].
+Data. known as [`arguments`][arguments], can be passed to the function by placing them inside the parenthesese.
+In the function definition any arguments are defined by [`parameters`][parameters].
```
BethanyG, line 11:
```suggestion
In the function definition,  accepted arguments are defined via [`parameters`][parameters].
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+Functions can perform different tasks depending on the arguments passed to the parameters.
```
BethanyG, line 12:
```suggestion
Functions can perform different tasks depending on the arguments passed.
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
 
-A function can also return a value using the `return` keyword. The value returned by the function is known as the `return value`. The return value is returned to the caller of the function.
+A function can also return a value using the [`return`][return] keyword.
```
BethanyG, line 14:
```suggestion
A function can also _return_ a value to the function caller by using the [`return`][return] keyword.
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+The value returned by the function is known as the `return value`.
+The return value is returned to the caller of the function.
```
BethanyG, line 16:
```suggestion
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
 
 ## Creation
 
-In python, functions are created using the `def` keyword. The function definition is followed by the function name and parenthesis [`()`]. Inside the parenthesis, the parameters are specified, separated by commas. After the close parenthesis, the colon (`:`) is used to separate the function signature from the function body.
+In python, functions are created using the [`def`][def] keyword.
+The function definition is followed by the function name and parentheses [`()`].
+Inside the parentheses, the parameters are specified, separated by commas.
+After the close parenthesis, the colon (`:`) is used to separate the function signature from the function body.
```
BethanyG, line 23:
```suggestion
After the close parenthesis, the colon (`:`) is used to separate the function _signature_ from the function _body_.
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
 
-The function body is a block of code that is executed when the function is called. The body of the function is indented. The indentation is important because Python relies on it to know where that block of code ends. A value can be returned from the function by using the `return` keyword, which can then be used by the caller of the function.
+The function body is a block of code that is executed when the function is called.
+The body of the function is indented.
```
BethanyG, line 26:
```suggestion
The body of the function is indented (_typically 2-4 spaces_).
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+The indentation is important because Python relies on it to know where that block of code ends.
```
BethanyG, line 27:
```suggestion
Indentation is important because the Python parser relies on it to know where that block of code ends.
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+A value can be returned from the function by using the `return` keyword, which can then be used by the caller of the function.
```
BethanyG, line 28: This is a repetition of the part above, so I think we can remove this sentence.

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
 We can also define a function without any parameters or return value.
 
 ```python
 def function_name():
     # function body
+    pass
+
 ```
 
+Note that the function does need a body, even if the body does nothing, or trying to run the program will generate an indentation error.
```
BethanyG, line 46: Maybe add an example of the error here?  You can't make it happen in the REPL, but when in happens in code, it's fairly inscrutable because all you get is a line number and a random-looking carrot with the error message.  It doesn't even manage to print the function definition.  😱 :

```python
File ~/temp.py:66
    
    ^
IndentationError: expected an indented block
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+
 ## Calling a Function
 
-To call a function, use the function name followed by parenthesis [`()`]. Parameters passed to the function are placed inside the parenthesis, separated by commas.
+To call a function, use the function name followed by parenthesese [`()`].
+Parameters passed to the function are placed inside the parenthesese, separated by commas.
```
BethanyG, line 51:
```suggestion
Arguments (_data_) passed to the function are placed inside the parentheses, separated by commas.
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
 The above function can be called by using the following syntax:
 
 ```python
->>> wish()
+>>> greet()
 Hello
 ```
 
-## Parameters
+## Parameters and their Arguments
```
BethanyG, line 68:
```suggestion
## Parameters and Arguments
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
 
-Parameters are values that are passed to the function when it is called. They can be of any data type.
+Arguments are values that are passed to the function when it is called.
```
BethanyG, line 70:
```suggestion
Arguments are data passed to the function when it is called.
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
-def add(x, y):
-    print(x + y)
+def add(first, second):
+    print(first + second)
+
 ```
 
-When the function is called, the parameters are passed to the function. We need to pass values for both the parameters, otherwise a [`TypeError`][type-error] will be raised.
+The parameters `first` and `second` define what arguments the `add` function will accept.
+(It should be noted that the words `parameter` and `argument` are often used interchangeably, albeit imprecisely.)
```
BethanyG, line 82:
```suggestion
(_It should be noted that the words `parameter` and `argument` are often used interchangeably, albeit imprecisely.  `parameters` are names used in function definition.  `arguments` are data passed during a function call._)
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+When the function is called, the arguments are passed to the function.
```
BethanyG, line 83:
```suggestion
When the function is called, _arguments_ are passed to the function.
```

concepts/functions/about.md
```diff
@@ -6,60 +6,82 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+We need to pass arguments for both of the parameters, otherwise a [`TypeError`][type-error] will be raised.
```
BethanyG, line 84:
```suggestion
Because the function definition outlines two parameters, arguments for both of the parameters will need to be passed on function call - otherwise a [`TypeError`][type-error] will be raised:
```

concepts/functions/about.md
```diff
@@ -85,13 +107,17 @@ TypeError: add() takes 2 positional arguments but 3 were given
 
 ## Return Value
 
-The return value is a value that is returned to the caller of the function. Return value can be of any data type. It can be used by caller of the function to perform further operations. If the function does not explicitly return a value, the value `None` is returned.
+The return value is a value that is returned to the caller of the function.
+Return value can be of any data type.
```
BethanyG, line 111:
```suggestion
A `return value` can be any data type.
```

concepts/functions/about.md
```diff
@@ -85,13 +107,17 @@ TypeError: add() takes 2 positional arguments but 3 were given
...
+It can be used by caller of the function to perform further operations.
+If the function does not explicitly return a value, the value `None` is returned.
```
BethanyG, line 113:
```suggestion
If a function does not explicitly define a `return value`, the value `None` will be returned by the Python interpreter.
```

concepts/functions/about.md
```diff
@@ -170,49 +198,56 @@ Now, we can call the functions in the order we want.
 
 ## Scope of Variables
 
-If variable is defined inside a function, then it will be only accessible inside the function. If we want to access the variable outside the function, we need to use the [`global`][global] keyword. [`nonlocal`][nonlocal] keyword is used to access the variable inside a nested function.
+If variable is defined inside a function, then it will be only accessible inside the function.
+If we want to access the variable outside the function, we need to use the [`global`][global] keyword.
```
BethanyG, line 202:
```suggestion
If we want to access the variable name outside the function scope (_or namespace_), we need to use the [`global`][global] keyword.
```

BethanyG: _ooof_ not sure here if we should go into the dangers of this or just link to them.  The `global` statement is pretty complicated: https://docs.python.org/3/reference/simple_stmts.html#the-global-statement.  What is especially concerning is that it is read in before the run starts and lasts for the life of the program.  The namespace collision risk is .... ugly.

concepts/functions/about.md
```diff
@@ -170,49 +198,56 @@ Now, we can call the functions in the order we want.
...
+[`nonlocal`][nonlocal] keyword is used to access the variable inside a nested function.
```
BethanyG, line 203: This might need a little more elaboration, since this is about referencing enclosing scope that's _not_ global, but not exactly local either and may or may not be important when making closures and other nested functions and structures.

concepts/functions/about.md
```diff
@@ -170,49 +198,56 @@ Now, we can call the functions in the order we want.
...
 ```
 
 ## Functions as first class objects
 
-In python, functions can be assigned to variables and passed as arguments to other functions. They can be used as return values. Functions can also be placed into a sequence([`list`][list], [`tuple`][tuple] etc) or as value in a [`dict`][dict]. Functions can be used anywhere than any other object can be used. This is because _functions are [`first class objects`][first class objects]_.
+In python, functions can be assigned to variables and passed as arguments to other functions.
+They can be used as return values.
+Functions can also be placed into a sequence([`list`][list], [`tuple`][tuple] etc) or as value in a [`dict`][dict].
+Functions can be used anywhere any other object can be used.
+This is because _functions are [`first class objects`][first class objects]_.
```
BethanyG, line 250:
```suggestion
This is because _functions are [`first class objects`][first class objects]_.
As such, they carry special attributes, and can even define getting and setting custom attributes of their own. 
```

BethanyG (changes requested):
@bobahop  💙  Apologies for taking so dang long in getting this review done and thank you **so so much** for all your hard work on this.  🌟 

Overall, I think this is a really solid start to both concepts -- but there is a LOT in the details, so i think we need to discuss/massage still.   I don't really want to break these down into even smaller concepts .. but when you dig into some of this stuff, you realize just how strange Python can be.   I don't know that these need massive changes, but there are some things I think need clarification.  Let me know if you have questions or issues.  Thanks again for doing these write-ups.  😄

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+
+## Parameter Names
+
+Paramater names, like variable names, must start with a letter or underscore and may contain letters, underscores, or numbers.
+Parameter names should not contain spaces or punctuation.
+
+## Positional Arguments
+
+Positional arguments are for parameters defined without a [default argument][default arguments].
+Positional arguments can optionally be called by their name.
```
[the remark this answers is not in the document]
bobahop: Sorry, but I think this is just crufting it up with more verbiage.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+
+Following is an example of positional arguments being called by position and by their name:
```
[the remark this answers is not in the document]
bobahop: Sorry, but I think this is just crufting it up with more verbiage.

concepts/functions/about.md
bobahop: Yeah, I don't do a happy dance re global variables. Some kind of warning is probably a good thing to implant...

concepts/functions/about.md
bobahop: Yeah, this area isn't my strong suit, as I've never needed to use either `global` or `nonlocal`. But my experience is pretty much limited to the Exercism exercises.

Anyway, this is from the original about.md which was originally with this verbiage long ago.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
-#TODO: Add about for this concept.
+# About
 
+For the basics on function arguments, please see the [function concept][function concept].
```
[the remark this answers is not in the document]
bobahop: Yes, I suppose it can be made more clear than it is presently.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+
+## Parameter Names
+
+Paramater names, like variable names, must start with a letter or underscore and may contain letters, underscores, or numbers.
+Parameter names should not contain spaces or punctuation.
+
+## Positional Arguments
+
+Positional arguments are for parameters defined without a [default argument][default arguments].
```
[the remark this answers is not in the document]
bobahop: I was vaguely dissatisfied with this, and I think you've put your finger on why.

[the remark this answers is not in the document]
bobahop: I edited to 

```
Positional arguments are values passed to a function in the same order as the parameters which bind to them.
Positional arguments can optionally be called by using their name.
```

concepts/function-arguments/about.md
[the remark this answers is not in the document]
bobahop: I rephrased.

concepts/function-arguments/about.md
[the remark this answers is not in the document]
BethanyG: I'm sorry -- I really have to disagree.  Arguments aren't _called_.  They're used/passed/referred to.  I think it's important to make the distinction, because some students don't know what a _call_ is.  If you don't like referred, we can pick another word..but using "call" makes it confusing wrt "function call".

concepts/functions/about.md
BethanyG: Well, the original was "optimistically merged".  😄   But when I get a chance, I'll propose something.  I don't know that we need much ...  but maybe a few sentences and an example.

concepts/functions/about.md
BethanyG: For the intro -- yes.  But I do think that for the `about.md` it's useful knowledge.

concepts/function-arguments/about.md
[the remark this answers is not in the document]
bobahop: You are right about this. I would substitute "called" with "passed."

concepts/functions/about.md
bobahop: I don't feel that strongly about it, so I'll accept the change.

concepts/function-arguments/about.md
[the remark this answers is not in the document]
bobahop: Went through the about and intro docs and changed all "called" to "passed"... I think. Sometimes I can look right at something and just not _see_ it. Sorry for the resistance.

concepts/functions/about.md
```diff
@@ -170,49 +201,57 @@ Now, we can call the functions in the order we want.
...
```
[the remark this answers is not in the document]
bobahop: You're right. I think the intent is to indicate that it is a link. It's hard to tell the blue link from regular black text. It would be nice if there was a set way we used to indicate a link that is more noticeable.

concepts/functions/about.md
```diff
@@ -251,9 +290,9 @@ Functions can also be nested inside other functions.
 
 ```python
 def outer():
-    x = 10
+    num = 10
     def inner():
-        print(x)
+        print(num)
```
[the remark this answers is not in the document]
bobahop: I think you may be using an older version of the file, as `x` is no longer used in the code example. The text below did still reference `x`, and that's now been fixed. Thanks!

concepts/functions/introduction.md
```diff
@@ -1,17 +1,105 @@
...
+Hello
 
-Python also provides many top-level [_built-in functions_][built-in functions] for common operations.
+```
 
+In the example above, the [`def`][def] keyword is used to signal the beginning of the function definition.
+`hello` is the name of the function.
+The empty parentheses represent that no values are being passed to the function.
+The body of the function is the single `print("Hello")` statement.
+`print` is also a function.
```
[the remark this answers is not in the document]
bobahop: Thanks. `print` is the name of the function. The parentheses are part of its signature, but not part of the name. And in the example it is not called with empty parentheses. So I'd prefer to leave it as is.

concepts/functions/introduction.md
```diff
@@ -1,17 +1,102 @@
...
 
+hello()
+
+```
+
+The [`def`][def] keyword is used to signal the beginning of the function definition.
+`hello` is the name of the function.
+The empty parentheses represent that no values are being passed to the function.
+The body of the function is the single `print("Hello")` statement.
+`print` is also a function.
```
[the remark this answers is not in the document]
bobahop: I decided to leave as is because `print` is the name of the function and it is not called in the example with empty parentheses.

concepts/functions/introduction.md
```diff
@@ -1,17 +1,102 @@
...
+The `"Hello"`  in the `print` function's parentheses is known as an [`argument`][arguments].
+The argument is used by the `print` function to know what to print. 
+Since the `hello` function does not specify a `return` value, it executes its side effect - which is calling the `print` function - and returns `None`.
```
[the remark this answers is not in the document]
bobahop: There is a link to side effects on line 7.

A function may be used to [`return`][return] one or more values as a result of some operation(s), or it may be used for one or more [`side effects`][side effects].

Although I would be happier with a different article to link to. Okay, I think I found one...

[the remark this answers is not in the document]
bobahop: But it's on Medium, so a limited viewing before paywall, which is why I chose the worse one earlier.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+... 
+>>> print(concat("Hello, ", "Bob"))
+Hello, Bob
+>>> print(concat(name="Bob", greeting="Hello, "))
+Hello, Bob
+
+```
+
+The first call to `concat` passes the arguments by position.
+The second call to `concat` passes the arguments by name, allowing their positions to be changed.
```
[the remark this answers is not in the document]
bobahop: I really think the original is okay here.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+```python
+>>> def concat(greeting="Hello, ", name="you"):
+...         return f"{greeting}{name}"
+... 
+>>> print(concat(name="Bob", greeting="Hello, "))
+Hello, Bob
+>>> print(concat("Hello, ", "Bob"))
+Hello, Bob
+>>> print(concat())
+Hello, you
```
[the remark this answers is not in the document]
bobahop: Will commit independently...

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+Traceback (most recent call last):
+    print(concat("Hello, ", "Bob"))
+TypeError: concat() takes 0 positional arguments but 2 were given
+
+```
+
+## Positional or Keyword Arguments
+
+Arguments can be positional or keyword if neither the `/` nor `*` operators are used in the parameter definitions.
+Alternately, the positional-or-keyword arguemtns can be placed between the positional-only parameters on the left and the keyword-only parameters on the right.
```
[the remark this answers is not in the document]
bobahop: I fixed the spelling independently. I think the original phrasing is okay.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+
+Following is an example of positional-only, positional-or-keyword, and keyword-only arguments:
```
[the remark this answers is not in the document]
bobahop: Leaving it

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+Hello, Bob
+>>> print(concat(name="Bob", greeting="Hello, "))
+Hello, Bob
+
+```
+
+The first call to `concat` passes the arguments by position.
+The second call to `concat` passes the arguments by name, allowing their positions to be changed.
+
+Note that positional arguments cannot follow keyword arguments.
```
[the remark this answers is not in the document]
BethanyG: In a sense, underneath they're all keyword arguments.  But when assigning _default values_ during function definition, you do have to have the default value assignment follow any parameters that _don't_ have default values....so there is a parallel with when a function is called.  Effectively, Python can't tell the difference between the keyword vs positional _unless_ positional always comes first in both calling and in defining functions.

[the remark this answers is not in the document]
BethanyG: The reason I put this here is to try to separate calling from defining, because default arguments are for definition, and keyword arguments are for calling...but as long as we are super-clear in our examples, we don't have to bludgeon people with it here.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+
+results in this error:
+
+```
+SyntaxError: positional argument follows keyword argument
+```
+
+Arguments can be forced to be positional-only through the use of the `/` operator.
+
+Following is an example of positional-only arguments:
```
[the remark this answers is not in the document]
BethanyG: It's a style thing.  If you put a colon on the end of the proceeding sentence, I think it's enough to signal to a reader that an example follows, so you don't need to state that in an additional sentence.  But I'm also fine leaving it if you think it reads better.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+>>> print(concat("Hello, ", "Bob"))
+Hello, Bob
+>>> print(concat())
+Hello, you
+
+```
+
+Arguments can be forced to be keyword--only through the use of the `*` operator.
+
+Following is an example of keyword-only arguments:
```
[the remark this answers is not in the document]
BethanyG: Same comment as earlier.  If you put a colon at the end of the proceeding sentence, it should be enough to signal and example follows.  This sentence then becomes repetitive.  But again if it reads better to you, we can keep it...we just need to then remove the colon from the preceding sentence both here and in the earlier example.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,231 @@
...
+    # code snipped
+
+```
+
+`*args` is a two-part name that represents a `tuple` with an indefinite number of separate positional arguments, also known as a [`variadic argument`][variadic argument].
+
+`args` is the name given to the `tuple` of arguments,  but it could be any other valid Python name, such as `my_args`, `arguments`, etc.
+The `*` is the operator which transforms the group of separate arguments into a [`tuple`][tuple].
+Since a tuple can be iterated, `args` can be passed to any other function which takes an iterable.
+Although `*args` is commonly juxtaposed with `**kwargs`, it doesn't have to be.
```
BethanyG, line 155: Flagging this again because I think it was missed.  This example is problematic because it confuses tuple unpacking with the functionality of `*args`.  While both use an overloaded `*` operator, one _makes_ a tuple from a group of arguments, the other _unpacks_ a data structure into multiple individual variables or arguments.  It is very confusing for students at first, so we really need to separate out the two behaviors.  We need to especially avoid implying that tuple unpacking for function calls needs the `*arg` parameter -- it doesn't.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,214 @@
...
+
+```
+
+What happened is that the `*` in `*args` only unpacked the list into its separate elements, and it was done.
+It did not reassemble the elements into a tuple.
+
+## `**kwargs`
+
+`**kwargs` is a two-part name that represents an indefinite number of separate [key-value pair][key-value] arguments.
+`kwargs` is the name of the group of arguments and could be any other name, such as `my_args`, `arguments`, etc.
```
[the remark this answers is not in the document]
bobahop: `kwargs` is not the _default_ name, but a name often used by convention, so leaving as is.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,245 @@
...
+...
+# Calling the function with positional data raises an error.
+>>> print(concat("Hello, ", "Bob"))
+Traceback (most recent call last):
+    print(concat("Hello, ", "Bob"))
+TypeError: concat() takes 0 positional arguments but 2 were given
+
+
+```
+
```
BethanyG, line 116:
```suggestion

## Default Argument Values

Default values for one or more arguments can be supplied in the parameter list.
This allows the function to be called with _fewer_ arguments if needed.
Default values can be overridden by calling the function with a new argument value, overriding the default:

```python
# Function with default argument values.
>>> def concat(greeting, name="you", punctuation="!"):
...         return f"{greeting}, {name}{punctuation}"
... 
>>> print(concat("Hello"))
Hello, you!

# Overriding the default values
>>> print(concat("Hello", name="Bob", punctuation="."))
Hello, Bob.
```
```

**Comment by bobahop** [comment]
@BethanyG , thank you for adding the default arguments section! I conflated them with keyword arguments, but they deserve their own section.

File: concepts/function-arguments/about.md [hunk 47]
```diff
@@ -1,2 +1,248 @@
...
+----
+If you have ever [unpacked a tuple][unpack a tuple] you may find the `*` in `*args` to be confusing.
+The `*` in a parameter definition, instead of unpacking a tuple, converts one or more positional arguments _into_ a tuple.
+We say that the `*` operator is [overloaded], as it has different behavior in different contexts.
+For instance, `*` is used for multiplication, it is used for unpacking, and it is used to define an arbitrary number of positional parameters.
+
+----
+    
+Since a tuple can be iterated, `args` can be passed to any other function which takes an iterable.
```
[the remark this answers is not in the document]
BethanyG: No.  You're right.  This is much more a "show, don't tell" situation.

concepts/function-arguments/about.md
```diff
@@ -1,2 +1,293 @@
...
+# Function with default argument values.
+>>> def concat(greeting, name="you", punctuation="!"):
+...         return f"{greeting}, {name}{punctuation}"
+... 
+>>> print(concat("Hello"))
+Hello, you!
+
+# Overriding the default values
+>>> print(concat("Hello", name="Bob", punctuation="."))
+Hello, Bob.
```
BethanyG, line 133:
```suggestion
Hello, Bob.
```
```

Comment (bobahop) [reply concepts/function-arguments/about.md:133]: Had to commit this separately.

File: concepts/functions/about.md [hunk 49]
```diff
@@ -6,60 +6,85 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
 We can also define a function without any parameters or return value.
 
 ```python
 def function_name():
     # function body
+    pass
+
 ```
 
+Note that the function does need a body, even if the body does nothing, or trying to run the program will generate an indentation error.
```
BethanyG, line 49:
```suggestion
Note that the function does need a body, even if the body does nothing, or trying to run the program will generate an indentation error:

```python
>>> def my_bodyless_func():
...      

File ~/temp.py:1
    
    ^
IndentationError: expected an indented block
```
The keyword `pass` can be used in the function body to avoid the error.
```

File: concepts/functions/about.md [hunk 50]
```diff
@@ -6,60 +6,85 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+>>> greet()
 Hello
 ```
 
-## Parameters
+## Parameters and their Arguments
 
-Parameters are values that are passed to the function when it is called. They can be of any data type.
+Arguments are values that are passed to the function when it is called.
+They can be of any data type.
```
BethanyG, line 74:
```suggestion
They can be of any data type, including other functions or classes.
```

concepts/functions/about.md
```diff
@@ -170,49 +201,57 @@ Now, we can call the functions in the order we want.
 
 ## Scope of Variables
 
-If variable is defined inside a function, then it will be only accessible inside the function. If we want to access the variable outside the function, we need to use the [`global`][global] keyword. [`nonlocal`][nonlocal] keyword is used to access the variable inside a nested function.
+If variable is defined inside a function, then it will be only accessible inside the function.
+If we want to access the variable name outside the function scope (_or namespace_), we need to use the [`global`][global] keyword.
```
BethanyG, line 205:
```suggestion
If we want to make the variable name accessible outside the function scope (_or namespace_), we need to use either the [`global`][global] or [`nonlocal`][nonlocal] keywords.
Using the `nonlocal` keyword will signal to Python to make the variable available in the _enclosing namespace_, and is most often used when a function is nested inside another function.
Using the `global` keyword signals Python to make the variable available in the _global namespace_.
This should be used cautiously, as any function or process could then modify the variable without notice.
```

concepts/functions/about.md
```diff
@@ -170,49 +201,57 @@ Now, we can call the functions in the order we want.
...
+[`nonlocal`][nonlocal] keyword is used to access the variable inside a nested function.
```
BethanyG, line 206:
```suggestion
```

BethanyG (commented):
Ok.  Some last suggestions.  Beyond these, I think we need a better local -> nonlocal -> global code example progression, but I couldn't suggest one in the interface.  But after that, this is good to go.

concepts/functions/about.md
```diff
@@ -85,13 +110,17 @@ TypeError: add() takes 2 positional arguments but 3 were given
 
 ## Return Value
 
-The return value is a value that is returned to the caller of the function. Return value can be of any data type. It can be used by caller of the function to perform further operations. If the function does not explicitly return a value, the value `None` is returned.
+The return value is a value that is returned to the caller of the function.
+A `return value` can be any data type.
```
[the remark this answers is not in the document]
BethanyG: Wow. Yeah. That is indeed really confusing.  I did not intend at all to delete `It can be used by caller of the function to perform further operations. If the function does not explicitly return a value, the value None is returned.`.  I like that as well.  Let me double check the file....

concepts/functions/about.md
```diff
@@ -6,60 +6,94 @@ Python, like other programming languages, has [_built-in functions_][build-in fu
...
+Note that the function does need a body, even if the body does nothing, or trying to run the program will generate an indentation error:
+
+```python
+>>> def my_bodyless_func():
+...      
+
+File ~/temp.py:1
+    
+    ^
+IndentationError: expected an indented block
```
BethanyG, line 58: formatting error.

BethanyG, line 58:
```suggestion
IndentationError: expected an indented block
```

```

BethanyG (approved):
I am going to approve this so we can merge it and work on a cleaner copy.  Points we still need to address:
1.  We need to remove the code examples for `global` and `nonlocal` and redo them for clarity.
2. There may be some formatting issues that have crept in.  If so, we'll need to address them in a follow-on PR.

concepts/functions/about.md
bobahop: Committed separately.

Issue #3068, opened 2022-05-27
[Tisbury Treasure Hunt]: Feedback


chuck-alt-delete:
Thanks as always for building Exercism! I have two pieces of feedback on tisbury-treasure-hunt:

### 1. Return None or raise error instead of "not a match" in create_record function

The `create_record` function says to return a record if coordinates match or the string "not a match" if the coordinates don't match. In my opinion, this encourages bad habits unnecessarily. I think instead, if there is no matching coordinate, the function should return none or should raise a `ValueError`.  At this point, learners aren't learning about error handling yet, so maybe just go with returning `None`.

### 2. The test for clean_up is inconsistent

Here is a solution that passes the tests:

```python
def clean_up(combined_record_group):
    """Clean up a combined record group into a multi-line string of single records.

    :param combined_record_group: tuple - everything from both participants.
    :return: str - everything "cleaned", excess coordinates and information are removed.

    The return statement should be a multi-lined string with items separated by newlines.

    (see HINTS.md for an example).
    """

    result = ""
    for record in combined_record_group:
        # hack to pass erroneous test
        if "Scrim" in record[0]:
            result += f"(\"{record[0]}\", \'{record[2]}\', {record[3]}, \'{record[4]}\')\n"
        else:
            result += f"(\'{record[0]}\', \'{record[2]}\', {record[3]}, \'{record[4]}\')\n"
    return result
```

Notice the special logic for the record that starts with "Scrim". It uses double quotes around record[0], whereas all the others use single quotes.

BethanyG:
Hi @chuck-alt-delete 👋🏽 

Thanks for the feedback, and for filing this issue!  💙  Apologies for the delay in getting back to you.

> Return None or raise error instead of "not a match" in create_record function.

For concept exercises, we're not permitted to use any language features that have not been introduced in a prior exercise or in the exercise itself, so raising an error here or _explicitly_ returning `None` is off the table.  However, we could have a student only return something if it's a match, and **not** return anything if there is no match  (_implicit_ return of `None`).  Not sure that's a better habit to teach.  

Python itself is all over the place.  `str.find()` returns `-1` when a substring is not found, and `str.index()` raises an error.   `dict[key]` raises an error, `dict.get()` returns `None`.  `list.index()` and `tuple.index()` raise errors.

[...]

PR #3071, opened 2022-06-02
Updating the test in the Tisbury Treasure Hunt

FranciscoRguezMelgar:
Since the first tuple contains a string with souble quotes it seems
the editor used to create the file made it a string with duouble quotes
to save the \' inside it. Putting this verbatim in the result string
make the test inconsistent with the instructions set.

### DIFF
--- a/exercises/concept/tisbury-treasure-hunt/tuples_test.py
+++ b/exercises/concept/tisbury-treasure-hunt/tuples_test.py
@@ -123,7 +123,7 @@ class TisburyTreasureTest(unittest.TestCase):
             ('Silver Seahorse', '4E', 'Hidden Spring (Island of Mystery)', ('4', 'E'), 'Yellow')
         )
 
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @FranciscoRguezMelgar 👋🏽 

Thank you for taking the time to file a PR. However, per the discussion in #3068, I will be closing this.  This is not about the test being inconsistent.  It is about students not fully understanding how quotes and escaping work in Python. 

Additionally, your expected result change looks incorrect to me.  `'Scrimshaw Whale's Tooth'` needs to have the `'` escaped, so the line should read `'Scrimshaw Whale\'s Tooth'`.  Both that and `\"Scrimshaw Whale's Tooth\"` are problematic, hence the change discussed in #3068.

bobahop:
Thank you for caring to improve Exercism!

In the future, please review the Contributing Guidelines before submitting a Pull Request. PRs should be based on changes proposed in an Issue. In this case,  Issue #3068 discusses that there are currently a couple of options being considered for how to deal with this situation. Because that decision has not been made, this PR is premature and will now be closed. Thank you, and please consider helping Exercism to improve in the future!

Issue #3069, opened 2022-05-28
[Diffie Hellman]:  Test File Needs Reformatting so that the Test Runner Does Not Cut Off Test Code in the UI


joshgoebel:
Test 3
DiffieHellman > can calculate public key using private key

Under Code run all I see is:

```
p = 23
g = 5
private_key = 6
self.assertEqual(
    8,
```

BethanyG:
Fixed in #3073 

![2022-06-03_05-50-17](https://user-images.githubusercontent.com/5923094/171857555-eaa464c7-a4ee-4f21-89e2-d31318a10636.png)

joshgoebel:
Awesome. Thanks so much. 😎

PR #3073, opened 2022-06-02
[Diffie- Hellman]: Corrected Malformed JinJa2 Template

BethanyG:
Addresses #3069 

- [x] Added a `# fmt: skip`instruction to Black
- [x] Reformatted JinJa2 template to have the `assert` test all on one line
- [x] Re-generated test file and tested example solution against it


bobahop (approved):
The code changes look good to me, although I don't know if they will have the desired effect, due to my limited knowledge in this area.

BethanyG:
I ran the new test file in the dockerized test runner, and got back `JSON` that included the full test code (it didn't before) - so I am holding out high hopes.  That being said, we won't really know until we deploy.  😄   So -- hopefully this works.

PR #3075, opened 2022-06-04
[Functions]: Config and Info Updates

BethanyG:
- [x]  Updated Functions blurb in `config.json`
- [x]  Added bobahop to authors array in `config.json`
- [x]  Added links on function definition and namespacing to `links.json`


bobahop (approved):
Changes look good.

PR #3080, opened 2022-06-08
[Test Generator]: Added  jinja2 `systemfileloader` Workaround for Windows Paths.

BethanyG:
See https://github.com/pallets/jinja/issues/767 for the reason why this workaround was necessary.

JinJa2 was failing to "find" the test generation templates due to not understanding Windows file paths.
While this was not _technically_ a bug in our test generator, it did prevent the script from successfully running on Windows.

Tests now successfully run on Windows 10 pro, was unable to test on Windows 11.

### DIFF
--- a/bin/generate_tests.py
+++ b/bin/generate_tests.py
@@ -28,7 +28,7 @@ import filecmp
 import importlib.util
 import json
 import logging
-from pathlib import Path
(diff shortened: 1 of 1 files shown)


bobahop (approved):
Looks good to me.

PR #3117, opened 2022-06-15
[String Formatting Concept]: Cleaned up Docs and Added Introduction

BethanyG:
- [x] Re-read for language, and added additional code examples.
- [x] Added introduction.md doc
- [x] Revised Blurb & removed double quote escapes.
- [x] Re-ordered links in `links.json`


bobahop (commented):
I think I saw only one error and one thing that might be nice to add but wouldn't be earthshaking if not.

Very good work done by both @valentin-p and @BethanyG .

bobahop (approved):

PR #3126, opened 2022-06-24
Correct config.json link to python

BethanyG:
Changed link to config.json to point at the python track config instead of the JavaScript track  config.


bobahop (approved):
Wow!

PR #3137, opened 2022-07-10
Update about.md

jhn--:
Corrected the result for `cat_words.split(',')` at line 143.


BethanyG (approved):
@jhn--  Thanks for the catch and the correction!  🌟

PR #3140, opened 2022-07-14
Update sets.py

tward333:
Update helper text to maintain consistency of constant names between docs and sets_categories_data.py


BethanyG (approved):
@tward333 - Nice Catch!  Many thanks for submitting this.  🌟 🎉

PR #3144, opened 2022-07-22
[Triangle]: Synced New Test Cases & Regenerated Test File.

BethanyG:
- [x] Synced `Tests.toml`
- [x] Regenerated test file
- [x] Tested that the `example.py` solution still passes.


bobahop (approved):
I was at first thrown in `tests.toml` by two sides being equal relating to scalene, but then saw in `triangle_test.py` that scalene should be false, so I think this looks good.

Issue #3078, opened 2022-06-07
Request to Remaster Pascal's Triangle


PaulT89:
### The TL;DR Version 🥱
I've rewritten Pascal's Triangle as an exercise to practice recursion, but the exercise is currently listed as 'deprecated'.  I am of the belief that the conditions under which it became deprecated 8 years ago no longer apply.  I am requesting permission to open a PR to re-instate Pascal's Triangle as an exercise.

### The Nuanced Version 🎓 
As has been brought to my attention, Pascal's Triangle was deprecated with PR #104 in response to Issue #99, the given reason being:

> I'd also like to focus the nitpicking community by deprecating some exercises whose core problem appears in other exercises.

My impression from this is that, back in 2014, the pool of contributors to the Python track was limited to such an extent that a decision had to be made about which exercises the maintainers wanted the community to ignore.  However, I read this enforced ignorance as a 'for the time being' kind of thing.  To my mind, the fact that these exercises were deprecated in the config file (rather than foregone) and still exist in the repository, implies an intention to, at some point in the future, bring them up to date and reinstate them.  My belief that deprecation is a temporary state of being for an exercise is reinforced by:

1. The reinstatement of the Raindrops exercise in PR [#1633](https://github.com/exercism/python/pull/1633) (which, on the whole, was a copying of the Ruby track core into the Python track, but explicitly included the de-deprecation of Raindrops).
2. The improvement over the years of the Pascal's Triangle exercise despite its deprecation (see issue [#1127](https://github.com/exercism/python/issues/1127), PR [#1164](https://github.com/exercism/python/pull/1164), issue [#606](https://github.com/exercism/python/issues/606), PR [#705](https://github.com/exercism/python/pull/705), PR [#2792](https://github.com/exercism/python/pull/2792), etc.).

As for the focusing of nitpicking efforts - well, from what I can tell, the Python community is larger today than it was 8 years ago, and while I don't know how active (proportionally speaking) it is now as compared to then, it stands to reason that, on average, a larger community is likely to result in more active contributors overall.  If this is so (which seems more likely than not), then I don't think the initial justification still stands.

[...]

PR #3150, opened 2022-07-24
[Pascals Triangle] Remastered for recursion

PaulT89:
As per Issue #3078.

## Changes
Rewrote example.py:
- Uses recursion rather than loops.
- Raises a meaningful error when row_count < 0, instead of just returning None.

Added template.j2:
- Based general layout of the template on the old pascals_triangle_test file.

Added instructions.append.md:
- Encourage student to use recursion to solve this problem
- Added boilerplate message about how to raise exceptions

.meta/config.json:
- Added name as co-author (though maybe contributor would be more appropriate?)

config.json:
- practices recursion
- added (seemingly) sensible sounding prerequisites
- kept difficulty
- moved entire block up to be in line with other exercises of similar difficulty.


BethanyG:
Hi @PaulT89 👋🏽 

Thank you so much for submitting this!  Jus to warn you:  I will probably be taking this in stages, since my time today/next week is a bit limited.  I'll move as quickly as I can, but you may get several "waves" or clusters of comments.  At a quick first glance:
1.  This looks to be missing the `./meta/tests.toml` file (see [this one](https://github.com/exercism/python/blob/main/exercises/practice/diffie-hellman/.meta/tests.toml) from `diffie-hellman` as an example).  The `tests.toml` file is needed by the generator so that it can figure out which test cases to generate/not generate for the exercise.  I think the file can be created by using [configlet](https://github.com/exercism/configlet), but you may need to dig in the practice exercise docs for details (_apologies - I don't have time at the moment to go dig for a link_).
2. Instead of putting additional tests in the JinJa2 template, we currently define them in a `./meta/additional_tests.json` file.  See [this one](https://github.com/exercism/python/blob/main/exercises/practice/clock/.meta/additional_tests.json) for the `clock` exercise as an example.  Seems tedious for just one test case, but that's the way we track them.
3. If the error handling case is from cannonical data, then you **don't** have to make an `additional_tests.json` file -- but you should also change the comment in the template, so that it doesn't confuse us in the future with implying that there are "additional" test cases for this exercise.
4. Making the `recursion` concept a _**prerequisite**_ for this exercise means that a student in learning mode won't be able to unlock this exercise without completing the `recursion` concept exercise.  But we don't currently _**have**_ a `recursion` concept exercise...so this exercise would never be unlocked. 😱  Suggest we remove the prerequisite until we can get a `recursion` concept exercise completed.  (_are you up for a challenge?  Wanna take on a `recursion` concept exercise?  Just let me know...._ 😄 )
5. **You totally should be an author here!**  The JinJa2 template alone is worth an author credit, but you also thought through error handling and other changes.
6.  We may want to elaborate a little bit on recursion in the instruction append.  I don't have time at the moment, but there is a way to cross-link concept docs inside instructions.  I think we should do that here.  We may also want to do an additional callout/warning for Python's recursion limit, so that students coming from a language that doesn't  limit recursion aren't caught by surprise.  Ditto for the lack of tail-call optimization.
7. We might want to consider a `hints.md` file here to give students some extra links around thinking through recursion.  It might also be useful to have some hints on how to do this via `for` or `wile` loop ... just to cover our bases.  But we can discuss that later on.

Apologies -- that's all I have for now, but will pick this up later today my time.  Thanks again for all your hard work on this! 🎉 🌈

PaulT89:
1. The `.meta/tests.toml` file already exists, but remains unmodified (which is why it doesn't show up in this PR).
2. Okay, I'll go back and fix that.
3. There's no test for negative rows in the canonical data.  The old `pascals_triangle_test.py` file tested for negative input, but it checked that `None` was returned, and I thought that it would be nice to keep the test, but raise a meaningful exception message instead.
4. Hmm... 🤔.  You're right, that does seem somewhat counterproductive - I should probably fix that.  **FIXED** As for creating the concept exercise, maybe later (like in a few weeks?) if nobody else has taken it (and I haven't got anything else on).  In theory at least, Pascal's Triangle could become the concept exercise.
5. YAY! 🎉
6. Yeah, I kept the instructions to a minimum because I imagined the concept exercise as already existing.  I forgot that we don't live in the Platonic Ideal Universe inside my head, so that's an oops on my part.  I'll go back and add a bit more to the instructions.
7. Re: `hints.md` - yeah, maybe.  Re: completing with `for` and `while` loops - I doubt that'll be an issue, though I suppose it would depend on how early/late the `Recursion` concept is put in the track.

Just one more thing.  You mentioned Python's recursion limit in point 6 (something that I completely forgot about), but it gave me an idea.  I've been trying to figure out a way to ensure that the student completes the exercise using recursion Instead of loops.  So far, the best that I could come up with was to simply rely on the student to complete this exercise in good faith (i.e. "I totally did it with recursion, not loops - trust me 😜").  **But** what if there's an additional test that deliberately uses Python's recursion limit to test for this?  Maybe something like this:

````python
import sys

def test_solution_is_recursive(self):
    with self.assertRaises(RecursionError) as err:
        rows(sys.getrecursionlimit() + 10) # +10 just to be sure
    self.assertEqual(type(err.exception), RecursionError)
    self.assertEqual(err.exception.args[0], "maximum recursion depth exceeded in comparison")
````

**Comment by PaulT89** [comment]
Instructions update plus additional hints still pending.  Can't seem to figure out why Exercises check / cannonical_sync 3.6, 3.7, and 3.8 keep failing.  I might need to download Python 3.6 to figure that out.

**Comment by PaulT89** [comment]
Success!  Well that's a fun little gotcha - the `RecursionError` message seems to be slightly different across versions (though at least the first 32 chars are identical).

**Comment by PaulT89** [comment]
Okay, I've updated the instructions and added hints.  It still might be a little too bare-bones, but I'll leave that decision up to you.

**Comment by BethanyG** [comment]
Gah!  😱   Apologies that this has been sitting so long, and has now incurred the wrath of the "did-not-updated-test-cases" bot.  @PaulT89  -- I am pretty swamped this morning, but I am hoping to get to this by the end of my day today (PDT), and get it un-stuck and on its way.  Thank you so so much for your patience.

**Comment by PaulT89** [comment]
@BethanyG Yeah, well, it's been a while since I've done anything on Github, so I decided to update all my repos and rebase various branches - and this happened.  I think the CI is objecting to differences in the prob-specs vs the actual tests, so I regenerated and added a PR with the changes.  Once that goes through, I'm assuming another rebase will fix this.

**Comment by BethanyG** [comment]
@PaulT89 - _indeed it did!_  Thanks again for doing that.  I went ahead and rebased your branch, so now the test are passing.  🎉   I won't have time until the weekend to review the PR.  But I will definitely do it then.  This has sat an obscenely long time, and is well past a push to production.  😉   My apologies for that.  Life and a new job completely swamped me.

**Comment by BethanyG** [comment]
@PaulT89  - I cannot apologize enough for letting this sit.  I've made a few edits to shorten the JinJa template and put in reflinks & widget links into the instruction append.  Everything's been re-tested and looks good.

THANK YOU for this exercise!  It is approved, and I am merging it in.  😄   I've also given you a little rep bump for your patience and efforts -- this one took way too long.

File: config.json [hunk 1]
```diff
@@ -1247,6 +1247,21 @@
...
+      {
+        "slug": "pascals-triangle",
+        "name": "Pascals Triangle",
+        "uuid": "e1e1c7d7-c1d9-4027-b90d-fad573182419",
+        "practices": ["recursion"],
+        "prerequisites": [
+          "basics",
+          "conditionals",
+          "functions",
+          "function-arguments",
```
BethanyG, line 1259:
```suggestion
```

BethanyG (approved):
Excellent work, @PaulT89!  Thank you again for working on this.  🌟

PaulT89:
> @PaulT89 - I cannot apologize enough for letting this sit. I've made a few edits to shorten the JinJa template and put in reflinks & widget links into the instruction append. Everything's been re-tested and looks good.
> 
> THANK YOU for this exercise! It is approved, and I am merging it in. 😄 I've also given you a little rep bump for your patience and efforts -- this one took way too long.

That's okay.  I deliberately didn't push the issue - it was funnier that way.  The longer it took for you to realize that this wasn't merged, the funnier it became.  Consider me suitably amused. 🤣

Issue #3159, opened 2022-08-13
[Improve Concept Docs] Rules for Using Underscores in PEP 515


KaiAragaki:
The [Numbers](https://exercism.org/tracks/python/concepts/numbers) concept tutorial states:

> 
> Rules for using underscores as outlined in [pep 515](https://www.python.org/dev/peps/pep-0515/) are as follows:
> * Only one consecutive underscore allowed, and only between digits.
> * Multiple consecutive underscores allowed, but only between digits.
> * Multiple consecutive underscores allowed, in most positions except for the start of the literal, or special positions like after a decimal point.

However, [PEP 515](https://peps.python.org/pep-0515/#underscore-placement-rules), from which the concept tutorial references, is in my interpretation stating _general trends among other languages_, rather than syntax allowed in Python. Python (at least 3.8.5) appears to follow the first bullets rules - only single consecutive underscores that appear strictly between two numbers - not between a number and a decimal, or at either end of a number.

BethanyG:
Hi @KaiAragaki 👋🏽 

Thanks so much for filing this issue, and catching that problem!  Would you be willing to submit a PR to correct it?  I'd be happy to help you walk through the process if needed.  😄 

Since exercises are comprised of multiple files, there are several places we'd want to double check the reference/wording:

1.  The _concept_ (numbers) [introduction.md](https://github.com/exercism/python/blob/main/concepts/numbers/introduction.md) markdown file.
2. The _concept_ (numbers) [about.md](https://github.com/exercism/python/blob/main/concepts/numbers/about.md) markdown file.
3.  The _exercise_ (currency exchange) [introduction.md](https://github.com/exercism/python/blob/main/exercises/concept/currency-exchange/.docs/introduction.md) markdown file.
4. (Potentially) the _exercise_ (currency exchange) [instructions](https://github.com/exercism/python/blob/main/exercises/concept/currency-exchange/.docs/instructions.md).

..and you would also want to add your GitHub user name to the list of contributors in both the [concept configuration](https://github.com/exercism/python/blob/main/concepts/numbers/.meta/config.json) and the [exercise configuration](https://github.com/exercism/python/blob/main/exercises/concept/currency-exchange/.meta/config.json) 😄 

Just let me know.  ⭐

KaiAragaki:
Thank you @BethanyG! PR #3160 should close this issue. Let me know if anything is amiss!

BethanyG:
Closed via #3160

PR #3160, opened 2022-08-14
Correct PEP 515 underscore rules (Ref. issue #3159)

KaiAragaki:
Corrected underscore rules in concept introduction.

Of the four files checked (concept introduction and about, exercise introduction and instructions) only one file needed to be fixed. Since only the concept was changed, contribution was only added to the concept.

I searched through the repo to determine if it came up anywhere else - later exercises, for instance - but that appears not to be the case.


BethanyG (approved):
@KaiAragaki  - Great work!  Thanks so much for submitting this PR.  🌟 🌈

PR #3165, opened 2022-08-25
[Currency Exchange] Improve the `int`/`float` usage consistency

### LINKED ISSUE #3157: [Currency Exchange] Inconsistent change value type (`float`/`int`)
Exchanged value is of type `float`:

https://github.com/exercism/python/blob/48e47c1f6af8050e135747461085d713a3231164/exercises/concept/currency-exchange/exchange.py#L6

Remaining value is of type `float` as well:

https://github.com/exercism/python/blob/48e47c1f6af8050e135747461085d713a3231164/exercises/concept/currency-exchange/exchange.py#L17

Value exchangeable in bills of a certain denomination is of type `int`, but that is justifiable because denominations are integers:

https://github.com/exercism/python/blob/48e47c1f6af8050e135747461085d713a3231164/exercises/concept/currency-exchange/exchange.py#L52

However, non-exchangeable value is of type `int` as well, which I believe is inconsistent:

https://github.com/exercism/python/blob/48e47c1f6af8050e135747461085d713a3231164/exercises/concept/currency-exchange/exchange.py#L65
 
Given the established structure, I see no reason why the non-exchangeable value should be integer. I mean, it clearly is not integer, but has to be rounded artificially to pass the tests.

---

Overall, I was surprised how convoluted this seemingly simple exercise was.

For one, there are few opportunities to reuse code in `exchangeable_value` (task 5) and `non_exchangeable_value` (task 6).
The essence of both of these functions is to calculate the exchangeable amount given the spread, which could've been implemented in `exchange_money` (task 1).
Then `exchangeable_value` could potentially reuse `get_value_of_bills` (task 3), whereas `non_exchangeable_value` could potentially reuse `get_change` (task 2) -- except for the fact that it expects an integer, therefore this issue.

As a side note, if we're really looking to encourage the usage of the modulo operator `%`, we could introduce a task to convert amounts using different denominations (and e.g. calculate the remainder that cannot be converted).

Happy to discuss & potentially submit a PR.

pranasziaukas:
Fixes #3157 and is somewhat discussed there.

Basically, what I ended up doing is:
- Added a new task asking to implement `get_leftover_of_bills`, which encourages using modulo operator `%`.
- Removed task 6, which was a weird semi-duplicate of task 5 (and also potentially hinted at using `%`).

Also, along the way:
- Tried to make `int` and `float` usage more intuitive.
- Used `assertAlmostEqual` method to compare floating point values in tests.
- Modified some test values to improve the variety of cases being tested.

### DIFF
--- a/exercises/concept/currency-exchange/.docs/instructions.md
+++ b/exercises/concept/currency-exchange/.docs/instructions.md
@@ -62,7 +62,19 @@ Effectively, you are rounding _down_ to the nearest whole bill/denomination.
 25
 ```
 
-## 5. Calculate value after exchange
+## 5. Calculate leftover after exchanging into bills
+
+Create the `get_leftover_of_bills()` function, taking `budget` and `denomination`.
+
+This function should return the _leftover amount_ that cannot be exchanged from your _budget_ given the denomination of bills.
+It is very important to know how much exactly the booth gets to keep.
+
+```python
+>>> get_leftover_of_bills(127.5, 20)
+7.5
+```
+
+## 6. Calculate value after exchange
 
 Create the `exchangeable_value()` function, taking `budget`, `exchange_rate`, `spread`, and `denomination`.
 
@@ -80,22 +92,4 @@ Remember that the currency *denomination* is a whole number, and cannot be sub-d
 80
 >>> exchangeable_value(127.25, 1.20, 10, 5)
 95
-```
-
-## 6. Calculate non-exchangeable value
-
-Create the `non_exchangeable_value()` function, taking `budget`, `exchange_rate`, `spread`, and `denomination`.
-
-This function should return the value that is *not* exchangeable due to the *denomination* of the bills.
-Remember - this booth gets to keep the change _in addition_ to charging an exchange fee.
-Start by calculating the value you would receive if you were able to keep subdivided bills, then subtract the amount you would receive in whole bills.
-Both amounts should take the spread, or the exchange fee into account.
-
-**Note:** Returned value should be `int` type.
-
-```python
->>> non_exchangeable_value(127.25, 1.20, 10, 20)
-16
->>> non_exchangeable_value(127.25, 1.20, 10, 5)
-1
 ```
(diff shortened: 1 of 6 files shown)


exercises/concept/currency-exchange/.meta/exemplar.py
```diff
@@ -39,36 +39,32 @@ def get_number_of_bills(budget, denomination):
     :return: int - number of bills after exchanging all your money.
     """
 
-    return int(budget / denomination)
+    return int(budget) // denomination
```
pranasziaukas, line 42: `//` is fully covered in the introduction.

exercises/concept/currency-exchange/.meta/exemplar.py
```diff
@@ -13,7 +13,7 @@ def get_change(budget, exchanging_value):
     """
 
     :param budget: float - amount of money you own.
-    :param exchanging_value: int - amount of your money you want to exchange now.
+    :param exchanging_value: float - amount of your money you want to exchange now.
```
pranasziaukas, line 16: No reason to restrict `exchanging_value` to `int` AFAICT.

exercises/concept/currency-exchange/.docs/instructions.md
```diff
@@ -62,7 +62,19 @@ Effectively, you are rounding _down_ to the nearest whole bill/denomination.
 25
 ```
 
-## 5. Calculate value after exchange
+## 5. Calculate leftover after exchanging into bills
+
+Create the `get_leftover_of_bills()` function, taking `budget` and `denomination`.
```
pranasziaukas, line 67: Couldn't think of a better function name. Suggestions (if any) are welcome.

BethanyG:
@pranasziaukas   Thank you for this!  I don't have time this morning (my time) to review, but will get to it in the evening/tomm morning latest.  😄

BethanyG (approved):
@pranasziaukas  - Nice work. Thanks for submitting this!  I've left comments, but they are optional suggestions and not blockers, so accept/change at your leisure. 🌟

exercises/concept/currency-exchange/.docs/instructions.md
pranasziaukas: I really don't have a strong preference so I'll just leave it as it is :)

PR #3168, opened 2022-09-01
Replace `assertIs` assertion with `assertEqual` for string literals in tests

aleksandr-shtaub:
Fixes wrong assertion method in "perfect numbers" tests.

Refs.: 
- https://github.com/python/cpython/issues/79031
- https://flake8.pycqa.org/en/latest/user/error-codes.html (F632)

### DIFF
--- a/exercises/practice/perfect-numbers/.meta/template.j2
+++ b/exercises/practice/perfect-numbers/.meta/template.j2
@@ -8,7 +8,7 @@
         self.assertEqual(type(err.exception), ValueError)
         self.assertEqual(err.exception.args[0], "{{ case["expected"]["error"] }}")
     {% else %}
-        self.assertIs(
+        self.assertEqual(
             {{ case["property"] | to_snake }}({{input["number"]}}),
             "{{ case["expected"] }}"
         )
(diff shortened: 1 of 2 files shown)


BethanyG:
Hi @aleksandr-shtaub 👋🏽 

Thanks for taking the time to submit a PR!  Unfortunately, these changes have been made in the wrong place.  

We use a test generator to automatically produce practice exercise test files.  They're not intended to be edited directly. 
To make the change, you will need to edit the [jinja2 template](https://github.com/exercism/python/blob/main/exercises/practice/perfect-numbers/.meta/template.j2) for this exercise and then regenerate the test file by running the [generator](https://github.com/exercism/python/blob/main/docs/GENERATOR.md).  You'll need to have both a clone of the Python repo and the [problem-specifications](https://github.com/exercism/problem-specifications) repo.  From `python/` on your cloned Python repo:

```bash
./bin/generate_tests.py perfect-numbers
```

Both the `jinja2` template and the re-generated file should then be checked in to the PR.

aleksandr-shtaub:
Hi @BethanyG,

Interesting, thank you for the info!
I did a fixup commit.

BethanyG (approved):
Nice work. Many thanks!  🌟

PR #3174, opened 2022-09-27
Fix bad assertion in DiffieHellmanTest

gma:
The original code passed a generator expression into the assertion, which was converted into a list prior to being evaluated for truthiness.

Even if all the values in the generator are False, the assertion will check whether `[False, False, False, ...]` is truthy, which it is.

I can't see how this test can ever have failed.

The most obvious solution is to have one assertion for each case that we want to test.

We could fix this by using `all()` around the generator expression, but I don't feel that's good style in tests. We want feedback about what failed, so I've used `subTest()` which is is used in other exercises.

### DIFF
--- a/exercises/practice/diffie-hellman/.meta/template.j2
+++ b/exercises/practice/diffie-hellman/.meta/template.j2
@@ -13,10 +13,7 @@ class {{ exercise | camel_case }}Test(unittest.TestCase):
         for prime in [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]:
             with self.subTest(f"prime={prime}"):
                 key = private_key(prime)
-                self.assertTrue(
-                    1 < key < prime,
-                    msg=f"{key} out of range, expected to be >1 and <{prime}"
-                )
+                self.assertTrue(1 < key < prime, msg=f"{key} out of range, expected to be >1 and <{prime}") # fmt: skip
     {%- elif property == "private_key_is_random" %}
         """
         Can fail due to randomness, but most likely will not,
--- a/exercises/practice/diffie-hellman/.meta/template.j2
+++ b/exercises/practice/diffie-hellman/.meta/template.j2
@@ -10,8 +10,13 @@ class {{ exercise | camel_case }}Test(unittest.TestCase):
     {% set input = case['input'] %}
     def test_{{ description }}(self):
     {%- if property == "private_key_is_in_range" %}
-        primes = [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
-        self.assertTrue(1 < private_key(p) < p for p in primes)
+        for prime in [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]:
+            with self.subTest(f"prime={prime}"):
+                key = private_key(prime)
+                self.assertTrue(
+                    1 < key < prime,
+                    msg=f"{key} out of range, expected to be >1 and <{prime}"
+                )
     {%- elif property == "private_key_is_random" %}
         """
         Can fail due to randomness, but most likely will not,
--- a/exercises/practice/diffie-hellman/diffie_hellman_test.py
+++ b/exercises/practice/diffie-hellman/diffie_hellman_test.py
@@ -14,10 +14,7 @@ class DiffieHellmanTest(unittest.TestCase):
         for prime in [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]:
             with self.subTest(f"prime={prime}"):
                 key = private_key(prime)
-                self.assertTrue(
-                    1 < key < prime,
-                    msg=f"{key} out of range, expected to be >1 and <{prime}",
-                )
+                self.assertTrue(1 < key < prime, msg=f"{key} out of range, expected to be >1 and <{prime}")  # fmt: skip
 
     def test_private_key_is_random(self):
         """
--- a/exercises/practice/diffie-hellman/diffie_hellman_test.py
+++ b/exercises/practice/diffie-hellman/diffie_hellman_test.py
@@ -11,8 +11,13 @@ from diffie_hellman import (
 
 class DiffieHellmanTest(unittest.TestCase):
     def test_private_key_is_greater_than_1_and_less_than_p(self):
-        primes = [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]
-        self.assertTrue(1 < private_key(p) < p for p in primes)
+        for prime in [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]:
+            with self.subTest(f"prime={prime}"):
+                key = private_key(prime)
+                self.assertTrue(
+                    1 < key < prime,
+                    msg=f"{key} out of range, expected to be >1 and <{prime}",
+                )
 
     def test_private_key_is_random(self):
         """


gma:
I'm pretty new to exercism, so not sure if this is worth mentioning or not, but I've seen a few community solutions to this problem that are incorrect.

These exercises would fail if tests were re-run after this assertion gets fixed.

BethanyG:
Hi @gma 👋🏽 

Thanks for taking the time to submit a PR!  Unfortunately (_as good as they are_), these changes have been made in the wrong place.  

`Diffie-Hellman` is what we call a "practice exercise".  Unlike the concept exercises (_those are the ones listed under the syllabus tab_), we use a test generator to automatically produce practice exercise test files.  They're not intended to be edited directly. 
To make the change, you will need to edit the [jinja2 template](https://github.com/exercism/python/blob/main/exercises/practice/diffie-hellman/.meta/template.j2) for this exercise and then regenerate the test file by running the [generator](https://github.com/exercism/python/blob/main/docs/GENERATOR.md).  You'll need to have both a clone of the Python repo and the [problem-specifications](https://github.com/exercism/problem-specifications) repo.  From `python/` on your cloned Python repo:

```bash
./bin/generate_tests.py diffie-hellman
```

Both the `jinja2` template and the re-generated file should then be checked in to the PR.

I also have one suggestion.  It might be friendlier to have an error message print out with the failure, so that a student is reminded what the test checks for, and why it might fail.  Here is a suggestion (_although you can choose some other error message if you'd like_):

```python
 def test_private_key_is_greater_than_1_and_less_than_p(self):
        for prime in [5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47]:
            with self.subTest(f"prime={prime}"):
                error_msg = f'Your private key is either less than 1 or greater than {prime}.'
                self.assertTrue(1 < private_key(prime) < prime, msg=error_msg)
```

But again, this would need to be translated into the JinJa2 template, and then regenerated via the generator script.

Let me know if you have any questions or issues.  Thanks again for the PR!

gma:
Hi @BethanyG, thanks for explaining that.

I've force pushed an update.

Are you sure I need the problem-specifications repo checked out? I did check it out to have a look at it, but didn't put it somewhere where the `./bin/generate_tests.py` would be able to find it.

Generation seems to have gone fine though, and moving it to a sibling of the `python` checkout didn't impact the generated output. I'm imagining that I'm not changing anything here that needs definitions in the problem spec repo?

If what I've got here isn't right for some tooling-related reason I'm afraid we might as well just close this PR; I've already spent a long time digging through multiple .md files in multiple repos and it's all quite confusing to be honest. If closing this we should raise an issue to get somebody familiar with the toolchain to fix it though; the test as written is broken and needs fixing.

Cheers.

BethanyG:
Hi @gma  -

**TL;DR**:  **You're good for tooling and process for this change**, and thank you for sticking with the process and wading through the documentation!  In the future, if you want to raise an issue but not submit a PR, it is perfectly fine to [log a problem](https://github.com/exercism/python/issues/new) in the issues board. 

The reason the test generation script needs a copy of the [problem-specifications](https://github.com/exercism/problem-specifications) repo is that it uses exercise-specific files called [`canonical-data.json`](https://github.com/exercism/problem-specifications/blob/main/exercises/diffie-hellman/canonical-data.json) to pull cross-track centralized test data and expected results.  These specifications are not just for Python -- they apply to most of our 61 programming language tracks and their practice exercises.  Some tracks use the data manually to maintain their tests; Python uses a generator and a series of JinJa2 templates.

Had you found an issue with the _**test data**_ or needed to add a test, I would have sent you to the problem specifications repo to make that suggestion or change.  However, your bug had to do with the Python test case **_implementation_**, so instead you made changes to the generation template for the track. 

We tell people to clone the problem-specs repo to make things smoother - but I checked -- the script will also [clone the repo if it doesn't find it](https://github.com/exercism/python/blob/main/bin/generate_tests.py#L428).  The down side of that clone is that it is not obvious, so if it fails out due to connectivity or some other reason, it is less clear to the user what has gone wrong.

Having problem-specifications next to where the track repo is means that the script should be able to find it.  However, you can also park the repo in an arbitrary place and [pass a flag](https://github.com/exercism/python/blob/main/bin/generate_tests.py#L412) to the script so that it can use that location.  If neither of those methods work, the script will clone a copy of the repo.

So thanks for sticking it through - and I hope some of this was interesting/a good learning experience.  I am going to run your re-generated tests against some sample solutions and make sure that everything looks good, then I'll approve it and push it out to production!  🎉 

-------

**As for the solutions that will now fail:**  That's what happens when things change, and those solutions will now be flagged as "passing at the time of submission" but will not show in community solutions if someone checks the "Only show solutions that pass the latests tests".  

image

<br>

Additionally, the students solution will show a flag that the exercise has updated, and will have the option to update their solution - at which point it will show that the tests will re-run and fail (or pass).

image

gma:
Thanks @BethanyG. I appreciate what you're saying about just filing issues, but when I see a problem that should have a quick fix it's nice to be able to just dive in and share it. I'm also wondering if it'd be possible to lower the barrier a bit.

exercism's behaviour when tests change and solutions start failing is impressively smooth, thanks for sharing what happens. It's a great site.

On the subject of lowering the barrier for contributions, do you know why the templates and docs are stored in hidden folders? After you pointed me towards the jinja templates I spent 20 minutes looking for them. I either use Neovim with the Telescope fuzzy file finder or ripgrep for finding files. Neither of them could locate the templates, so I spent a while (probably 20 minutes) bouncing around trying to find them.

Do you know if people have considered renaming `.meta` to `meta` and `.docs` to `docs`? I appreciate there are over 60 repos that will all follow the same pattern, but I'm thinking that over the long haul, making those directories discoverable for people submitting changes could make quite a big difference. And the change could be scripted; I'm not suggesting it'd have to be made manually. A bunch of PRs would no doubt need updating too, but they could be scripted with `git rebase`.

If they'd been visible via `ls` I'd probably have got this PR right in the first place, even though I didn't discover the relevant `CONTRIBUTING` file (and I did actually have a quick look for contribution guidelines).

Just a thought, though I appreciate this might not be your area…

BethanyG (approved):
Thank you again for this PR!  I pushed a small change:  I put the assertion all on one line for the template and the test file.  One of the bugs/quirks of the code that extracts tests for display through the website causes truncation when the assertions get formatted as multi-line.  It's something we'd like to fix (along with 10,000 other things), but we haven't gotten to it yet.  But you didn't know that.  😄   Otherwise -- this is great.

BethanyG:
@gma  - 

To your other points here.  These are all great questions, and great pieces of feedback!  😄   Let me see if I can address (_some of them_):

> _do you know why the templates and docs are stored in hidden folders?_

**TL;DR:**  Hidden files/directories are flags for content that is not _primarily_ public (non-contributor) or student (end user of website) -facing, and can include files that control various configurations or use cases, files that need assembly, and files that need pre processing.
 
[Here (GH)](https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md) and [here (website)](https://exercism.org/docs/building/tracks/practice-exercises) is one detailed rundown for practice exercise arrangement.  There are additional ones for [concept exercises (GH)](https://github.com/exercism/docs/blob/main/building/tracks/concept-exercises.md) | [concept exercises (website)](https://exercism.org/docs/building/tracks/concept-exercises), as well as for other components.

I think what can be frustrating (_and we are working on it!!_) is that we have both documents that describe the _intent_ of the exercises to a student along side documents that describe the logic of creation/organizing to a _contributor_ -- but haven't really provided clear paths to get to each.  As a Python track maintainer, I apologize that our contributing docs don't do a better job of things.  They are currently under revision, but (obviously) need more work.  😄 

So.  Like many things in exercism, _it's a bit complicated on closer examination_.  This repo not only holds content that is assembled for consumption on the website (_`.docs` folders, `.meta` folders among other things_) - it also holds content that students can download via the [exercism CLI](https://exercism.org/docs/using/solving-exercises/working-locally) (_the exercise stubs, the test files, the configs for pytest and pylint_) - as well as functioning as the jumping off point to those who'd like to contribute writing or code.  Its a LOT -- and doesn't even include the repos that hold the test runner, analyzer, and representer for the track.  Multiply that by 61 and add in the 4 repos that cover the website .. and you have an organizational nightmare in the making.  😉 

And since there are two different (_three, if you count GitHub itself_) "paths" of access, we have a system of organizing and flagging to keep things straight for the humans and the bots.  But then welcoming new contributors and getting them oriented becomes ....not insignificant.  So there is a cutoff point at which someone can't contribute as causally as they (or we) might like.  Going forward, we're discussing a more formal and staged onboarding for contributors, since getting hit with all of this as you are trying to make what looks to be a small change is not helpful or healthy.  

> _Do you know if people have considered renaming `.meta` to meta and `.docs` to docs?_ 

_oof_.  For that question, you'd want to talk to our OSS engineer Erik or community manager Johnathan, or another member of the exercism staff.  I can say as a maintainer and community member we had a LOT of discussions on how to organize all of this as we built out V3.  Months of discussions.  This is the system we settled on as being the clearest and best for those involved in doing the bulk of the work.  That doesn't mean we nailed it -- but it does mean (along with the sheer amount of stuff we're organizing) we probably wouldn't casually change it.

> _If they'd been visible via ls I'd probably have got this PR right in the first place, even though I didn't discover the relevant CONTRIBUTING file (and I did actually have a quick look for contribution guidelines — I was reading contribution info on the main site, not in the repo)_

I agree that having a swarm of things hidden or in parallel is not ideal, and we've had a series of community discussions on improving the "paths" of entry on the website -- including giving better visibility into contributor and track-specific documents.  Some redesigns of various sections of the website are in active progress.  And to that end -- if you are interested in becoming a longer-term contributor, or just having more insight into community discussions and progress, I'd encourage you to join in on our community calls.  They're typically on Fridays at 16:30 GMT, and are announced in our Slack channel for mentors and maintainers.  We're taking a break for the next couple of weeks, but they should start up again mid-October.

Alright -- I sincerely hope I haven't chased you away with this big wall of text.  If you are interested in getting involved further -- give a shout!  We'd love help here -- and in a bunch of other places.  I'd also encourage you to get involved as a mentor (and join our Slack channel), and participate in discussions on our (upcoming!!) student forum.  You can also pose questions/issues/points of discussion on our main `exercism/exercism` repo. 😄

BethanyG:
As an additional FYI, here is your change on the website:

image

🎉

gma:
Thanks @BethanyG, there's some fantastic info in there — I appreciate the effort. I've actually been quite interested in helping out since Katrina first announced the site, I just never got around to it.

Finding out how to join Slack was interesting. Google were no help, which suggested there isn't a page on the web site about it. After a few minutes of poking about and scratching my head (it can't be *that* secret!) I searched the exercism/website repo, and found [some Rails code][clue] that contained a big clue!

[clue]: https://github.com/exercism/website/blob/main/app/commands/user/become_mentor.rb

So I realised at that point that I needed to become a mentor. I will definitely try and join one of the regular calls; it'd help me get a feel for where I should spend my effort.

PR #3183, opened 2022-10-21
[Tree Building] swap parent_id and record_id in the exception message according to the task instruction

ypypy28:
From exercise's [instruction](https://github.com/exercism/problem-specifications/blob/main/exercises/tree-building/description.md) we know that 

> All records have a **parent ID lower than their own ID**, except for the root record, which has a **parent ID that's equal to its own ID**.

But the current exception message in the tests and the example code says the opposite

> Node record_id should be smaller than it's parent_id.

This could lead to misunderstanding while solving the exercise. PR fixes this by replacing parent_id with record_id and vice versa in the exception message

> Node parent_id should be smaller than it's record_id.


BethanyG (approved):
Hi @ypypy28 👋🏽   Good catch. Thanks for submitting the fix!

PR #3185, opened 2022-10-26
Fixing CLI test script for Guido's Gorgeous Lasagna

juditkisistok:
The `lasagna_test.py` script did not return any errors except for the import errors. This happened because the call to the unittest was missing from the end so I added it :)


BethanyG:
Hi @juditkisistok 👋🏽 

Thank you for submitting this PR.  Unfortunately, we can't accept it.  The code you added calls `unittest` when the test file is run from the command line.  However, all of our concept exercise tests (_and most of our practice exercise tests_) require that you use [pytest](https://docs.pytest.org/en/7.2.x/) for running test files.  See this [python track dock](https://exercism.org/docs/tracks/python/tests) for the details.

While `unittest` **_may work for now_** since the majority of our tests are written with `unittest`  syntax, they will break as soon as we start to use `pytest` syntax -- which we may do at any moment.  We removed the call to `unittest` from the majority of our test files so that students get used to using `pytest`, and don't run into issues when we start to re-write things.

PR #3189, opened 2022-10-28
Adjust return from hello function

peedrofranco:
IDK why the function was adjusted to "Goodbye, Mars!".
I only put the correctly return to "Hello World!"

### DIFF
--- a/exercises/practice/hello-world/hello_world.py
+++ b/exercises/practice/hello-world/hello_world.py
@@ -1,2 +1,2 @@
 def hello():
-    return 'Goodbye, Mars!'
+    return 'Hello, World!'


BethanyG:
@peedrofranco  -- Thank you for taking the time to make a PR, but unfortunately I will need to close it.

This is a stub file we provide for exercises, intended to be filled in/changed by the student when they solve the problem [online](https://exercism.org/tracks/python/exercises/hello-world) or via the [CLI](https://exercism.org/docs/using/solving-exercises/working-locally), which uploads to the website. 

It's not intended to be "solved" here in the repo.  🙂 

The goal of `Hello World` as a problem is to get a student oriented to the language and the system - so we provide a string that they need to _replace_ in the function, in the hopes that they get oriented to making _functions_ (not using `print()` statements) for their work.

PR #3200, opened 2022-11-15
Bob approaches

bobahop:


bobahop:
@ErikSchierboom Bob approaches for Python is ready for review

ErikSchierboom:
@BethanyG  We're approaching the approaches (pun intended) PRs with an optimistic merging strategy. As such, we're merging fairly quickly. If you have any comments, please leave them as review comments such that they can be addressed in a follow-up PR.

exercises/practice/bob/.approaches/answer-list/content.md
```diff
@@ -0,0 +1,67 @@
+# Answer list
+
+```python
+_ANSWERS = ['Whatever.', 'Sure.', 'Whoa, chill out!',
```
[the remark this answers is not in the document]
BethanyG: Additionally, a `tuple` might be preferable here for `ANSWERS`, making it immutable and therefore "more protected" from arbitrary change.

exercises/practice/bob/.approaches/answer-list/content.md
```diff
@@ -0,0 +1,67 @@
...
+
+```
+
+In this approach you define a [list][list] that contains Bob’s answers, and each condition is given a score.
+The correct answer is selected from the list by using the score as the list index.
+
+Python doesn't _enforce_ having real constant values,
+but the `_ANSWERS` list is defined with all uppercase letters, which is the naming convention for a Python [constant][const].
+It indicates that the value is not intended to be changed.
+Python also does not have real [private][private] variables,
```
BethanyG, line 24: we just shouldn't be using a "private" flag here.  Not unless it's internal to a class or function.

exercises/practice/bob/.articles/performance/content.md
```diff
@@ -0,0 +1,31 @@
+# Performance
+
+In this approach, we'll find out how to most efficiently determine the response for Bob in Python.
+
+The [approaches page][approaches] lists two idiomatic approaches to this exercise:
+
+1. [Using `if` statements][approach-if].
+2. [Using `if` statements nested][approach-if-nested].
+
+For our performance investigation, we'll also include a third approach that [uses an answer list][approach-answer-list].
```
BethanyG, line 10: Using a `tuple` here may or may not be faster than a list.

BethanyG (commented):
@bobahop - thanks for submitting this.  I've left some comments about variable names and other things I think need to be addressed.

@ErikSchierboom  - I agree that optimistic merging is a good approach.  However, things like camelCasing variable names or using underscores for a constant name aren't correct Python, so in future I would prefer those things get corrected before publishing.

bobahop:
@BethanyG Thanks for the great points! I've addressed them on a local branch which I'll double-check and be looking to merge tomorrow.

BethanyG:
@bobahop   Please make sure to address the constant/private variable stuff - it could really really confuse a student (_it actually confused me for a bit, and I had to double-check PEP8_  😱).  Personally, even in approaches, I think we should be saving `private` flags and `name mangling` for those exercises that are more OOP focused.

ErikSchierboom:
> @ErikSchierboom - I agree that optimistic merging is a good approach. However, things like camelCasing variable names or using underscores for a constant name aren't correct Python, so in future I would prefer those things get corrected before publishing.

Point taken. I may have been a bit too optimistic. 

Just for clarity, when you're saying "aren't correct Python," do you mean that it doesn't work or that it isn't idiomatic Python?

BethanyG:
> Just for clarity, when you're saying "aren't correct Python," do you mean that it doesn't work or that it isn't idiomatic Python?

#### TL;DR

I mean it's not idiomatic Python - which I know is a perspective both pedantic and conservative.  But that (_unfortunately_) is the whole point here since we're showing students **_exemplars_**.  

On one hand, all play is good and a running program is far better than an incomplete one that has perfect formatting.  On the other?  These silly **_conventions_** are used to judge if someone "knows" Python, and can really hurt someone who's doing a code challenge for an interview, submitting a PR for an OSS project or otherwise showcasing their code.

_feel free to skip the rest of this if you don't want to geek out on Python, or read through my reasons for being such a pendant in these cases._

The problem (style-wise) is that Python's compiler/interpreter is very very forgiving in a lot of ways.  It will ignore things like `()` around things or even `;` as a statement terminator - as long as there is no ambiguity introduced when parsing.  It mostly doesn't care what you call things, or how they're capitalized or hyphenated/underscored (_with a few exceptions_).  As long as you are consistent, it doesn't even care if you use `tab`, one space, or 15 for indentation.  The fact that the docs on the [lexical analyzer](https://docs.python.org/3/reference/lexical_analysis.html) talk about "logical lines" vs "physical lines" and "line joining rules" speaks volumes.

And the Python Style Guide even says [A Foolish Consistency is the Hobgoblin of Little Minds](https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds) - the overriding principal being that you should not slavishly follow rules  -- you should be consistent with surrounding code, team conventions, and anything that would break compatibility if it changed.

And I think that that is part of what makes Python both "easy" and completely infuriating to learn. It allows you to write totally functional code that is also not recognizable as _Python_ in many circles.  At which point you are free to get into endless discussions (_and holy wars_) around convention.  

By example, the Python interpreter doesn't care **at all** if you name something in a function `_ANSWERS`.  But others who follow the conventions around [constants](https://peps.python.org/pep-0008/#constants) and ["private" variables](https://peps.python.org/pep-0008/#designing-for-inheritance) would **_definitely_** care - and for those learning those conventions, it can be almost impossible to figure out what is is what.

Another example are all the ways `()` are overloaded in Python:

* `()` after a name signifies that something is [callable](https://docs.python.org/3/reference/expressions.html#calls) - even when it isn't supposed to be.  This can be exploited to do things like have dictionaries of function names that can be looked up and called, like `line 45` of this questionable solution to [robot simulator](https://exercism.org/tracks/python/exercises/robot-simulator/solutions/BethanyG).
*  `()` in a mathematical expression can be used to enforce order of precedence.
*  `()` can be used to group boolean AND and OR statements for clarity and/or precedence.
*  `()` around a comprehension forms a [generator expression](https://snarky.ca/not-unravelling-generator-expressions/).
*   `()` around a bunch of comma-demarked data signifies a [tuple](https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences) - even though it's really the _commas_ doing the heavy lifting here. 
*  `()` can be used around single expressions to "cheat" indentation rules  because  Python will consider everything between `(` and `)` a logical line. 
*  `()` can be used in `import` statements when multiple libraries are imported, so that you can either group a long line or put each import on its own line (example [here](https://github.com/exercism/python/blob/main/exercises/concept/cater-waiter/sets_test.py#L5)).  And in this one case, that doesn't mean that the import forms a `tuple`.
*  As of Python `3.10`, `()` can now be used to create [multi-line context handlers](https://docs.python.org/3/whatsnew/3.10.html#parenthesized-context-managers)
*  There are undoubtedly others I don't know about or are forgetting.

So given that laundry list, using `()` arbitrarily is ... loaded.  The interpreter will ignore them _usually_,  but as a learner, if you don't understand all the use cases above, you could be setting yourself up for problems in the future.  Others reading your code might not understand your intentions,  so it's probably best to never get in the out-of-context habit.  (And I often feel like an ass when I point this out.).

I could go on and on, but this is already waaaaaay too long, so I am going to step off my soapbox.  I am truly not trying to be difficult.  I just wanted to detail why I am the crazy person I am around  this. 😄

ErikSchierboom:
Wow, that was a bigger reply that I expected :D

BethanyG:
Apologies. I guess I've gotten a few too many _"what's the big deal?  It compiles."_   comments. :P

ErikSchierboom:
No that was very informative!

PR #3201, opened 2022-11-16
Bob approaches fixes

bobahop:
Fixes

- snake case names
- removes leading underscore and private info
- adds info on tuples

### DIFF
--- a/exercises/practice/bob/.approaches/answer-list/content.md
+++ b/exercises/practice/bob/.approaches/answer-list/content.md
@@ -1,7 +1,7 @@
 # Answer list
 
 ```python
-_ANSWERS = ['Whatever.', 'Sure.', 'Whoa, chill out!',
+ANSWERS = ['Whatever.', 'Sure.', 'Whoa, chill out!',
             "Calm down, I know what I'm doing!"]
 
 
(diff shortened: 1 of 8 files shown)


ErikSchierboom (approved):
LGTM @BethanyG could you check if the comments you made previously have been addressed?

BethanyG (approved):
I think there is a stray `isShout` ref in the final couple prose paragraphs, and  `isShout` | `isQuestion`  in the table header.  Beyond that, LGTM.

There are a couple of other approaches I can think of, but both would probably be over engineered (using an answers `dict` and structural pattern matching).  If I get inspired, I'll file a follow-on PR.  🙂

BethanyG:
P.S.  Leaving unmerged in case you want to fix those last three things.  But feel free to merge at will.

ErikSchierboom:
> There are a couple of other approaches I can think of, but both would probably be over engineered & slow (using an answers dict and structural pattern matching). If I get inspired, I'll file a follow-on PR. 

Sounds like they'd make for excellent approaches 😉

ErikSchierboom (approved):
LGTM! All occurances of `isShout` and `isQuestion` have been fixed

PR #3202, opened 2022-11-16
Sync exercise metadata with problem-specifications

kytrinyx:
This syncs all the metadata for all the practice exercises with
upstream changes in problem-specifications, in one fell swoop.

The changes are all nitpicky changes made in order to ensure that
source urls all use https rather than http, along with some
tweaks to better describe the source of certain exercises.

If you have a different preferred process for `configlet`-related changes,
feel free to close this. If so, and if it's not too much trouble, please point me
in the right direction, and I will happily follow it. Or I can stay out of your way and
just let you get on with things :-) 💙

### DIFF
--- a/exercises/practice/affine-cipher/.meta/config.json
+++ b/exercises/practice/affine-cipher/.meta/config.json
@@ -21,5 +21,5 @@
     ]
   },
   "source": "Wikipedia",
-  "source_url": "http://en.wikipedia.org/wiki/Affine_cipher"
+  "source_url": "https://en.wikipedia.org/wiki/Affine_cipher"
 }
--- a/exercises/practice/allergies/.meta/config.json
+++ b/exercises/practice/allergies/.meta/config.json
@@ -30,6 +30,6 @@
       ".meta/example.py"
     ]
   },
-  "source": "Jumpstart Lab Warm-up",
-  "source_url": "http://jumpstartlab.com"
+  "source": "Exercise by the JumpstartLab team for students at The Turing School of Software and Design.",
+  "source_url": "https://turing.edu"
 }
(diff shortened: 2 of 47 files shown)


BethanyG:
Hi @kytrinyx 👋🏽 

Thank you for this!

> If you have a different preferred process for configlet-related changes,
feel free to close this. If so, and if it's not too much trouble, please point me
in the right direction, and I will happily follow it. Or I can stay out of your way and
just let you get on with things :-) 💙

No  need to close the PR or get out of my way!  While the PR rep for these is nice, its more important they _actually get done_, so I appreciate you doing it. 😄 

-- My general process is to make myself sit down every so often (_that every so often is usually triggered by the activity I see in problem specs_) & update all the things using configlet.  I'd generally group change types together - so metadata, then instructions, and so forth.  😄   With the exception that if  I have to also update test cases and [re-generate test files](https://github.com/exercism/python/blob/main/bin/generate_tests.py) for a given exercise, I usually break all the change for that exercise out into its own PR.

I generally know when tests need updating, because it "breaks" the [CI](https://github.com/exercism/python/blob/main/.github/workflows/ci-workflow.yml),  which is rather rude for any PR that gets caught in that crossfire - but it's an excellent system for lighting a fire under my arse.  😆  That also usually motivates me to do the other updates, and also check the exercises we don't have test templates for.

I've been procrastinating on updates and reviews this week, so this is great -- I can cross the meta-data off the list and it will push me to go and update the other things now as well. 💙 

And now that I've looked at the CI workflow again, it needs updating to remove Python `3.6` dependancies.....

BethanyG (approved):

kytrinyx:
@BethanyG Thanks so much for sharing how you approach these!

I will go ahead and update the instructions (in one fell swoop as well), even though some of the exercises also have pending test changes. For the most part the improvement to the instructions are formatting changes, or improvements to the description that aren't "wrong", even if the tests don't get fully updated.

PR #3207, opened 2022-11-19
Update instructions.md

aguillenATC:
If the ability method is not implemented, the tests cannot be passed.

Another possibility is to provide that method already implemented, here it is a proposal:

def ability(self):
        
        chosen_ability = round(random.random()*6)       
        
        sw_dict ={
            0: self.strength,
            1: self.dexterity,
            2: self.constitution,
            3: self.intelligence,
            4: self.wisdom,
            5: self.charisma
        }
        return sw_dict.get(chosen_ability)

### DIFF
--- a/exercises/practice/dnd-character/.docs/instructions.md
+++ b/exercises/practice/dnd-character/.docs/instructions.md
@@ -24,6 +24,8 @@ For example, the six throws of four dice may look like:
 
 Because constitution is 3, the constitution modifier is -4 and the hitpoints are 6.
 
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @aguillenATC 👋🏽 

Thank you for this PR.  Unfortunately, I am going to have to close it.  The instructions and test cases for this exercise are pulled from a common upstream repo called [problem specifications](https://github.com/exercism/problem-specifications), so any changes would need to be suggested/PR'd there, then pulled into this repo.  Thanks for your understanding on this.  💙

PR #3211, opened 2022-11-20
Corrected parameter type stated in docstring

c1505:
I noticed the parameter type for `dish_ingredients` in the docstring is `set`  while the actual type is `list`.  The readme is correct and states the parameter as `list`.


BethanyG (approved):
Hi @c1505 👋🏽 

Thanks for filing this!  Good catch.  😄.

PR #3235, opened 2022-11-30
Update autoresponder for pausing community contributions

kytrinyx:
We realized belatedly that we should create a shared, re-usable workflow.

This simplifies the workflow logic using a variable for the forum category
that the new topic gets created in, if they click the link to copy their post
from here to the forum.


BethanyG (approved):
Nice!

PR #3257, opened 2022-12-03
[Maintenance]: Update sgf-parsing.

BethanyG:


exercises/practice/sgf-parsing/.meta/example.py
```diff
@@ -47,7 +47,7 @@ def parse(input_string):
     root = None
     current = None
     stack = list(input_string)
-
+
```
kytrinyx, line 50: Is this an accidental addition?

kytrinyx (approved):
This looks good to me. 🌻

exercises/practice/sgf-parsing/.meta/example.py
BethanyG: Might be there for "readability", but not strictly necessary.  I can back it out.  😄

PR #3258, opened 2022-12-03
[Maintenance]: Update poker

BethanyG:


kytrinyx (approved):
I heartily approve. With hearts. 💛 🧡 ❤️ 🧡 💛 

(Please feel free to ping me on PRs that update practice exercises)

BethanyG:
Awesome sauce!  _Many thanks for that!_ 💜 💛 💙 

@meatball133 and I are going through the practice exercises and updating test cases and test generation templates (where we can), so you might see some PRs get opened but then stay open for a bit pending a template addition, instruction append, or other edit.

These are some of the exercises missing templates:

* Bank account
* Dot DSL
* Hangman
* Ledger
* Robot Name
* Simple Linked List
* Tree Building
* Hangman
* React

We're also implementing as many additional practice exercises as we can, so you might see a flurry of those in the next week or so.

PR #3263, opened 2022-12-04
Reformat exercise config files

kytrinyx:
This runs the `bin/configlet fmt --update` command to
reformat the exercise configs.

Feel free to close this if it gets in the way of the work that you're doing; I'm simply doing a big sweep across all the tracks.


BethanyG (approved):
Apologies for the delay in reviewing.  Thanks so much for doing this!  ✨

PR #3264, opened 2022-12-04
Pig latin approach

bobahop:
Approach for Pig Latin. Another one where I felt comfortable explaining only one approach.


BethanyG (approved):
LGTM

PR #3274, opened 2022-12-07
[Fix]: Fix so the data shows correct in json

meatball133:


BethanyG (approved):
Many thanks for this!  🌟 ✨

PR #3303, opened 2022-12-27
[Plane_ticket & generator]: improvments

meatball133:
This pr is meant to improve the **unreleased** exercise "plane ticket" and its concept to achieve a better state.

As of this has the following changes been made:

- Changed the code format.
- Changed the link formating

What is left:

- Add an introduction for the exercise
- Add/improve hints
- Add a blurb for the concept
- Various improvements for the exercise and concept

### DIFF
--- a/concepts/generators/.meta/config.json
+++ b/concepts/generators/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "Learn about generators by assigning seats to passengers.",
+  "blurb": "Generator are functions that returns a lazy iterator, lazy iterator is an iterator like: list, tuple, etc. But doesn't need store its content in memory",
   "authors": ["J08K"],
   "contributors": []
 }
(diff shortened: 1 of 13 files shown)


meatball133:
This last commit added the following:

- Changed icon (discussable the choice I made)
- Added a new starting exercise, since I felt the step from start to complete the first exercise was quite big.
- Various fixes and changes.

concepts/generators/.meta/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "blurb": "Learn about generators by assigning seats to passengers.",
+  "blurb": "Generator are functions that returns a lazy iterator, lazy iterator is an iterator like: list, tuple, etc. But doesn't need store its content in memory",
```
BethanyG, line 2:
```suggestion
  "blurb": "Generators are functions that return lazy iterators. Unlike iterators such as lists or tuples,  generators do not have to store data members in memory and can yield them on demand.",
```

BethanyG (approved):
@meatball133 - Thank you for submitting this, and for your patience as I reviewed it.  Apologies for the long delay.  I am still on the fence about weather or not to have the `introduction.md` and the `about.md` match in these circumstances, but I think this is good enough that we should merge this and then take a pass again when we look at the exercise in detail, prior to making it live.

Nice work, you!  🌟 🌈

PR #3309, opened 2022-12-30
[Spell-fix]: leap-approach 

meatball133:


BethanyG (approved):
Nice catch!  🌮   Flagging as a small change, since it's only a couple of words.

PR #3315, opened 2023-01-07
Update black_jack_test.py

maxyodedara5:
Test case for condition for doubling down were incorrect, cannot double down when cards are 4,5. Should be False in test.

### DIFF
--- a/exercises/concept/black-jack/black_jack_test.py
+++ b/exercises/concept/black-jack/black_jack_test.py
@@ -88,7 +88,7 @@ class BlackJackTest(unittest.TestCase):
     @pytest.mark.task(taskno=6)
     def test_can_double_down(self):
         data = [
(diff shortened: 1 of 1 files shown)


IsaacG:
The instructions read,

> When the original two cards dealt total 9, 10, or 11 points, a player can place an additional bet equal to their original bet. This is known as "doubling down".

Cards `['4', '5']` has a sum value of 9. That is one of the double-down values, `9, 10, 11`.

In the future, if you suspect a bug with one of the tests, it might be worth first requesting a mentoring session and double checking with a mentor. Unless an exercise or test changed recently, it's probably been solved by others already who figured out a reading of the requirements that they were able to solve.

maxyodedara5:
Yes I realized as soon as I pushed the commit, but it was auto closed by bot so couldn't delete the PR. Thanks for the info though !

PR #3316, opened 2023-01-07
Clarify Currency Exchange instructions

MatthijsBlom:
Simply remove "new" and "old" specifiers, i.e. clarify through abstraction.


BethanyG:
@MatthijsBlom  **_clarify through abstraction_**  is going to be my new catch phrase.  😆

BethanyG (approved):
@MatthijsBlom  - thank you for submitting this.  I think this works (_of course, I may have extreme bias here, having dealt with this exercise a LOT_).  I don't recall any other places besides here in the instructions that refer to "old" and "new" currency.  Task 6 also has "new" in it, but I think that one is appropriate.

MatthijsBlom:
Yeah, I did not mean other cases with "old"/"new" specifically. Rather, I did not check for other consistency/coherency issues between the (exemplar) solution and the instructions.

PR #3319, opened 2023-01-08
Improve test for Ellen's Alien game

alatiera:
The test is using a zip iterator from the expected input
`position_data` and output, `obj_list`, returned by the
user's new_aliens_collection implementation.

However the zip is not Strict, meaning that if the user provided
a list with less objects than the expected output, the test would
still pass. For example: `return [Alien(x,y)]`.

Call the zip() function with the strict argument, so that it will
error out if the `obj_list` doesn't match the length of the
`position_data`.

### DIFF
--- a/exercises/concept/ellens-alien-game/classes_test.py
+++ b/exercises/concept/ellens-alien-game/classes_test.py
@@ -142,7 +142,7 @@ class ClassesTest(unittest.TestCase):
         obj_list = new_aliens_collection(position_data)
         obj_error = "new_aliens_collection must return a list of Alien objects."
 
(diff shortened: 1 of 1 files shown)


alatiera:
As the bot asked, the discourse thread is here: https://forum.exercism.org/t/bug-fix-for-ellens-alien-game-in-python/2441

I've also [force pushed](https://github.com/exercism/python/commit/bb9a0818f730ce275b44ff649690b4194c34c5e7) the branch and added a custom error message to make it simpler, but it doesn't seem to be reflected in the PR

PR #3335, opened 2023-02-05
Fix test

hexeberlin:
Average of [2, 3, 4, 8, 8] is 5, median is 4, so the test result should be `False`


IsaacG:
> Return True if **either one or both** of the, above named, strategies result in a number equal to the actual average.

Median is only half of the tests.

hexeberlin:
> > Return True if **either one or both** of the, above named, strategies result in a number equal to the actual average.
> 
> Median is only half of the tests.

oh damn, sorry and thanks :)

PR #3340, opened 2023-02-10
[Cater-Waiter] Fixes type issue raised in 3190

### LINKED ISSUE #3190: [Cater Waiter] test for separate_appetizers function accepts a set in return statement
I was doing the task no. 6 'Pull out Appetizers for Passing on Trays' and found out that my result passes tests even though it's a set and not a list.

This is because the test uses the `sorted()` function which returns an object with `list` type. I think this is a bug, because the requirements said I should return a `list`.   
My idea of fixing the test is to use the method `sort()` instead of `sorted()`.

`self.assertEqual((separate_appetizers(item[0], item[1]).sort()), (result.sort()), msg=error_message)`

In this case my solution does not pass and I get the following error:
 
`AttributeError: 'set' object has no attribute 'sort'`

Which may probably be a bit confusing for someone checking test output, but maybe there is a way to catch this error and throw a custom message.

BethanyG:
Fixes #3190 by adding a `self.assertIsInstance` check on line **106** under the `test_separate_appetizers` test.


ErikSchierboom (approved):

BethanyG:
Late addition of a correction to `config.json` that I hope fixes [this sync error ](https://github.com/exercism/python/issues/3361)

PR #3341, opened 2023-02-14
Adding missing double quotes

matiasp95:
One of the examples was missing double quotes. (Approach: Str translate)


BethanyG (approved):
Thanks for the catch!

Issue #111, opened 2014-09-25
Allergies test case issue


ghost:
https://github.com/exercism/xpython/blob/master/allergies/allergies_test.py#L26
... to ... 
https://github.com/exercism/xpython/blob/master/allergies/allergies_test.py#L36 

TC: test_allergic_to_just_peanuts -- self.assertEqual(['peanuts'], Allergies(2).list
TC: test_allergic_to_everything -- sum([2**x for x in range(0,8)]) == 255
TC: test_ignore_non_allergen_score_parts -- self.assertEqual(['eggs'], Allergies(257).list

257-255 == 2 == ['peanuts']
257-2**8 == 1 == ['eggs']

The last test case with 257 being passed-in, should approximate to the same as the Line 26 test case.

PR #3347, opened 2023-02-25
[Hello World]: Modified ImportError Message to be More Direct.

BethanyG:
In conjunction with [Python Test Runner #111](https://github.com/exercism/python-test-runner/pull/111), modified the template and regenerated the test file to include explicit error messages about needing a function with a `return`, and not using `print()` for "Hello, World!".

Error message will now read:

```python
ImportError:
        
       MISSING FUNCTION --> In your 'hello_world.py' file, we can not find or import the function named hello(). 
       The tests for this first exercise expect a function that returns the string "Hello, World!"
        
        Did you use print("Hello, World!") instead?

```

If the student instead uses the function but replaces `return` with `print()`, they should see the following:

```python

 def test_say_hi(self):
        msg = "\n\nThis test expects a return of the string 'Hello, World!' \nDid you use print('Hello, World!') by mistake?"
       self.assertEqual(hello(), "Hello, World!", msg=msg)
        AssertionError: None != 'Hello, World!' : 
       
        This test expects a return of the string 'Hello, World!' 
        Did you use print('Hello, World!') by mistake?

```

### DIFF
--- a/exercises/practice/hello-world/.meta/template.j2
+++ b/exercises/practice/hello-world/.meta/template.j2
@@ -1,10 +1,28 @@
 {%- import "generator_macros.j2" as macros with context -%}
-{{ macros.header() }}
+
+import unittest
+
+try:
+    from hello_world import (
+    hello,
+    )
+
+except ImportError as import_fail:
+    message = import_fail.args[0].split('(', maxsplit=1)
+    item_name = import_fail.args[0].split()[3]
+
+    item_name = item_name[:-1] + "()'"
+
+    # pylint: disable=raise-missing-from
+    raise ImportError("\n\nMISSING FUNCTION --> In your 'hello_world.py' file, we can not find or import the"
+    f' function named {item_name}. \nThe tests for this first exercise expect a function that'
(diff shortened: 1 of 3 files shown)


BethanyG:
@iHiD @kytrinyx No huge rush on this - but wanted both of you to look at it because it is a significant change to `Hello World`.

This is ....experimental.  The thought here being that if we make the UI error message more explicit, we might stave off some of the _"Help, I am stuck on Hello World"_ posts.  Of course, this won't work if a student is doing the exercise locally, but I am hoping that scenario is more of an edge case.

This would be in conjunction to a change in the test runner [here](https://github.com/exercism/python-test-runner/pull/111).

kytrinyx:
@BethanyG I'm pretty sure that if they're stuck on _Hello, World!_ then they're working in the browser, not the command-line!

In general anything that we can do to improve the error messages is a good thing, in my book.

(Note to self: once we have the learn-to-code product we should email everyone who started hello-world but didn't finish it and suggest that they check it out...)

iHiD:
Nice work :)

PR #3352, opened 2023-03-02
Correct rounding explanation

ceddlyburge:
Fractional parts of 0.5 and above round up, whereas the documentation had examples of 0.5 rounding both up and down!

### DIFF
--- a/exercises/concept/making-the-grade/.docs/instructions.md
+++ b/exercises/concept/making-the-grade/.docs/instructions.md
@@ -9,7 +9,8 @@ You decide to make things a little more interesting by putting together some fun
 While you can give "partial credit" on exam questions, overall exam scores have to be `int`s.
 So before you can do anything else with the class scores, you need to go through the grades and turn any `float` scores into `int`s. Lucky for you, Python has the built-in [`round()`][round] function you can use.
 
-A score of 75.45 or 75.49 will round to 75. A score of 40.50 will round to 40. A score of 43.50 (_or above_) will round to 44.
+The round function will round down fractional parts less than 0.5, and round up fractional parts greater than or equal to 0.5.
+For example, a score of 75.45 or 75.49 will round to 75, and score a of 75.50 or 75.64 will round to 76.
 There shouldn't be any scores that have more than two places after the decimal point.
 
 Create the function `round_scores()` that takes a `list` of `student_scores`.
--- a/exercises/concept/making-the-grade/.docs/instructions.md
+++ b/exercises/concept/making-the-grade/.docs/instructions.md
@@ -9,10 +9,6 @@ You decide to make things a little more interesting by putting together some fun
 While you can give "partial credit" on exam questions, overall exam scores have to be `int`s.
 So before you can do anything else with the class scores, you need to go through the grades and turn any `float` scores into `int`s. Lucky for you, Python has the built-in [`round()`][round] function you can use.
 
-The round function will round down fractional parts less than 0.5, and round up fractional parts greater than or equal to 0.5.
-For example, a score of 75.45 or 75.49 will round to 75, and score a of 75.50 or 75.64 will round to 76.
-There shouldn't be any scores that have more than two places after the decimal point.
-
 Create the function `round_scores()` that takes a `list` of `student_scores`.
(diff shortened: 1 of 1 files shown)


exercises/concept/making-the-grade/.docs/instructions.md
```diff
@@ -9,7 +9,8 @@ You decide to make things a little more interesting by putting together some fun
 While you can give "partial credit" on exam questions, overall exam scores have to be `int`s.
 So before you can do anything else with the class scores, you need to go through the grades and turn any `float` scores into `int`s. Lucky for you, Python has the built-in [`round()`][round] function you can use.
 
-A score of 75.45 or 75.49 will round to 75. A score of 40.50 will round to 40. A score of 43.50 (_or above_) will round to 44.
+The round function will round down fractional parts less than 0.5, and round up fractional parts greater than or equal to 0.5.
```
BethanyG, line 12: While I like this clarification, it doesn't hold true for all floating point numbers.  See the [note](https://docs.python.org/3/library/functions.html#round)  in the documentation for `round()`.

`40.50` really does round to `40`.  I think (_I don't remember exactly_) that's why we chose it for both an example and in the test data.  The same issue shows up for all floats ending in zero with a `.5` fractional part.  To make matters worse, `.50` rounds to `0`.

I'm now thinking that we might want to add a note box the way the docs do, to make sure folx really understand what is otherwise bizarre-seeming behavior?

ceddlyburge: Interesting, that is some crazy rounding. I don't think it really matters for the exercise, so maybe it is easier to leave it out, or just add a link to the note?

BethanyG: It is indeed counterintuitive.  😁  Its called "Bankers Rounding", or "Round to Even".
Believe it or not, it is the first of 5 choices in the [IEEE 574-2019](https://en.wikipedia.org/wiki/IEEE_754#Rounding_rules) standard for floating point arithmetic.
It was also the source of much debate when Python changed to it in 3.0.  The Python 2.7.x series used "away from zero".  Not that you want to, but here is a [Stack Overflow post on Python](https://stackoverflow.com/questions/10825926/python-3-x-rounding-behavior) and [another on Banker Rounding](https://stackoverflow.com/questions/45223778/is-bankers-rounding-really-more-numerically-stable) for all the crazy details.

______

**TL;DR**:  suggest we change:

> The round function will round down fractional parts less than 0.5, and round up fractional parts greater than or equal to 0.5.

to something like:

The round function uses what is called "Banker's Rounding".
Fractional parts less than 0.5 are rounded down, and fractional parts greater than 0.5 are rounded up.
Fractional parts that are _exactly_ 0.5 are rounded down for even numbers, and up for odd numbers.
For more information, see the documentation for  [`round()`](https://docs.python.org/3/library/functions.html#round).


And then omit the number examples.

Conversely, we can maybe take out the rounding explanation all together, and just use various number examples?  And then say "for more details on behavior, see the [round](https://docs.python.org/3/library/functions.html#round) documentation?

ceddlyburge: I guess bankers rounding means that there is no bias one way or the other.

> The round function uses what is called "Banker's Rounding".
Fractional parts less than 0.5 are rounded down, and fractional parts greater than 0.5 are rounded up.
Fractional parts that are exactly 0.5 are rounded down for even numbers, and up for odd numbers.
For more information, see the documentation for [round()](https://docs.python.org/3/library/functions.html#round).

I like this, but it is maybe a lot of detail for somebody learning python, I think you just rarely if ever need to know or worry about it.

Maybe we could just remove the whole thing, and just have `Lucky for you, Python has the built-in [round()][round] function you can use.`?

BethanyG: I like that!  😄   Let's do that, and remove the detail.  If students are curious, they can read the docs.   And we can always chuck it into the `about.md` for numbers or some other place....just not here.  _Thank you_  💜

ceddlyburge: Cool, I've made that change, let me know what you think ...

BethanyG (approved):
💜   _many thanks!_

PR #3353, opened 2023-03-02
Remove TODO comments

ceddlyburge:
Most of the other concept exercises up to this one communicate the requirement to implement the functions by using docstrings and `pass`, so I think it makes sense to be consistent and do it here as well.


BethanyG:
Agree.  The only concept exercise where we make use of `#TODO` right now is `lasagna` - because of the constants and the missing functions we want the students to learn how to write.  

We might also use them for the actual `functions` concept exercise, given that it will cover creating functions from scratch in more detail.  But for **this** exercise, it makes sense to remove them

BethanyG (approved):
Thanks for giving this the once-over!  Much appriciated.

PR #3355, opened 2023-03-05
[Numbers Concept] Correct arithmetic parenthesis example

MatthijsBlom:


IsaacG:
I think this is the 3rd or 4th PR I've seen with this fix :D It might be
worth opening on the forums to see if the track maintainers would like a PR
for  this.

MatthijsBlom:
I sent this in response to [this thread](http://forum.exercism.org/t/how-is-that-correct-python-learning-track-syllabus/4322).

IsaacG:
My bad. I often check emails and Github notifications before I get around
to seeing what's going on in the forums.

BethanyG:
@MatthijsBlom  - thanks for this.  Another contributor also opened a [PR](https://github.com/exercism/python/pull/3351) for this a while ago,  but I've been swamped with other priorities and so left it closed.  I guess now that there is also fuss on the forum, we should address it before there's a pile-on.

Can you also check the `concept intro` and the `exercise intro` to see if the same mistake is replicated there (and if so fix it)?  _Many thanks_ for your help.  💜

BethanyG:
PS - in future, if you can put what exercise or concept the PR applies to in brackets (_like I did above_) -- that would be super helpful.  That way, the PR sticks out on my notify and email list.  😄

MatthijsBlom:
> Can you also check the `concept intro` and the `exercise intro` to see if the same mistake is replicated there

The exercise being Currency Exchange. Both files are clear. I also checked the exercise instructions for consistency with the exemplar solution.

> PS - in future, if you can put what exercise or concept the PR applies to in brackets

I'll try to remember, but it's easy to forget.

BethanyG (approved):
_Thank you!_

PR #3362, opened 2023-03-09
[Raindrops] Approaches and Performance Article

BethanyG:
Fleshed out the `Introduction.md` and the Following Approaches:

1.  Using `if` statements and `+` to assemble a return string
2.  Exploiting "Truthy" and "Falsy" values in ternary checks and an `f-string`
3.  Using one or more sequences, a `loop`, and an `f-string`
4.  Using one or more sequences and a `generator-expression` within `join()`
5.  Using a `dict` of `{factors : sounds}` for lookup and `join()
6.  Using `itertools.compress()`
7.  Using `functools.reduce()`

**Also added:**
Performance article & Benchmark code


BethanyG:
@colinleach  -- added you to the perf. article as a co-author.  Many thanks for the benchmark code.  😉

PR #3364, opened 2023-03-10
fix typo in the instruction of guidos-gorgeous

thuvh:
- remove double in


BethanyG (commented):
Thanks for the catch!

thuvh:
I create forum link: https://forum.exercism.org/t/fix-typo-in-the-instruction-of-guidos-gorgeous/4571

iHiD:
I've updated this here: #3366

(@ErikSchierboom Do we have a way of reassigning the rep etc to @thuvh?)

ErikSchierboom:
@iHiD We can do an arbitrary rep token, but it would not show up as the "right" contribution I think

PR #2589, opened 2021-09-17
[New Concept Exercise on `lambda`]: Add Secrets draft

Steffan153:
Closes #2357

My checklist:
- [x] Write introduction
- [x] Write instructions
- [x] Write tests
- [x] Write skeleton file
- [x] Write exemplar file
- [x] Write hints
- [x] Write config.json
- [x] Write design.md
- [x] Write concept introduction.md
- [ ] Write concept about.md
- [x] Write concept config.json
- [x] Add something about using `sorted`

PR #3375, opened 2023-03-24
[Sum of Multiples] Add Approaches

MatthijsBlom:


exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md
```diff
@@ -0,0 +1,233 @@
...
+This latter lambda expression evaluates to a function that takes three arguments (`a`, `b`, `x`) and returns the value `a * x + b`.
+Except for not having a name, it is equivalent to the function defined by
+
+```python
+def some_name(a, b, x):
+    return a * x + b
+```
+
+A lambda expression need not necessarily be passed as an argument.
+It can also be applied to arguments immediately, or assigned to a variable:
```
BethanyG, line 176: PEP 8 specifically sets out that you **should not assign a name to a lambda** (emphasis mine):

image

exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/snippet.txt
```diff
@@ -0,0 +1,3 @@
+def sum_of_multiples(limit, factors):
+    is_multiple = lambda n: any([n % f == 0 for f in factors if f != 0])
+    return sum(filter(is_multiple, range(limit)))
```
BethanyG, line 3: This is problematic, as it binds a name to a `lambda`, so needs re-work.

MatthijsBlom: &hellip;I now see I also left a temporarily inserted list comprehension in there.

Would you rather have

```python
def sum_of_multiples(limit, factors):
    return sum(filter(
        lambda n: any(n % f == 0 for f in factors if f != 0),
        range(limit)
    ))
```
and keep the explanation of `lambda`, or
```python
def sum_of_multiples(limit, factors):
    def is_multiple (n):
        return any(n % f == 0 for f in factors if f != 0)

    return sum(filter(is_multiple, range(limit)))
```

BethanyG: Either work for me.  The first might be preferable, considering that you've already done a good explanation of `lambda`.  The second one would need an explanation of nested functions.  Absolutely not opposed to doing that - but it is extra work for you.

`lambda` can sometimes be slower in `filter` or `map`, since it opens another stack frame.  But that's sorta irrelevant to this particular exercise, methinks.

MatthijsBlom: > `lambda` can sometimes be slower in `filter` or `map`, since it opens another stack frame.

Can you elaborate or link to a source on this? I had no idea.

exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md
MatthijsBlom: I agree that we should not encourage learners to do this. I will rework the text to avoid this.

The current text tries hard to explain that _lambda's are expressions_. However, I think this would become significantly harder to convey without the assignment example. There isn't really a lot one can do with a function. Immediately applying a lambda is unfamiliar, as is passing it as an argument (especially so if one is unfamiliar with passing named functions). Assignment on the other hand is _very_ familiar. If lambda's are only ever passed as argument, lambda's being expressions loses [productivity](https://en.wikipedia.org/wiki/Productivity_(linguistics)) (in the linguistic sense). Then they might as well be interpreted as a grammatical element that only occurs in argument lists. (You and I, knowing about lambda's being expressions, know to also expect them in `return`, `yield`, `… if … else …`, etc.)

Please let me know your thoughts about the latest commit. Note: the `lambda` and `filter` sections should probably swap places now (because of the `filter`-less example); I haven't done this yet to keep the diff simple. The changed code samples seem worse to me; I am tempted to rip out the `lambda` entirely.

exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/snippet.txt
BethanyG: The **TL;DR** is that `lambda` has all the overhead/execution of any other function call (_[here's a link](https://pythontutor.com/visualize.html#code=def%20sum_of_multiples%28limit,%20factors%29%3A%0A%20%20%20%20return%20sum%28filter%28%0A%20%20%20%20%20%20%20%20lambda%20n%3A%20any%28n%20%25%20f%20%3D%3D%200%20for%20f%20in%20factors%20if%20f%20!%3D%200%29,%0A%20%20%20%20%20%20%20%20range%28limit%29%0A%20%20%20%20%29%29%0A%0Asum_of_multiples%2810,%20%5B3,%205%5D%29%0A%0Adef%20sum_of_multiples_II%28limit,%20factors%29%3A%0A%20%20%20%20def%20is_multiple%20%28n%29%3A%0A%20%20%20%20%20%20%20%20return%20any%28n%20%25%20f%20%3D%3D%200%20for%20f%20in%20factors%20if%20f%20!%3D%200%29%0A%0A%20%20%20%20return%20sum%28filter%28is_multiple,%20range%28limit%29%29%29%0A%0Asum_of_multiples_II%2810,%20%5B3,%205%5D%29&cumulative=false&curInstr=0&heapPrimitives=nevernest&mode=display&origin=opt-frontend.js&py=3&rawInputLstJSON=%5B%5D&textReferences=false) to python tutor for the two examples above_), so in any situation where you are using a `lambda` over a built-in, or in place of a generator expression that does the same/similar operation, you will incur the 'extra' overhead of the function call.  Trivial in small/medium cases -- but it can add up for larger data sets.

And where you're converting that `filter` + `lambda` into a list or other 'realized' structure, it will be a lot slower than the corresponding comprehension or generator, due to the overhead of calling an additional function _for every item in the list_. 

 But this varies widely (_since python 3.x returns iterators instead of `lists`_) - if you don't need to realize the values and can consume them lazily (_like in a call to `sum()`_), then `filter`/`map`/`reduce` outperform comprehensions, and are mostly even for generators.  But again, a generator that _doesn't_ call an extra function will be faster than `filter` + `lambda` (_because of the `lambda` function call_).

Here are some articles - but many of them assume that values need to be realized, and so aren't really comparing apples to apples.  The finxter blog (_apologies for the aggressive ads there!_) does do the comparisons for both realized and un-realized data - and you can really see the difference.

- [finxter:  Which is Faster?](https://blog.finxter.com/python-lists-filter-vs-list-comprehension-which-is-faster/)
- [Towards Data Science: List Comprehensions vs built-in functions](https://towardsdatascience.com/comparing-list-comprehensions-vs-built-in-functions-in-python-which-is-better-1e2c9646fafe)
- [Stack Overflow: Comprehension vs Lambda + Filter](https://stackoverflow.com/a/66059414)
- [Stack Overflow: Generators vs Filter](https://stackoverflow.com/questions/73226697/generator-expressions-versus-filter)
- [Stack Overflow: Do lambda expressions help reduce execution time?](https://stackoverflow.com/questions/53568926/do-python-lambda-functions-help-in-reducing-the-execution-times)
- (_this one is a bit of a digression, and shows function call overhead_)[How not to be slow using Python: functions](https://pawroman.dev/how-not-to-be-slow-using-python-functions/)
- (_more and wider digression on function cacheing_)[Python caching functions](https://bargsten.org/wissen/python-caching-functions/)

And I'll provide this for completeness, although its really mostly a rant about personal preference with readability, and not on performance:  [Trey Hunner: Stop Writing Lambda Expressions](https://treyhunner.com/2018/09/stop-writing-lambda-expressions/#Overuse:_lambda_with_map_and_filter)

exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md
BethanyG: I came across [this DataQuest tutorial](https://www.dataquest.io/blog/tutorial-lambda-functions-in-python/#:~:text=A%20lambda%20function%20is%20an,and%20returns%20only%20one%20expression.&text=Note%20that%2C%20unlike%20a%20normal,a%20lambda%20function%20with%20parentheses.) that I think does a nice job, but even it assigns a variable to a `lambda` -- then says "don't do that!" , and encourages different practice.  One thing I though was really interesting was that they showed `if/else` in a `lambda` there.

So I know I am asking for something that might not be do-able in a teaching context.  But I think we should try, because all those `yacht` solutions assigning constants to `lambdas` are making me want to _scream_.  

I'll review your changes today, and have another think on it.  Wow, is that upcoming concept about `lambdas` going to be a b*$ch to write.....

MatthijsBlom: > [&hellip;] all those `yacht` solutions assigning constants to `lambdas` are making me want to _scream_.

Might this be catchable with the Analyzer?

> Wow, is that upcoming concept about `lambdas` going to be a b*$ch to write.....

I'm not so worried. There isn't that much more to them than already explained here, is there?

exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/snippet.txt
MatthijsBlom: Thanks. I haven't read all the links yet, but the gist certainly makes sense. I thought previously you meant that

```python
def f(x): return f_body(x)

map(f, xs)
# be faster than
map(lambda x: f_body(x), xs)
```

That in-lined functions in comprehensions are faster than mapped functions I already expected.

exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md
BethanyG: > > […] all those `yacht` solutions assigning constants to `lambdas` are making me want to _scream_.
> 
> Might this be catchable with the Analyzer?

Oh, YESH.  Buuut...first I need to upgrade to Python 3.11.  Then I need to get the analyzer turned on for practice exercises.  Then I can enable that PyLint Rule for `Yacht`.  Maybe when I get there, I can ping you and we can see if there is messaging we can add that isn't just "PEP 8 says NO".  😁 


> > Wow, is that upcoming concept about `lambdas` going to be a b*$ch to write.....
> 
> I'm not so worried. There isn't that much more to them than already explained here, is there?

_You volunteering to get the [`lambda` concept and exercise](https://github.com/exercism/python/pull/2589) over the finish line?? (or start over with something else?)_ Because it can be ALL YOURS.  😄

IsaacG: Just my two cents here. While I generally agree with PEP-8 and most the linter outputs ... I think `lambda` is a fairly concise and clean way to solve Yacht. There are exceptions to every rule and I don't think using `lambda` for Yacht should trigger a warning.

BethanyG: @IsaacG - While I appreciate your two cents here, my comment about Yacht here was an _aside_.  If you'd like to debate a `lambda` check for Yacht, that's best done on the forum.

exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md
```diff
@@ -0,0 +1,253 @@
...
+
+```python
+for word in filter(str.isupper, ["THUNDERBOLTS", "and", "LIGHTNING"]):
+    print(word)
+# prints:
+# THUNDERBOLTS
+# LIGHTNING
+```
+
+An iterator is a bit like a cursor that can move only to the right.
```
BethanyG, line 100: I really like this explanation.  Both from a cursor for typing ... and a cursor from a DB.  Both can only be gone through once, and you can't back up.  😄

exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md
```diff
@@ -0,0 +1,253 @@
...
+~~~~exercism/note
+Immediately applying a lambda expression is possible, but generally pointless:
+
+```python
+# Instead of
+(lambda a, b, x: a * x + b)(2, 3, y)
+# you might as well write
+2 * y + 3
+```
+~~~~
```
BethanyG, line 213:
```suggestion
```
```

Comment (MatthijsBlom) [reply exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md:213]: That `~~~~` is matched with the `~~~~exercism/note` 9 lines up.

Comment (BethanyG) [reply exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md:213]: Huh.  Ok.  It was reading really weird to me.  Feel free to ignore.

Comment (MatthijsBlom) [reply exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md:213]: Was it the code or the rendering that read weird?

Comment (BethanyG) [reply exercises/practice/sum-of-multiples/.approaches/filter-for-multiples/content.md:213]: It was the rendering, which is sorta gross on GH:

image

<br>

But it was more that I missed the first `note` block, as I wasn't expecting it.  😄   Now that I see there is a note and a `caution`, it makes sense.

**Comment by MatthijsBlom** [comment]
U-oh, yet another approach:

```python
from heapq import heapify, heapreplace
from itertools import takewhile


def sum_of_multiples(limit, factors):
    def multiples():
        queue = [(_, _) for _ in factors if _ != 0]
        heapify(queue)
        previous_multiple = 0
        while queue:
            multiple, factor = queue[0]
            if multiple != previous_multiple:
                yield multiple
                previous_multiple = multiple
            heapreplace(queue, (multiple + factor, factor))

    return sum(takewhile(lambda n: n < limit, multiples()))
```

MatthijsBlom:
@BethanyG Do you still wish for these lambda assignments to be avoided?

My own preference is to revert most of 5dbe577. I could add some more prose on the matter, or lift the deleted paragraph into a `exercism/caution`, or something.

PR #3386, opened 2023-04-05
[Darts] : Add Approaches

BethanyG:


exercises/practice/darts/.approaches/booleans-as-ints/content.md
colinleach, line 1: Line 33, word missing. Should be "it is also more difficult..."

exercises/practice/darts/.approaches/dict-and-generator/content.md
colinleach, line 1: Line 31 should be "requires *an* index..."

colinleach:
Looking good! I spotted a couple of small typos, but nothing major, and it reads well.

As a Brit (specifically, a _northern_ Brit, from a region where this is a very common pub game), I find it startling to see a dart throw called a "toss". However, that's from `problem-specifications` and not worth messing with. Is this standard American English?

BethanyG:
> _As a Brit (specifically, a northern Brit, from a region where this is a very common pub game), I find it startling to see a dart throw called a "toss". However, that's from problem-specifications and not worth messing with. Is this standard American English?_

TIL that it was William 'Bigfoot' Anakin we have to thank for it being a pub game.  He proved it wasn't a "game of chance", and therefore could be allowed in Pubs.  [_who knew?_](https://en.wikipedia.org/wiki/Darts).  And a Lancashire carpenter (_Brian Gamlin_) who came up with the scoring system.

`Toss` is what I intuitively used, but I _suspect_ the canonical data primed me to use that.  Being that this game likely originated in The British Isles and France (_where it was called `butt` at one point_), perhaps we should use the term that they most commonly use.  I'll comb through and replace things with `throw`.  😄

colinleach:
> I'll comb through and replace things with throw. 😄

A pretty gutsy decision, to open this can of worms! I'll do nothing for today, hoping you'll just merge your PR and move on to something more urgent...

But it is, of course, your choice.

BethanyG:
> _A pretty gutsy decision, to open this can of worms! I'll do nothing for today, hoping you'll just merge your PR and move on to something more urgent..._

I mean, we also have the issue that this really isn't _Darts_ -- we use an Archery target for the exercise icon, and we are only doing the 1x, 2x, 3x portion of the target -- not the individual scoring sections.  But I am not going to be that a-hole that renegotiates the exercise for the sake of .. _reasons_ 😉   In fact, we can keep the description as-is until someone notices or brings up `toss` vs `throw`.  Then we can wade in at that point.  Or not.

But it did give me a flash of inspiration:  If we make an exercise that does flesh out the Darts rules, we could possibly set that up as the complex numbers learning exercise.  **_Bear with me here ....._** since you could use the imaginary part for rotation, that could be the setup for scores, with the distance from center being the multipliers.   I would have to really play with it to see if it could be set up without massive confusion, but.

I am indeed going to move on to more urgent concerns.  This is about getting the queue down, and also getting your work reviewed and merged.  **Thank you** for reviewing this.  Never thought I would be doing so much writing for a coding site.

PR #3388, opened 2023-04-06
[Largest Series Product] Synchronize Addendum Example Code with Tests

MatthijsBlom:
This patch assumes that the intent of this `instructions.append.md` was to match the actual requirements.


BethanyG:
Leaving a note here:  https://github.com/exercism/problem-specifications/pull/2246 is discussing the removal of the tests.  Once it is PR'd to the tracks, this PR will need to be adjusted/amended.  

We'll need to adjust the [JinJa2 template](https://github.com/exercism/python/blob/main/exercises/practice/largest-series-product/.meta/template.j2), regenerate the test files, and then re-edit this addendum accordingly.

BethanyG (approved):
Alrighty.  Now that we have things sorted with the intro and which tests are included, we can merge this!  Thank you for submitting it.  😄

PR #3394, opened 2023-04-10
Fix English-Ukrainian translation typo

HliebS:
The provided translation was from English to Russian. In Ukrainian "nine" is written with an apostrophe - дев'ять.


iHiD:
@HliebS Thanks!

PR #3397, opened 2023-04-11
[Tests.toml] Synced Deprecated Exercises.

BethanyG:
These exercises are deprecated for the track, so the `tests.toml` files have been updated to exclude all "new" tests.  The thought being that if we un-deprecated these exercises, they'd be in the old state, and we could see where they'd need changes.

- new file:   exercises/practice/accumulate/.meta/tests.toml
- new file:   exercises/practice/parallel-letter-frequency/.meta/tests.toml
- new file:   exercises/practice/strain/.meta/tests.toml


kytrinyx (approved):
That makes a lot of sense.

PR #3402, opened 2023-04-12
[list-ops]: change `//` to `/` and swap `foldr` arg order

IsaacG:
[Related Forum Discussion here](https://forum.exercism.org/t/bug-in-python-list-ops-test/4835/31).

### DIFF
--- a/exercises/practice/list-ops/.meta/additional_tests.json
+++ b/exercises/practice/list-ops/.meta/additional_tests.json
@@ -6,7 +6,7 @@
       "input": {
         "list": ["e", "x", "e", "r", "c", "i", "s", "m"],
         "initial": "!",
-        "function": "(x, y) -> x + y"
+        "function": "(acc, el) -> el + acc"
       },
       "expected": "exercism!"
     },
--- a/exercises/practice/list-ops/.meta/example.py
+++ b/exercises/practice/list-ops/.meta/example.py
@@ -29,7 +29,7 @@ def foldr(function, list, initial):
     if len(list) == 0:
         return initial
     else:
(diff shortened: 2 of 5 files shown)


BethanyG:
@IsaacG  -- you should add your name as a contributor to this exercise.  😄

BethanyG:
Can we also put in a comment in the JinJa2 template to maybe remind the next set of folx (_more than likely our future selves_) why we're swapping arguments?  I just think that this will come up again, so any places we can comment for posterity will be helpful.  [JinJa2 Comments](https://tedboy.github.io/jinja2/templ5.html).

IsaacG:
> Can we also put in a comment in the JinJa2 template to maybe remind the next set of folx (_more than likely our future selves_) why we're swapping arguments? I just think that this will come up again, so any places we can comment for posterity will be helpful. [JinJa2 Comments](https://tedboy.github.io/jinja2/templ5.html).

Sure! I can add that once I get homen in a few hours. 

> If we are going to break existing solutions, I would vastly prefer to have foldr & foldl both be (acc, el), so that foldl is consistent with Haskell tradition. But I am going to go with your first PR, so we can be done with this debate. 😄

That was actually my first commit. That approach is simpler (less code, no need to swap, no need for a comment). I'm happy to do that, too. It's closer to what is in the problem spec.

IsaacG:
I added a comment to the JinJa2 template.

If you prefer `func(acc, el)`, #3405 does that.

BethanyG:
Alright.  I am going to go with #3405, and close this one in favor of that.  @IsaacG -- can you add yourself as an exercise contributor to that PR?  As soon as that is done, I'll approve and merge.  Many thanks!

PR #3403, opened 2023-04-12
[list-ops]: change // to / and swap foldl/foldr arg order to both call func(el, acc)

IsaacG:


BethanyG:
@IsaacG - If we are going to break existing solutions, I would vastly prefer to have `foldr` & `foldl` both be (acc, el), so that `foldl` is consistent with Haskell tradition.  But I am going to go with your first PR, so we can be done with this debate.  😄

PR #3405, opened 2023-04-13
[list-ops]: change // to / and make have foldl/foldr call func(acc, el), consistent with the problem spec

IsaacG:


BethanyG:
@IsaacG  -- Thank you for this.  Pls remember to add yourself as an exercise contributor.  After that, I will approve and merge.

IsaacG:
> @IsaacG -- Thank you for this. Pls remember to add yourself as an exercise contributor. After that, I will approve and merge.

Done. Thanks!

BethanyG (approved):

asarkar:
The arguments to the lambda are inconsistent with the problem description.

> foldr (given a function, a list, and an initial accumulator, fold (reduce) each item into the accumulator from the right using function(item, accumulator));

```
    def test_foldr_direction_dependent_function_applied_to_non_empty_list(self):
        self.assertEqual(foldr(lambda acc, el: el / acc, [1, 2, 3, 4], 24), 9)
```

The first argument to the lambda should be `el`, not `acc`.

BethanyG:
@asarkar - if you'd like to report an issue, please do so on our [forum](https://forum.exercism.org/).  A long-closed PR is not the place to do it.  Thanks.

PR #3412, opened 2023-04-22
Removed footer macro and regenerated test file

gelbelle:
All tests passed with new test files.

When I save the template.j2 files it auto formats and removes the indents. It seems to work without being indented though.

Let me know if you would like me to change this.

### DIFF
--- a/exercises/practice/all-your-base/.meta/template.j2
+++ b/exercises/practice/all-your-base/.meta/template.j2
@@ -1,32 +1,30 @@
 {%- import "generator_macros.j2" as macros with context -%}
 
 {%- macro func_call(case) -%}
-    {{ case["property"] }}(
-        {% for arg in case["input"].values() %}
-            {{ arg }}{{- "," if not loop.last }}
-        {% endfor %}
-    )
+{{ case["property"] }}(
+{% for arg in case["input"].values() %}
+{{ arg }}{{- "," if not loop.last }}
+{% endfor %}
+)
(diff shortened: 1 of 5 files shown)


BethanyG:
Awesome start!  Thank you for these.  😄 

We'll need to also commit the test files you regenerated from the templates for each exercise, or the CI will complain about them. 

> When I save the template.j2 files it auto formats and removes the indents. It seems to work without being indented though.

If the process still works (IIRC, there have been times when it doesn't) and creates valid test files that the examples pass -- we're good.  😄

BethanyG:
🤔  So. Upon  reading your message in Discord and trying to generate a Python test file from these templates, it would appear that removing indents has negative consequences.  **You're going to have to disable your auto-formatter.**  I took a look to see if there were any valid auto-formatters for JinJa2 Python templates, and found something _close_, but it still mangles the expected output lines for the test cases, so we can't use it at this time. 

Removing indents in the templates removes the indents from the Python output, and then Black has a freakout (_those error messages you were seeing when you ran the generate script_).  For some lines, even a single space or other wrapping can screw things up.  🙄 

It sucks, but that's the breaks.  I think the best thing at this point - rather than try to re-indent things - is to start over with a new branch, a turned-off auto-formatter, and re-do these templates.  If you get any errors when you then run the `generate_tests.py` script, ping me on Discord, and we can walk through the errors together and resolve them. 😄

PR #3413, opened 2023-04-23
Removed footer macros

gelbelle:
Regenerated test files and checked that all tests still pass with example.py


BethanyG:
Hi @gelbelle 👋🏽 

Thank you for this PR. 😄  This is looking better!  🚀   I think there may still be something wrong with your branch tho - this looks like it is coming from `gelbelle:main`, and it seems that there are still un-indented templates and missing test files for: 

- `all-your-base`
-  `beer-song`
-  `armstrong-numbers`
-  `atbash-cipher`
-  `binary-search-tree`
-  `alphametics`

We can pull those out of this PR, or if you'd like to troubleshoot them, let me know.  But as-is, this is not passing CI because there is something wrong with those templates, and the test files are missing.

PR #3436, opened 2023-06-08
Fix formatting issue

safwansamsudeen:
This doesn't have a forum post associated, as it's a basic typo correction. The note currently looks like this:
image

I believe by changing this it should be formatted correctly. If perhaps you use an automated system and this PR never gets looked at, I'll open a topic on the forum about it.


BethanyG:
@safwansamsudeen  -- this is a practice problem which means its instructions are pulled from the [problem-specifications](https://github.com/exercism/problem-specifications/blob/main/exercises/secret-handshake/instructions.md) repo, so you have to PR the change there.

safwansamsudeen:
I forgot - thanks!

PR #3437, opened 2023-06-10
Remove Python 2 references from the track

IsaacG:
Remove Python 2 references from any files in the track.

### DIFF
--- a/exercises/practice/binary-search/binary_search_test.py
+++ b/exercises/practice/binary-search/binary_search_test.py
@@ -9,33 +9,26 @@ from binary_search import (
 
 class BinarySearchTest(unittest.TestCase):
     def test_finds_a_value_in_an_array_with_one_element(self):
-
         self.assertEqual(find([6], 6), 0)
 
     def test_finds_a_value_in_the_middle_of_an_array(self):
-
         self.assertEqual(find([1, 3, 4, 6, 8, 9, 11], 6), 3)
 
     def test_finds_a_value_at_the_beginning_of_an_array(self):
-
         self.assertEqual(find([1, 3, 4, 6, 8, 9, 11], 1), 0)
 
     def test_finds_a_value_at_the_end_of_an_array(self):
-
         self.assertEqual(find([1, 3, 4, 6, 8, 9, 11], 11), 6)
 
     def test_finds_a_value_in_an_array_of_odd_length(self):
-
         self.assertEqual(
             find([1, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 634], 144), 9
         )
 
     def test_finds_a_value_in_an_array_of_even_length(self):
-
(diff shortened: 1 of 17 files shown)


BethanyG:
@IsaacG - left you a not on Discord, but noting here too (so we have a record).

It is an unqualified PITA, but even blank lines in test files need to go through the templating system, sadly.  So for the following exercises/test files, you'll need to edit the JinJa2 template, then regenerate the test files and check BOTH into the PR.  Otherwise, the CI will complain and/or fail. 

I'd hold `complex-numbers` for its own PR.  Looks like the upstream tests have changed, so we'll need to run configlet, judge which cases we want to keep or omit, and re-do things.  Considering that we recently changed how that exercise works on this track, we probably want to take a hard look at it and make sure that nothing has gone wrong there.

The other exercises that need their template + re-generation are:

- [ ] Binary Search
- [ ] Rational Numbers
- [ ] Robot Simulator
- [ ] Roman Numerals
- [ ] Scale Generator

Option B is to just leave them out, and we can deal with them when the whole re-queueing thing is completed.

exercises/practice/binary-search/binary_search_test.py
BethanyG, line 1: Can we omit this change?  Its only the removal of lines in the file and isn't related to Python 2.7.

exercises/practice/complex-numbers/complex_numbers_test.py
BethanyG, line 1: Can we omit this file change as well?  Not related to Python 2.7.

exercises/practice/rational-numbers/rational_numbers_test.py
BethanyG, line 1: Again, this is only a blank line removal.  Let's omit this file since its not related to the Python 2.7 change.

exercises/practice/robot-simulator/robot_simulator_test.py
BethanyG, line 1: Omit this as well.

exercises/practice/roman-numerals/roman_numerals_test.py
BethanyG, line 1: And this one.

exercises/practice/scale-generator/scale_generator_test.py
BethanyG, line 1: And this one.

BethanyG (changes requested):
Please remove the test files that only have blank lines removed or added.  They're not really related to Python 2.7 removal, and they are causing the CI to not pass.  

Additionally, those files are going to have to be regenerated once the whole re-queueing issues are resolved, so I'd rather just group or change them then.  Thanks!

IsaacG:
Whitespace-only test file changes have been reverted. Thanks!

BethanyG:
Putting this on hold until the issue related #3373 is resolved & pushed.

BethanyG (approved):
LGTM.

PR #3443, opened 2023-06-16
Correct formatting issue

safwansamsudeen:


BethanyG (approved):
I do this all the time!  😆   Sad thing is that I failed to catch it in review.  😱 

One thing that helps me remember is to put in the link placeholders in the text while I am writing, and let my IDE highlight them as errors.  That then reminds me to put the bottom links in before I commit (_well...at least sometimes_).  Another strategy I've used is to compile the links I want ahead of time at the bottom, so that I remember to put them into the text.  But it is a ... process.  😄

safwansamsudeen:
Haha, relieved to know that!

PR #3444, opened 2023-06-17
Remove ResistorColorMaster reference

BNAndras:


BethanyG (commented):
Many thanks for this, @BNAndras!  🌟   Normally, this would be flagged as "small", but I think this might be your first contribution here, so giving you the full "med" rep.  😄 🚀

BethanyG (approved):

PR #3445, opened 2023-06-18
[Dictionary & Dictionary-Methods]: Updates & Edits + "Mecha Munch Management" Concept Exercise

BethanyG:


concepts/dict-methods/introduction.md
```diff
@@ -1,16 +1,28 @@
...
-The `dict` class in Python provides many useful [methods][dict-methods] for working with dictionaries.
-Some are introduced in the concept exercise for `dicts`.
-This concept tackles a few more - along with some techniques for iterating through and manipulating `dicts`.
+`dict`s enable the retrieval of a `value` in (on average) constant O(1) time, given the `key`.
+Compared to searching for a value within a `list` or `array` (_without knowing the `index` position_), a dictionary uses significantly more memory, but has very rapid retrieval.
+Dictionaries are especially useful in scenarios where the collection of items is large and must be accessed and updated frequently.
 
+The `dict` class in Python provides many useful [methods][dict-methods], some of which are introduced in the concept exercise for dictionaries.
+
+This concept tackles a few more:
```
vaeng, line 14: I like the handy reference

concepts/dict-methods/introduction.md
```diff
@@ -1,16 +1,28 @@
...
+Dictionaries are Python's only built-in [mapping type][mapping-types-dict].
+As of Python 3.7, `dict` key order is guaranteed to be the order in which entries are inserted.
 
-Compared to searching for a value within a list or array (_without knowing the index position_), a dictionary uses significantly more memory, but has very rapid retrieval.
-It's especially useful in scenarios where the collection of items is large and must be accessed/updated frequently.
 
-The `dict` class in Python provides many useful [methods][dict-methods] for working with dictionaries.
-Some are introduced in the concept exercise for `dicts`.
-This concept tackles a few more - along with some techniques for iterating through and manipulating `dicts`.
+`dict`s enable the retrieval of a `value` in (on average) constant O(1) time, given the `key`.
```
vaeng, line 8: Maybe the more advanced compsci topics like big O can be simplified?
```suggestion
Given the `key`, dict`s enable the retrieval of a `value` in (on average) constant time (independent of the number of entries).
```

concepts/dicts/about.md
```diff
@@ -1,55 +1,271 @@
...
-  `values` can be of any or multiple data type(s) or structures, including other dictionaries, built-in types, custom types, or even objects like functions or classes.
-   `values` associated with any `key` are _mutable_, and can be replaced, updated or altered as long as the `key` entry exists.
-  Dictionaries enable the retrieval of a `value` in (on average) constant O(1) time, given the `key`.
+`Keys` can include `numbers`, `str`, `tuples` (of _immutable_ values), or `frozensets`, but must be hashable and unique across the dictionary.
+  `values` can be of any or multiple data type(s) or structures, including other dictionaries, built-in types, custom types, or even objects like functions.
 
-  Compared to searching for a value within a `list` or `array` (_without knowing the `index` position_), a `dict` uses significantly more space in memory, but has significantly more rapid retrieval.
- Dictionaries are especially useful in scenarios where the collection of items is large and must be accessed and/or updated frequently.
+`dict`s enable the retrieval of a `value` in (on average) constant O(1) time, given the `key`.
```
vaeng, line 10: I would remove O(1), when constant is in the same sentence anyway.

concepts/dicts/about.md
```diff
@@ -1,55 +1,271 @@
...
+As of Python 3.7, `dict` key order is guaranteed to be the order in which entries are inserted.
 
-A dictionary (`dict`) is a [mapping type][mapping-types-dict] data structure that associates [hashable][term-hashable] `keys` to `values` -- known in other programming languages as a resizable [hash table or hashmap][hashtable-wikipedia].
-  `Keys` can include `numbers`, `str`, `tuples` (of _immutable_ values), or `frozensets`, but must be hashable and unique across the dictionary.
-  `keys` are _immutable_ - once added to a `dict`, they can only be removed, they cannot be updated.
-  `values` can be of any or multiple data type(s) or structures, including other dictionaries, built-in types, custom types, or even objects like functions or classes.
-   `values` associated with any `key` are _mutable_, and can be replaced, updated or altered as long as the `key` entry exists.
-  Dictionaries enable the retrieval of a `value` in (on average) constant O(1) time, given the `key`.
+`Keys` can include `numbers`, `str`, `tuples` (of _immutable_ values), or `frozensets`, but must be hashable and unique across the dictionary.
```
vaeng, line 7: Do students know about `frozensets` at this point? Could this be shortened to "numbers, str, or other hashable types?"

concepts/dicts/about.md
```diff
@@ -1,55 +1,271 @@
...
+  `values` can be of any or multiple data type(s) or structures, including other dictionaries, built-in types, custom types, or even objects like functions.
```
vaeng, line 8: Is there something that can't be a value?

concepts/dicts/about.md
```diff
@@ -1,55 +1,271 @@
...
+              'color': 'Brindle'}
+
+>>> wombat
+{'name': 'Wombat', 'speed': 23, 'land_animal': True, 'color': 'Brindle'}
+```
+
+Dictionaries can be arbitrarily nested:
+
+```python
+animal_collection = {
```
vaeng, line 61: Nice example! It might be to wide to be displayed correctly on the website?

vaeng (approved):
That is a huge exercise! I would love to try it. I think I would need a lot of time. At least if feels that way after reading through (most) of it. I really like the concepts, but I wonder if the exercises can touch all those topics that are touched in the two concepts.

It might feel overwhelming to some students to read through the two introductions and the instruction for mega munch management.

(It is possibly the github view and it is far less of a wall of text, than it looks like over here.)

I think it is a good idea to have two exercises on the dict topics and split them up

BethanyG:
>  _but I wonder if the exercises can touch all those topics that are touched in the two concepts._

Oh!  Apologies!!  This was a big roundup of edits needed for two separate concepts and two linked concept exercises.

1.  **Dictionaries**:  updated the concept and made corrections to the `inventory-management` concept exercise.
2.  **Dictionary Methods**:  Pulled out methods from the dictionary concept, to cover in this new concept.  Also used those to create the `Mecha Munch Management` exercise.

So there are two different concepts and two different exercises. Are you proposing they be broken into 3 or 4?

concepts/dicts/about.md
BethanyG: erg.  Yeah.  Big `dicts` are hard.  Do you know how wide the code window is?  I should format to that.

concepts/dicts/about.md
BethanyG: I haven't hit a limit.   This is more to let students know that `value` isn't limited to "primitive" values.   Dictionary values can be of arbitrary types, and arbitrarily nested.  🤔   Maybe that's what I should say here instead?

concepts/dicts/about.md
BethanyG: Since this is an `about.md`, I feel a bit more comfortable sending students to resources or challenging them with things that haven't been introduced.  The "hashable" thing is tricky in Python:

> **hashable**
An object is hashable if it has a hash value which never changes during its lifetime (it needs a __hash__() method), and can be compared to other objects (it needs an __eq__() method). Hashable objects which compare equal must have the same hash value.
>
>Hashability makes an object usable as a dictionary key and a set member, because these data structures use the hash value internally.
>
>Most of Python’s immutable built-in objects are hashable; mutable containers (such as lists or dictionaries) are not; immutable containers (such as tuples and frozensets) are only hashable if their elements are hashable. Objects which are instances of user-defined classes are hashable by default. They all compare unequal (except with themselves), and their hash value is derived from their [id()](https://docs.python.org/3/library/functions.html#id).

Maybe here we simple refer (by link) to the glossary definition?

concepts/dicts/about.md
BethanyG: Whelp, the code window is about 67 characters.  Tried to format it, but left some overlap.  It just doesn't read well if I do too much.

concepts/dicts/about.md
BethanyG: Ok.  I **_think_** I've made it better.  At least I hope so.  😉

BethanyG:
Ugh.  Going to merge this as a WIP, and come back for more cleanup.  Thank you so much for the feedback.

PR #3447, opened 2023-06-19
correct example

12rambau:
If I read correctly the instructions of the game then "3 3 3 3 3" is a Yacht so it count for 50 points. thus the example for "threes" is wrong. I updated the example numbers, keeping the correct number of "3" to keep 15 as a result.


IsaacG:
The same set of dice can be scored in multiple categories. The goal is not to determine if a hand is a THREES or a YACHT. The goal is to figure out what the score is for a given set of dice using a specific category. The instructions are entirely correct; `[3, 3, 3, 3, 3]` when scored in the THREES category, should return 15 points.

12rambau:
my bad deleting https://github.com/exercism/problem-specifications/pull/2292 then

PR #3450, opened 2023-07-01
Approaches for Robot Name

safwansamsudeen:
Suggestions?

### DIFF
--- a/exercises/practice/sublist/.approaches/introduction.md
+++ b/exercises/practice/sublist/.approaches/introduction.md
@@ -53,8 +53,7 @@ def sublist(list_one, list_two):
         return SUPERLIST
     return UNEQUAL
 ```
-To understand more about this approach and **why it fails**, [read here]
-[approach-using-strings].
+To understand more about this approach and **why it fails**, [read here][approach-using-strings].
 
 [approach-list-manipulation]: https://exercism.org/tracks/python/exercises/sublist/approaches/list-manipulation
 [approach-using-strings]: https://exercism.org/tracks/python/exercises/sublist/approaches/using-strings


exercises/practice/robot-name/.approaches/mass-name-generation/content.md
```diff
@@ -0,0 +1,42 @@
...
+letter_pairs = (''.join((l1, l2)) for l1 in ascii_uppercase for l2 in ascii_uppercase)
+numbers = (str(i).zfill(3) for i in range(1000))
+names = [l + n for l in letter_pairs for n in numbers]
+```
+
+After the name generation, the names are shuffled - using the [default `seed`][random-seed] in the `random` module (the current timestamp). 
+When the tests reseed `random`, this has no effect as the names were shuffled before that.
+
+We then set `NAMES` to the iterable of names, and in `reset`, set the robot's name to the  `next(name)`. 
+If you'd like, read more on [`iter` and `next`][iter-and-next].
```
IsaacG, line 38: It might be worth discussing the pros and cons/drawbacks of the different approaches. The mass generation has a high start up cost and memory cost. It's hugely inefficient if we're going to generate a relatively small number of names.

Conversely, the "on the fly" approach can be very expensive for a large number of names, since collisions occur more often as the number of available names gets smaller.

safwansamsudeen: Shouldn't that come under articles instead of approaches? Perhaps a follow-on PR?

IsaacG: > I like the idea, but shouldn't that come under articles instead of approaches?

The approaches aren't just a list of possible solutions; they are meant to dig deeper into the problem and different approaches to solving the problem. I think it makes sense to discuss the pros and cons of approaches alongside the approach. When describing "you can solve it like this", it is good to provide context and details, like why or when you would want to solve it this way. 

> Perhaps a follow-on PR?

Whatever works for you.

safwansamsudeen: I've committed, but out of curiosity, what's the difference between approaches and articles? I thought articles were supposed to compare approaches' performance or other interesting stuff like that, while approaches were meant to detail how an exercise could be solved in a certain manner.

BethanyG: @safwansamsudeen -- you are correct.  Here are the [approaches docs](https://exercism.org/docs/building/tracks/approaches).  Here are the [articles docs](https://exercism.org/docs/building/tracks/articles).

But also:  I have not yet had time to review this.  I thought you were going to put this on hold?

IsaacG: Article is definitely better suited for in depth comparison/exploration. For approaches, the docs say,

> Feel free to dig deep into the topic

I wouldn't expect an extensive comparison of approaches in the approach docs, but I think a line or three giving a high level overview of the pros and cons of the approach is appropriate.

BethanyG:
I'm not going to be able to look at this until next week, so I will comment/approve/disapprove then.

safwansamsudeen:
Ugh, git troubles. Survived a scare after wiping out all of this.

New PR: #3451. My apologies!

PR #3458, opened 2023-07-13
[RNA-Transcription]: Removed JinJa2 Footer Macro & Regenerated Test File

BethanyG:
@iHiD @ErikSchierboom -- I was going to list out a small PR for testing, but realized that all of the ones queued by Gelbelle included 10-12 files, which is probably too big a chunk.  This PR only touches the test template and file for one exercise.

This is not a change that should re-queue solutions, as it is only removing the 

```python
if __name__ == "__main__":
    unittest.main()
```

Footer from the test file.

### DIFF
--- a/exercises/practice/rna-transcription/.meta/template.j2
+++ b/exercises/practice/rna-transcription/.meta/template.j2
@@ -9,5 +9,3 @@ class {{ exercise | camel_case }}Test(unittest.TestCase):
             '{{ case["expected"] }}'
         )
     {% endfor %}
-
-{{ macros.footer() }}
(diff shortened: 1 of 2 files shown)


ErikSchierboom:
@BethanyG Looking at this PR, it changes two files:

1. A file in `.meta`, which is ignored when we considering whether to update an exercise
2. A test file, which _is_ affecting the exercise version

In this case, if this PR is merged as normal, it would trigger re-runs because we can't know that the test file does not actually influence any test results. However, we allow track maintainers to tell us they know this change can't influence any test results by adding the `[no important files changed]` string somewhere in their commit message. Could you update your PR to include this magic marker? It can be anywhere in the commit message and we'll do a case-insensitive comparison.

We have also added a feature that will ignore any changes to `.docs` for the purpose of deciding whether to re-run the tests. It _is_ still used to determine whether to update an exercise (we do want students to get to update documentation), but if a PR _only_ contains changes to `.docs` files, we'll not re-run the tests.

PR #3470, opened 2023-07-19
Update lists_test.py

ryuheechul:
`sum([2,3,4,8,8]) / 5` is 5 not 4 but median here is `4`.

So this can be fixed by below.

```diff
- [2, 3, 4, 8, 8]
+ [2, 3, 5, 7, 8]
```

Alternative to this one:
change the second last value of `results` to `False`


IsaacG:
Note the instructions say,

> She has thought of **two ways** of getting an average-like number

PR #3484, opened 2023-08-01
[Collatz Conjecture]: Update recursion approach snippet

AndrewLawendy:
The current approach to recursion is not ideal, as it utilizes a suboptimal way of returning values.
This approach should not be encouraged or taught to students, as it may lead to the use of tail recursion.

An alternative approach to recursion, which is considered more ideal, can be found at https://exercism.org/tracks/python/exercises/collatz-conjecture/solutions/bobahop.

Reference: https://stackoverflow.com/questions/33923/what-is-tail-recursion

### DIFF
--- a/exercises/practice/collatz-conjecture/.approaches/recursion/snippet.txt
+++ b/exercises/practice/collatz-conjecture/.approaches/recursion/snippet.txt
@@ -1,12 +1,8 @@
-def count_steps(conjecture_steps, number):
-    if number <= 1:
-        return conjecture_steps
-    return count_steps(
-        conjecture_steps + 1, number * 3 + 1 if number % 2 else number / 2
-    )
-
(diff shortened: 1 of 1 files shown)


AndrewLawendy:
Ah, OK, thanks for sharing that.

I come from a JS background, and this code could lead to stack overflow.
I'll check the references.

BethanyG:
> I come from a JS background, and this code could lead to stack overflow.

Yup.  It will in Python as well, without memoization or careful management.  If you read to the bottom of the approach, it talks a bit about Python's recursion limit because due to this danger.

BethanyG:
I am going to go ahead and close this.  I've alerted the other author of the approach, who is going to review things and maybe add a little more explanation or links about how recursion works in Python.  🙂

PR #3489, opened 2023-08-18
little-sisters-vocab: Update stub to match instructions

iHiD:
Tiny tweak based on https://forum.exercism.org/t/little-sisters-vocabulatry-discrepency-between-test-and-docstring-for-adjective-to-verb/6949/3


BethanyG:
Thanks for this.  It was on my list for the afternoon, but now it isn't!  😄

PR #3490, opened 2023-08-19
Fix link rendering

safwansamsudeen:
Three links in the approaches I authored weren't being rendered. I think this should fix them.


BethanyG (approved):
Thanks for this.

PR #3501, opened 2023-09-11
[meltdown-mitigation] Be super explicit

iHiD:
It seems a few people get stuck on this (maybe because the sentence is quite long and for non-native speakers it's tough to parse - e.g. "plus or minus 10%" could feel confusing maybe?) so I thought maybe we become SUPER explicit.

(To be clear, I find the instructions very clear already, but it seems others don't)

Some context here: https://forum.exercism.org/t/exercise-meltdown-mitigation/6545/2

### DIFF
--- a/exercises/concept/meltdown-mitigation/.docs/instructions.md
+++ b/exercises/concept/meltdown-mitigation/.docs/instructions.md
@@ -61,7 +61,7 @@ Implement the function called `fail_safe()`, which takes 3 parameters: `temperat
 - If `temperature * neutrons_produced_per_second` < 90% of `threshold`, output a status code of 'LOW'
   indicating that control rods must be removed to produce power.
 
-- If `temperature * neutrons_produced_per_second` are within 10% of the `threshold` (so either 0-10% less than the threshold, at the threshold, or 0-10% greater than the threshold), the reactor is in _criticality_ and the status code of 'NORMAL' should be output, indicating that the reactor is in optimum condition and control rods are in an ideal position.
+- If the value `temperature * neutrons_produced_per_second` is within 10% of the `threshold` (so either 0-10% less than the threshold, at the threshold, or 0-10% greater than the threshold), the reactor is in _criticality_ and the status code of 'NORMAL' should be output, indicating that the reactor is in optimum condition and control rods are in an ideal position.
 
 - If `temperature * neutrons_produced_per_second` is not in the above-stated ranges, the reactor is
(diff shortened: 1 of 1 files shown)


exercises/concept/meltdown-mitigation/.docs/instructions.md
```diff
@@ -61,9 +61,7 @@ Implement the function called `fail_safe()`, which takes 3 parameters: `temperat
 - If `temperature * neutrons_produced_per_second` < 90% of `threshold`, output a status code of 'LOW'
   indicating that control rods must be removed to produce power.
 
-- If `temperature * neutrons_produced_per_second` are within plus or minus 10% of the `threshold`
-  the reactor is in _criticality_ and the status code of 'NORMAL' should be output, indicating that the
-  reactor is in optimum condition and control rods are in an ideal position.
+- If `temperature * neutrons_produced_per_second` are within 10% of the `threshold` (so either 0-10% less than the threshold, at the threshold, or 0-10% greater than the threshold), the reactor is in _criticality_ and the status code of 'NORMAL' should be output, indicating that the reactor is in optimum condition and control rods are in an ideal position.
```
BethanyG, line 64:
```suggestion
- If the value `temperature * neutrons_produced_per_second` is within 10% of the `threshold` (so either 0-10% less than the threshold, at the threshold, or 0-10% greater than the threshold), the reactor is in _criticality_ and the status code of 'NORMAL' should be output, indicating that the reactor is in optimum condition and control rods are in an ideal position.
```

BethanyG (approved):
_Thank you!_ 🌟 💙

BethanyG:
Another thought would be to write it out as a sort of equation, but I don't know if that would make it worse or better:

If  `(threshold * .9) <= (temperature * neutrons_produced_per_second) <= (threshold * 1.1)`
the status code of 'NORMAL' should be output, indicating that the reactor is in optimum condition and control rods are in an ideal position.

...but that also gives a direct answer, so......  🤷🏽

iHiD:
> Another thought would be to write it out as a sort of equation, but I don't know if that would make it worse or better:

I wondered about a graphic like 
```
| ---- LOW ---- | --- NORM --- | --- HIGH --- |
| 0-90.              | 90-110         | 110-200 |
````

Or such with sane values and better indenting ;)

BethanyG:
🤔  _good thought_.  I could do a quick MD table or mermaid graphic.  I'll add it to my list for after I finish committing the tooling violence that's underway.  😉 .

PR #3517, opened 2023-10-18
Fix grammar in currency-exchange exercise

kytrinyx:
The word _amount_ is generally used for things that come in piles (like money) or
that can otherwise be measured but not counted (like water).

The word _number_ is typically used if you can count the things.

Technically money can be counted, but it's an exception (see 'piles' above).

Here we count bills and digits, so 'number' is the more common choice of word.


BethanyG (approved):
Thank you for this!  🌟 🚀

PR #3520, opened 2023-10-20
Clarify bills in currency exchange exercise

LittleMangoSeed:
The exchange booth in this exercise only deals in
cash of specific increments. That is to say, you can bring any combination of bills and coins to exchange, but you get a collection of bills of a single denomination in the target currency in return.

The explanation was quite clear on this, but the parameter names and docstrings in the methods dealing with this part of the task used the term 'budget', which could lead readers to assume that the bills had to do with the original currency.

This reworks the methods related to bills to be more generic, not referring to the exchange process at all, hopefully avoiding this misinterpretation.

See forum thread here for details of the discussion leading to this change:

https://forum.exercism.org/t/currency-exchange-exercise-suggested-naming-tweaks/7790

### DIFF
--- a/exercises/concept/currency-exchange/.meta/exemplar.py
+++ b/exercises/concept/currency-exchange/.meta/exemplar.py
@@ -39,7 +39,7 @@ def get_number_of_bills(amount, denomination):
     :return: int - number of bills that can be obtained from the amount.
     """
 
-    return int(budget) // denomination
+    return int(amount) // denomination
 
 
 def get_leftover_of_bills(amount, denomination):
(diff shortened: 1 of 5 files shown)


exercises/concept/currency-exchange/.meta/exemplar.py
```diff
@@ -50,7 +50,7 @@ def get_leftover_of_bills(amount, denomination):
     :return: float - the amount that is "leftover", given the current denomination.
     """
 
-    return budget % denomination
+    return amount % denomination
```
LittleMangoSeed, line 53: Oh my. Yes, I completely missed these! Thanks for fixing them.

PR #3526, opened 2023-10-26
Change 'amount' to 'number' for countables

kytrinyx:
This is a follow-up to https://github.com/exercism/python/pull/3517

I've been meaning to do this for a while, and finally found the time to sit down and meticulously check all the occurrences of "[Aa]mount of" in the repository.

I've fixed the following occurrences:
- number of values
- number of digits
- number of function calls
- number of hours
- number of health points
- number of (intermediate) delivery points
- number of keyword arguments
- number of grains
- number of names

I have also verified that the following are correct:
- amount of reputation
- amount of work
- amount of functionality
- amount of memory
- amount of money
- amount of domestic currency
- amount of tree climbing
- amount of energy

There are also two things that we need to figure out:

**Currency Exchange concept exercise**
https://github.com/exercism/python/blob/358e406e5a7a3c6b37fabda9bd2a867d0200dd4a/exercises/concept/currency-exchange/.docs/hints.md?plain=1#L13

This says `get the amount of changes` -- do we mean "number of changes" or "amount of change"? I _think_ it should be amount of change, but I'd like a second opinion. If so, it might belong in a separate PR since it's a completely different problem.

**Plane Tickets concept exercise**
This is a WIP exercise, so it might not matter yet, but the exercise talks about _amount of seats_, and it should be _number of seats_.

### DIFF
--- a/concepts/function-arguments/about.md
+++ b/concepts/function-arguments/about.md
@@ -182,7 +182,7 @@ For instance, `*` is used for multiplication, it is used for unpacking, and it i
 Since a tuple can be iterated, `args` can be passed to any other function which takes an iterable.
 Although `*args` is commonly juxtaposed with `**kwargs`, it doesn't have to be.
 
-Following is an example of an arbitrary amount of values being passed to a function:
+Following is an example of an arbitrary number of values being passed to a function:
 
 ```python
 
@@ -196,7 +196,7 @@ Following is an example of an arbitrary amount of values being passed to a funct
 
 If `*args` follows one or more positional arguments, then `*args` will be what is left over after the positional arguments.
 
-Following is an example of an arbitrary amount of values being passed to a function after a positional argument:
+Following is an example of an arbitrary number of values being passed to a function after a positional argument:
 
 ```python
 
@@ -210,7 +210,7 @@ Following is an example of an arbitrary amount of values being passed to a funct
 
 If one or more default arguments are defined after `*args` they are separate from the `*args` values.
 
-To put it all together is an example of an arbitrary amount of values being passed to a function that also has a positional argument and a default argument:
+To put it all together is an example of an arbitrary number of values being passed to a function that also has a positional argument and a default argument:
 
 ```python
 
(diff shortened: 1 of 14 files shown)


BethanyG:
> Currency Exchange concept exercise
Line 13 in [358e406](https://github.com/exercism/python/commit/358e406e5a7a3c6b37fabda9bd2a867d0200dd4a)
> _- You can use the [subtraction operator][subtraction-operator] to get the amount of changes._
>
> This says get the amount of changes -- do we mean "number of changes" or "amount of change"? I think it should be amount of change, but I'd like a second opinion. If so, it might belong in a separate PR since it's a completely different problem.

🤦🏽‍♀️   I agree:  It is _**amount of change**_.  I might go so far as to say _**amount of change (e.g. amount of leftover currency**)_.  But that might be over-explaining.  😅 . I also think maybe it goes in a separate "correct inaccurate pluralizations"  PR - but it is totally up to you.  I do not, in any way want to increase your workload!

> Plane Tickets concept exercise
> This is a WIP exercise, so it might not matter yet, but the exercise talks about amount of seats, and it should be _number of seats_.

It matters, because I probably can't be relied upon to remember down the road.  😄   That exercise and the accompanying concept docs. could do with a thorough going-through, top-to-bottom, to be honest.  I _want_ to turn it live, but its been through so many revisions that I have lost perspective on weather or not it is hitting its goal and is making sense.

Happy for you to do a PR just for the `amount` vs `number` -- or if you want to, an "everything that should be adjusted for this exercise" PR.  Or you can leave it.  Whatever you have time/desire to do! 😄

BethanyG (approved):
Thank you so much for this PR!  🌟   `Currency Exchange` has been problematic (_we even tried replacing it with `Electric Bill` - which caused a different uproar_),  so improvements to it are very welcome.  And the verification for other usages of `amount` vs `number` in the repo is a relief - sad you found 14 files that needed changing, but glad it wasn't a lot more than that!

kytrinyx:
@BethanyG thanks so much for the extra context! I'll do the little "amount of change" PR now (I think that when people are in the context of that exercise, we don't need to say leftover currency; the only reason I was unsure was that I wasn't actually doing the exercise).

I'll put the plane tickets one on my list as something to look at. I might just do the "amount" vs "number" bit, as you suggested, but it's good to know what the status is so that I can dive deeper if I have the energy.

PR #3529, opened 2023-10-28
Fix grammar in currency-exchange hint

kytrinyx:
This is a tiny thing, and it probably wouldn't confuse people who are
actually doing the exercise.

I came across it because I was doing a full audit of instances where
we say 'amount of ...' since there were several instances where it
should have been 'number of ...'.

At first I thought this was supposed to be 'number of changes', but
then in the context of the exercise it turns out that it was a
different problem entirely :)


BethanyG (approved):
💙 🌟 🚀   _Thank you!_

PR #3534, opened 2023-11-01
pop-count: update metadata

ErikSchierboom:
This PR updates the metadata for the `pop-count` exercise.


BethanyG:
should the slug change for this too, or is that a different PR?

PR #3546, opened 2023-11-06
[Mecha Munch Management]: Modified Test Error Messages & Fixed Ordering Tests

BethanyG:
Given the [modifications to the runner](https://github.com/exercism/python-test-runner/pull/114) that remove the test data from the subtests headlines in the UI, the error messages for failed tests and subtests needed modification.  New error messages now include the function with the arguments it is called with, in addition to the actual result and the expected result.

Also discovered errors in the test data and methodology for two tests:

`test_send_to_store()` and `test_sort_entries()`

Because `dicts` are hashmaps, two will compare equal if they have the same keys and values **_even if those appear in a different order_**.  While insertion order is guaranteed for dictionaries in Python, that ordering is not used for equality tests. 😱 so these tests were always passing....

To fix this issue, the assert check now converts the input and expected `dicts` into `OrderedDicts` (_where order does matter for equality_).

Additionally, multiple result_data `dicts` were not actually in  sorted order, so that has also been corrected.

Sadly, this will mean a requeue/retest of student solutions for this exercise.

### DIFF
--- a/exercises/concept/mecha-munch-management/dict_methods_test.py
+++ b/exercises/concept/mecha-munch-management/dict_methods_test.py
@@ -1,5 +1,6 @@
 import unittest
 import pytest
+from collections import OrderedDict
 from dict_methods import (add_item,
                           read_notes,
                           update_recipes,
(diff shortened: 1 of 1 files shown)


iHiD (approved):
Not reviewing the code, but happy for you to merge this and student's solutions should auto get re-tested

PR #3547, opened 2023-11-09
CI: pin GitHub Actions runners

ErikSchierboom:
This PR updates GitHub Actions runners to a specific version.
This ensures that the workflow will always run on the same runner, which makes your build _stable_.

The PR updates the *-latest version with the current version, as specified in https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.

See https://exercism.org/docs/building/github/gha-best-practices#h-pin-test-runners-to-version for more information.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   housekeeping:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


BethanyG (approved):
Thanks for this!

PR #3564, opened 2023-12-08
The `bitwise-operators` concept

colinleach:
Another concept, to review non-urgently at your leisure.

There was some minor confusion about calling it `bitwise-operators` (better) or `bitwise-operations` (occurs in a few places in the plans). That's why the concept uses one, the branch uses the other name.

### DIFF
--- a/concepts/bitwise-operators/.meta/config.json
+++ b/concepts/bitwise-operators/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb for this concept",
-  "authors": ["bethanyg", "cmccandless"],
+  "blurb": "Python supports bitwise operations such as left/right shift, and, or, xor.",
+  "authors": ["BethanyG", "colinleach"],
   "contributors": []
 }
--- a/concepts/bitwise-operators/.meta/config.json
+++ b/concepts/bitwise-operators/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "Python supports bitwise operations such as left/right shift, and, or, xor.",
+  "blurb": "Python supports bitwise operations such as left/right shift, and, or, xor, and not.",
   "authors": ["BethanyG", "colinleach"],
   "contributors": []
 }
--- a/concepts/bitwise-operators/about.md
+++ b/concepts/bitwise-operators/about.md
@@ -133,8 +133,11 @@ This makes it hard to give a useful example of bitwise not:
 
 This is _not_ the `0b10011001` we would see in languages with fixed-size integers.
 
+The `~` operator works as expected with `unsigned` byte or integer types.
+These are supported in third-party packages such as `NumPy`, but not in core Python.
+
 In practice, Python programmers quite often use shift operators and `& | ^` with positive numbers.
-Bitlise operations with negative numbers are much less common.
+Bitwise operations with negative numbers are much less common.
 
 
 
--- a/concepts/bitwise-operators/about.md
+++ b/concepts/bitwise-operators/about.md
@@ -1,144 +1,197 @@
 # About
 
-Down at the hardware level, transistors can only be on or off: two states that we traditionaly represent with `1` and `0`. 
-These are `binary digits`, abbreviated as `bits`.
+Down at the hardware level, transistors can only be on or off: two states that we traditionally represent with `1` and `0`.
+These are the [`binary digits`][binary-digits], abbreviated as [`bits`][bits].
+Awareness of `bits` and `binary` is particularly important for systems programmers working in low-level languages.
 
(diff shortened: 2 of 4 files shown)


BethanyG:
We also have [this concept](https://github.com/exercism/python/blob/main/concepts/binary-octal-hexadecimal/about.md) we can refer/link to, but as I read through what you've written, perhaps it needs a go-over/edit, so that the two work together?  What do you think?

And we also have this [pending concept](https://github.com/exercism/python/tree/main/concepts/bitflags), should we combine them, or keep it by itself?  

[Real Python](https://realpython.com/python-bitwise-operators/#bitmasks) does it all in a fell swoop, but I ... _sorta_ feel like the article is TMI.  There is a lot to explore with bitmasks....Thoughts?

colinleach:
Ummm, OK - I had completely failed to notice the existing concepts, so obviously didn't take them into account. It's probably best if I think about this tomorrow and get back to you, instead of rushing it now. We certainly need to make everything fit together!

This is a foretaste of what we'll be dealing with on the `functions` sub-tree, which has even more of what patent lawyers call "prior art" to work around.

BethanyG:
Ok.  That took considerably longer than I wanted to.  I've pushed a first pass, but will wait for you to look things over and re-work.  I am not attached (particularly) to this edit, so if you need to, you can revert it.  😄 

Onward to `collections` and `datetime`!!  🚀

colinleach:
That's a big improvement on my draft. I've no big changes to suggest, just some small details:
- Line 46, the Bitwise Logic sub-header has a broken link. We need to add a reference [bitwise-operators].
- Line 63 refers to "both the input strings". Technically, these are (binary) ints/numbers, not strings. Worth changing?
- Line 92, broken link to [concept:python/sets]. Is this a GitHub thing that fixes itself on the Exercism website, or do we need to fix it?
- Line 116, the most minor of niggles. I would call this "most significant bit" or MSB. By all means stick with "significant bit" if that's what you prefer.

BethanyG:
Those edits make total sense.  😄   I will investigate the widget link pronto.  

Any further thoughts on https://github.com/exercism/python/pull/3564#issuecomment-1869878450?  Should I hold off on any edits pending your reviews/thoughts?  I can work on the other concepts and exercises meanwhile. 😄

BethanyG:
> That's a big improvement on my draft. I've no big changes to suggest, just some small details:
> 
> * Line 46, the Bitwise Logic sub-header has a broken link. We need to add a reference [bitwise-operators].

Thank you for the catch!  ✅ 

> * Line 63 refers to "both the input strings". Technically, these are (binary) ints/numbers, not strings. Worth changing? 

_worth changing!_ ✅ 
> * Line 92, broken link to [concept:python/sets]. Is this a GitHub thing that fixes itself on the Exercism website, or do we need to fix it?

Good eyes here!  The link was missing a set of `()`.  I added them in, and now it should work.  ✅ 🤞🏽 
> * Line 116, the most minor of niggles. I would call this "most significant bit" or MSB. By all means stick with "significant bit" if that's what you prefer.

Not a minor niggle at all!  If it is MSB, then we should use MSB.  I spelled it out, then used the abbreviation later. LMK if it looks OK.  ✅ 

_Just pushed a new version with those changes ++ other things I saw on a re-read._

colinleach:
I like your latest commit. That just leaves the question of how to integrate with the `binary-octal-hexadecimal` concept. 

Can we answer that question now? As both lack concept exercises (and there is nothing good to fork from - crystal/secrets is pretty lame), we may need to revisit this once we have something to deploy on the learning syllabus. Or am I just evading the question?

BethanyG:
You are evading, but that's fine (_and totally understandable!_).

I am pretty happy with this write-up, and I don't _hate_ the `binary-octal-hexadecimal` concept.  It needs a go-over, but it can be done when we find an exercise idea to implement.  I had been working on one about an explosion in a paint-by-numbers factory (_putting different color sets back together_) or some other colorspace or image conversion problem where you need to use/convert between the three numberspaces.  But didn't get too far. 

So maybe we merge this and work on the `bitflags` concept?  And focus on trying to dream up good exercises for both `bitwise` and `bitflag`.

One thought is a Unicode version of [this exercise](https://github.com/exercism/problem-specifications/blob/main/exercises/micro-blog/description.md), where students would need to track MSB in order to keep diacritical marks together with the user-characters they modify, in order to properly "truncate" text?  

Or we could implement a Unicode version of reverse-string, with a similar issue.  Though that could be really complicated, especially for triple-byte character sets....

BethanyG (approved):
I am going to bite the bullet and merge this for now.  We can revisit the other issues as we move along.  Thank you again for all your hard work on this!  🌟 🌈

colinleach:
Yay!

PR #3565, opened 2023-12-08
The `secrets` concept

colinleach:
A teeny, tiny concept. I wish they were all this quick to put together.

### DIFF
--- a/concepts/secrets/.meta/config.json
+++ b/concepts/secrets/.meta/config.json
@@ -1,0 +1,5 @@
+{
+  "blurb": "The secrets module is a cryptographically-secure alternative to the random module, intended for security-critical uses.",
+  "authors": ["BethanyG", "colinleach"],
+  "contributors": []
+}
--- a/concepts/secrets/about.md
+++ b/concepts/secrets/about.md
@@ -1,0 +1,37 @@
+# About
+
+A previous concept discussed the `random` module, which produces random numbers and random list orderings.
+
+The `secrets` module apparently overlaps with `random` in some of its functionality, but the two modules are designed with very different priorities.
+- `random` is optimized for high perfomance in modelling and simulation uses.
+- `secrets` is optimized to be crytographically secure, for security-critical applications.
+
+Further details on why `secrets` proved necessary are given in [PEP 506][PEP506]
+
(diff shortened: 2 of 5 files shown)


BethanyG:
[random vs pseudo-random](https://www.khanacademy.org/computing/computer-science/cryptography/crypt/v/random-vs-pseudorandom-number-generators)  too much?  Or would this be a good link?

BethanyG:
That being said, [wikipedia](https://en.wikipedia.org/wiki/Pseudorandomness) has an easily scannable and understandable text version as well.

colinleach:
The Wikipedia entry certainly looks worth including. Maybe Khan Academy as well?

To do it properly, there is Knuth, Chapter 3. We might get into trouble for adding that.

BethanyG:
Alrighty.  I did the edits via [GitHub Desktop](https://desktop.github.com/) this time, rather than the web interface.  Seemed quicker.  But I am also happy to do it via the web interface, if you prefer.  I'll probably do that for longer ones, or in cases where there is discussion to be had/lots of suggestions to be made.

If all this looks good, I can merge this down.  Just let me know. 😄

BethanyG:
> _To do it properly, there is Knuth, Chapter 3. We might get into trouble for adding that._

🤣 

Do. Not. Tempt me!  .... _oooh, I so want to ... but the blowback t'would be **fierce**_

BethanyG:
**BTW**:  This `[concept:python/random]()` formulation is a [widget](https://github.com/exercism/docs/blob/main/building/markdown/widgets.md), used for when we refer back to a concept or exercise.  I don't remember to use them, but we should do so as we build out the tree more.

colinleach:
Looks good to me, merge when you're ready.

I'd been meaning to get my head around Exercism widgets, so this is a timely prompt.

colinleach:
I've not used GitHub Desktop recently. Mostly, I use GitKraken (on Linux and Windows), though not to anywhere near its full potential. I really must read the documentation sometime (after all, I'm paying for the Pro version).

BethanyG (approved):
Woot!  One more off the list!  💥 🌟 🚀 

Thank you for this.  😄

colinleach:
To repeat: I wish it was always this easy!

PR #3571, opened 2023-12-26
`captains-log` concept exercise

colinleach:
A possibility for the `random` concept? Please rate as OK/fixable/junk.

The tests are unusual, to deal with non-deterministic output. Please sanity-check them, both for accuracy and test-runner performance. I tried to pick a number of repeats that will give good coverage within a reasonable time, but this may need some fine-tuning.

### DIFF
--- a/config.json
+++ b/config.json
@@ -226,6 +226,18 @@
           "loops"
         ],
         "status": "wip"
+      },
+      {
+        "slug": "captains-log",
+        "name": "Captains Log",
+        "uuid": "c781da14-40c0-4aeb-8452-769043592775",
+        "concepts": ["random"],
+        "prerequisites": [
+          "numbers",
+          "lists",
+          "string-formatting"
+        ],
+        "status": "wip"
       }
     ],
     "practice": [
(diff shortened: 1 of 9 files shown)


colinleach:
Yes, I admit that's a better title.

Meanwhile, I tried estimating the chances that task 1 would randomly fail by excluding one of the possible planetary_classes from the results. I think the chance of a particular entry not being picked is 0.9 in each run (there are 10 to choose from). With 1000 runs, the chances of missing one is `0.9 ** 1000`, about `1.7e-46`. Good enough?

BethanyG:
This has the danger of pulling me into a black hole around how to test random processes.  😉 
But better we think through this now, since this won't be the only exercise with randomness in it.

> _chances of missing one is `0.9 ** 1000`, about `1.7e-46`. Good enough?_

Yes.  That _sounds like_ good enough, but I'll probably do more thinking on it as I read through.  First, I want to tackle some of the concept write-ups, so we can merge them down and cross them off our list.  But this is certainly near the top of the TODO.  😄

PR #3589, opened 2024-01-13
Fixed contradictory tests.

bobbypinard:
self.assertEqual(score(-9, 9), 0),
self.assertEqual(score(0.8, -0.8), 5),
self.assertEqual(score(-3.6, -3.6), 1), and
self.assertEqual(score(7.1, -7.1), 0) contradict the rules of the exercise.

### DIFF
--- a/exercises/practice/darts/darts_test.py
+++ b/exercises/practice/darts/darts_test.py
@@ -11,7 +11,7 @@ from darts import (
 
 class DartsTest(unittest.TestCase):
     def test_missed_target(self):
-        self.assertEqual(score(-9, 9), 0)
+        self.assertEqual(score(-9, 9), 1)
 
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @bobbypinard  Thanks for taking the time to submit a PR.

The tests for this exercise on the Python track are pulled from a [common repository](https://github.com/exercism/problem-specifications/blob/main/exercises/darts/canonical-data.json) for ALL language tracks, and have been reviewed many many times across our 67 different programming languages.  It's highly unlikely that these tests and  their expected results are incorrect.

 However, if you would like to discuss the test data, you should do so in the **problem-specifications repo**, which is the place the test data and results are kept for cross-language problems such as these.  The tests on the Python track are auto-generated from problem specs.

PR #3597, opened 2024-01-16
[Square Root]:  Add that Solutions are Natural Numbers

### LINKED ISSUE #3594: Clarify that the root is a natural number in square-root documentation
The instructions in `exercises/practice/square-root/.docs/instructions.append.md` says that they are "only testing natural numbers," but more clearly, the *solutions* are only natural numbers.  Two is a natural number and its root is real (and irrational).

ksnortum:
Clarifies that the solutions are natural numbers.  Fixes #3594


BethanyG (approved):
Nice work!  Many thanks for this.  🌟 🌈

PR #3606, opened 2024-01-26
Update instructions.md

Nerwosolek:
Two instructions missed additional parameter in function declaration. Added.


BethanyG (approved):
Thanks for this!

PR #3615, opened 2024-02-02
`pop-count`: rename to `eliuds-eggs`

ErikSchierboom:
_We are renaming the `pop-count` exercise to `eliuds-eggs`. See https://github.com/exercism/problem-specifications/pull/2382 for more information._

---

This PR renames `pop-count` to `eliuds-eggs`. We have attempted to rename everything that is automatable, but you may need to make specific changes in the files.

If you could prioritise finishing and merging this, we would be grateful, so we can get the exercise in the correct state across Exercism. Please let me know if you have any questions.

Thanks! 🙂


BethanyG:
The exercise config, test file, and stub also needed renaming.  Hopefully, that won't cause too much chaos with retesting and solution invalidation.

Because the test class in the test file is named for the exercise, the test file also needed to be regenerated.

PR #3619, opened 2024-02-04
Mecha Munch Management - Fix misspelling of 'aisle' as 'isle'

ryanplusplus:


ryanplusplus:
This changes important files and will invalidate existing solutions, but it does so to correct an error that I found confusing when solving the exercise.

BethanyG (approved):
Thanks for this.  😄   I caught the ones in the instructions, but didn't even **think** about the ones in the exemplar and test file!  🤦🏽‍♀️

PR #3620, opened 2024-02-05
dict-and-join simplify for raindrops approaches

Nerwosolek:
Probably this was done on purpose but as a way to discuss this I wanted to propose a little simplification to the approach.
I think this is more sound and readable to just use the item from the dict.
Let me know what you think.
Thanks in advance.

### DIFF
--- a/exercises/practice/raindrops/.approaches/dict-and-join/content.md
+++ b/exercises/practice/raindrops/.approaches/dict-and-join/content.md
@@ -5,16 +5,16 @@
 def convert(number):
     sounds = {3: 'Pling', 5: 'Plang', 7: 'Plong'}
 
-    results = ''.join(sound for 
-                      divisor, sound in sounds.items()
+    results = ''.join(sounds[divisor] for 
+                      divisor in sounds.keys()
                       if number % divisor == 0)
 
     return results or str(number)
 ```
 
 This approach uses a [dictionary][dict] called 'sounds' with factors as `keys` and sound strings as `values`.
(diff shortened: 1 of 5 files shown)


BethanyG:
Hi there!  👋🏽 

Thanks for submitting this PR.  😄 

> Probably this was done on purpose but as a way to discuss this I wanted to propose a little simplification to the approach.
I think this is more sound and readable to just use the item from the dict.

I did do it on purpose, but my 'copypasta' got the better of me!  It is more sound when using `dict.items()` to have the dictionary view unpack both the keys and the values (_and then use them!_).  And that was my intent here -- to show an approach that was different looking, but the _same_ as what is in the `introduction.md` summary:

```python
def convert(number):
    #This is clear and easily added to.  Unless the factors get
    # really long, this won't take up too much memory.
    sounds = {3: 'Pling',
              5: 'Plang', 
              7: 'Plong'}

    results = (sounds[divisor] for 
               divisor in sounds.keys()
               if number % divisor == 0)

    return ''.join(results) or str(number)
```

Buuut.  I copied the code above and missed modifying `sounds[divisor]`!  😱  

Now that I am looking at all this again, I think perhaps the best thing to do is use the example above in the approach document, and use the code with `dict.items()` in the `introduction.md`.  I'll update this PR with those changes, and you can let me know if it reads better to you that way.

Nerwosolek:
Yes, looks even better. The distinguish between using items() and keys() adds even more knowledge to the already rich article!

BethanyG (approved):
Thanks for giving this one last read-through!  🌈

BethanyG:
> Yes, looks even better. The distinguish between using items() and keys() adds even more knowledge to the already rich article!

Going to push the big, green button then!

PR #3633, opened 2024-02-20
Update introduction.md

oezg:
keys(), values(), items() can be recalculated. Assigning to a variable beforehand makes it the dynamic attribute of the views clear. 
This is my first contribution and very first pull request. Therefore I am not sure if I am doing the pull request as it should.


BethanyG:
Hi @oezg 👋🏽 

Thank you for taking the time to submit a PR.  Unfortunately, I think assigning a variable to `palette_I.keys()` confuses things here.  Especially with the section that follows.  I would rather leave the introduction as it stands for now.

PR #3634, opened 2024-02-20
Update introduction.md reference link name wrong

oezg:
The link is from W3 Schools instead of Real Python


BethanyG:
Hi @oezg 👋🏽 

Thank you for submitting this PR.  The link was intended to be Real Python, so I will correct it.

PR #3640, opened 2024-02-25
[Circular Buffer] draft approaches

colinleach:
For this one I experimented with brevity to see how it looks.

What do you think - too short? Or about what this exercise deserves (in Python)?

### DIFF
--- a/exercises/practice/circular-buffer/.approaches/built-in-types/content.md
+++ b/exercises/practice/circular-buffer/.approaches/built-in-types/content.md
@@ -1,0 +1,88 @@
+# Built In Types
+
+```python
+class CircularBuffer:
+    def __init__(self, capacity: int) -> None:
+        self.capacity = capacity
+        self.content = []
+
+    def read(self) -> str:
+        if not self.content:
+            raise BufferEmptyException("Circular buffer is empty")
+        return self.content.pop(0)
+
+    def write(self, data: str) -> None:
+        if len(self.content) == self.capacity:
+            raise BufferFullException("Circular buffer is full")
+        self.content.append(data)
+
+    def overwrite(self, data: str) -> None:
+        if len(self.content) == self.capacity:
+            self.content.pop(0)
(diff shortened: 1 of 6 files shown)


colinleach:
Oops - I just found a blank section that I had intended to fill in later, then forgot about.

BethanyG:
Overall, I liked this, and for some exercises, I do think being briefer is a good thing.  

I did go ahead and edit/add some things - I wanted to add more, but I think (given our timeframe) that this is going to have to do.

I really wanted to implement a `memoryview` approach, and one using either a `ctype` or `struct` -- but ended up not having the wherewithal to do so.  I stuck with the simplest use of `array.array` -- which is essentially a mimic of using `bytearray` or `list` to underpin.

Maybe we can loop back to it, because it is really interesting to dig down underneath the sugar and poke at the C API.  That being said, dig too far, and it becomes a shadow C track.  _Who says Python is only high-level?_  😉

BethanyG (approved):
🚀 🌟   I am going to go ahead and merge this.  Please do give it a read, and if there are any issues, I can PR something quickly to fix.  But I think given the delays I caused, we should get this live to the site.  😄   Thank you so much for your hard work on this!

colinleach:
Only  issue I can see is a bit of grammar
>Approach: Using built-in types
>Python has an exceptionally flexible and widely-used list type. Most submitted solutions to Circular Buffer are based on this data type.

>**A less versatile variants include** [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray) and [array.array](https://docs.python.org/3.11/library/array.html#module-array)

"A less versatile variants include' is sign of edits gone a bit wrong. Intoduction.md, line 41

BethanyG:
Will submit a fix here shortly, unless you want to?

colinleach:
I'm a bit distracted by other stuff this morning. You could probably do it quicker and more reliably.

PR #3652, opened 2024-02-29
[All Your Base] draft approaches doc

colinleach:
Once I got into this, it became clear that there's only one approach with minor syntax variations. Hence, there is just an `introduction.md`: no separate approaches, no benchmarking.

Are you OK with this?


BethanyG:
> Are you OK with this?

Not really.  But this has drug on long enough and I am tired enough that I am going to approve this one.  Let's not do more like this going forward unless we have a really good reason to.

I made a few changes - can you give it a read to make sure I didn't do anything crazy?  _Many thanks!_.  😄 💙

PR #3660, opened 2024-03-23
[Black Jack]: ✏️ Updates to instructions.

fefulowe:
Previously discussed in the [forum](https://forum.exercism.org/t/suggested-changes-to-python-learning-exercise-black-jack-instructions/10436).

The instructions on the exercise [black_jack](https://exercism.org/tracks/python/exercises/black-jack) might be improved.
It could be better to refer to the value of the upcoming ace instead of the value of those in hand since this doesn't apply if we have more than one ace in hand.

Current instructions:
 `Hint: if we already have an ace in hand then its value would be 11.`

Proposed instructions:
 `Hint: if we already have an ace in hand, then the value for the upcoming ace would be 1.`

Sources: "[...] and aces count as either 1 or 11 according to the player's choice." from [Wikipedia](https://en.wikipedia.org/wiki/Blackjack?lang=en#Rules_of_play_at_casinos).


BethanyG (approved):
Many thanks for this!  🌈

PR #3667, opened 2024-04-01
Fix maximum length of a roman numeral

mpg:
As discussed on the forum: http://forum.exercism.org/t/48in24-exercise-02-06-roman-numerals/9399/23


colinleach:
Thanks for that. It reminds me, yet again, of the dangers of repeating something I read on the internet!

I don't have write access to this repo, it will need @BethanyG to approve and merge.

PR #3668, opened 2024-04-01
Added Anaconda link to Installation

Silverasdf:
Hello,

As I use Anaconda a lot for my Data Science projects, I wanted to share the link to Anaconda so that new programmers may have a lot of useful features out of the box.

Thanks,

Ryan.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -3,6 +3,8 @@
 If Python isn't already available on your system, detailed instructions by platform can be found in the [Python Setup and Usage][using python] section of the official documentation.
 Real Python also offers a [nice guide][helpful guide] to installation on various platforms, including iOS and Android.
 
+For aspiring Data Scientists, using [Anaconda][Anaconda-Link] may also prove useful, although this is not an official distribution of Python. Anaconda has many useful features that work out of the box, such as Jupyter Notebooks and Virtual Environment support.
+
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @Silverasdf 👋🏽 

Thank you for taking the time to submit a PR to the Python repo. 😄 

The main distribution of Python that I use on a daily basis is Anaconda, and I often recommend it to those who are learning data engineering, data science, or machine learning -- and to people who need to use multiple versions of Python at the same time.  Unfortunately, we can't recommend Anaconda to Exercism students for multiple reasons:

1. We aim to teach Python language fluency (_and general programming_) to students.  Data Science libraries (_while really important_) are not general-purpose programming, and not part of the core Python language. 
2. None of the libraries that are part of the Anaconda distribution (_except for the different versions of Python itself_) are in use on the Exercism platform.  You can't use NumPy in an Exercism solution and have tests pass -- our tooling doesn't install third-party libraries.  This means that students can't learn to use those libraries online here (_at least not in the near future_).
3. The standard install of Anaconda (_not miniconda_) is huge -- gigs of stuff, and lots of time to download on a slower internet connection.  Since almost all of that download is not useful for Exercism, it would be problematic to recommend it to newer students.
4. Most installations of Anaconda on most platforms change the Python PATH, install shims and pointers, add special Anaconda shells, and generally rewire things.  This can cause real problems for students who are unfamiliar with how those things work.

So I am sorry - but I can't accept this PR at this time.

PR #3702, opened 2024-05-21
[Ghost Gobble arcade game]fix semantics.

demeesterdev:
Small semantic fix within the first task of the exercise.

---

The requested functionality isn't to to return if you can eat **the** ghost but given the input if you can eat **a** ghost


BethanyG (approved):
Thanks for this!

PR #3706, opened 2024-05-25
Queen Attack - Improve clarity of errors

ryanplusplus:
The errors for negative rows and columns states that row/columns must be positive, but zero is a valid value and zero is not positive. If we change the language from "row/column not positive" to "row/column is negative" then the error will correctly describe the error case.


ryanplusplus:
Whoops, looks like this is from the upstream problem specifications. I'll raise the issue there instead.

PR #3753, opened 2024-08-13
[Collatz Conjecture] Approaches - missing words fix

ebogucka:
The General guidance text was saying: "The key to this exercise is to check if the number and then do the correct operation." I have changed it to say that we check if the number is even.

### DIFF
--- a/exercises/practice/collatz-conjecture/.approaches/introduction.md
+++ b/exercises/practice/collatz-conjecture/.approaches/introduction.md
@@ -6,7 +6,7 @@ You can also solve it by using if and else statements or the ternary operator.
 
 ## General guidance
 
-The key to this exercise is to check if the number and then do the correct operation.
+The key to this exercise is to check if the number is even and then do the correct operation.
 Under this process you are supposed to count how many steps it takes to get to one.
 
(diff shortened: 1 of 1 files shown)


exercises/practice/collatz-conjecture/.approaches/introduction.md
```diff
@@ -6,7 +6,7 @@ You can also solve it by using if and else statements or the ternary operator.
 
 ## General guidance
 
-The key to this exercise is to check if the number and then do the correct operation.
+The key to this exercise is to check if the number is even and then do the correct operation.
```
BethanyG, line 9:
```suggestion
The key to this exercise is to check if the number is even or odd and then perform the correct operation.
```

BethanyG (approved):
Thanks for catching this and submitting a PR!  I've slightly re-worded, just so we capture that the operations are different with even vs odd.  🙂

BethanyG:
Hi @ebogucka 👋🏽 

We've currently got an issue with CI failing.  I am looking into it, and as soon as it is resolved, I can merge your PR.  😄

PR #3762, opened 2024-08-26
Making the test run, when the script is run.

Ornataweaver:
Without `unitest.main()`, the tests wouldn't run.

Please consider explaining to the student why the `if` statement is necessary, and why it checks if the name is `__main__`.
9 years ago I didn't understand this, I just found out about [modules](https://docs.python.org/3/tutorial/modules.html), and I'm very excited to share this.

Thank you. 🍇

### DIFF
--- a/exercises/practice/hello-world/hello_world_test.py
+++ b/exercises/practice/hello-world/hello_world_test.py
@@ -28,3 +28,12 @@ class HelloWorldTest(unittest.TestCase):
     def test_say_hi(self):
         msg = "\n\nThis test expects a return of the string 'Hello, World!' \nDid you use print('Hello, World!') by mistake?"
         self.assertEqual(hello(), "Hello, World!", msg=msg)
+
+if __name__ == '__main__':
+    # Why is this if implanted?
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @Ornataweaver 👋🏽 

As the comments on your [forum post](https://forum.exercism.org/t/making-the-test-run-when-the-script-is-run/12714/5) imply, this was an on-purpose decision.  

We used to use `if __name__ == __main__` in our test files, but since the decision was made to use [`pytest`](https://docs.pytest.org/en/8.0.x/getting-started.html) for our online test runner and `pytest` syntax for our learning exercises, we've [pointed students to](https://exercism.org/docs/tracks/python/tests) installing and using `pytest` for testing.  We retain the `unittest` syntax for now in test files, but might move to all-pytest syntax in the future.

If students so choose, they can add that block back to tests (_although it won't work for learning exercises_) themselves. but we don't support/encourage it, since it is a poorly understood feature for beginners to use and we do not yet have any learning exercises around it.

Ornataweaver:
Thank you so much for putting the time and care. @BethanyG 🙏🏻

PR #3772, opened 2024-09-12
Create intoduction for dig deeper in gigasecond exercise

SNEAXIII:


BethanyG (changes requested):
Hi @SNEAXIII 👋🏽 

Thanks for submitting this approaches PR.  While I appreciate the effort, the introduction as it currently stands is not really in line with the other exercise approaches introduction documents.  I think there needs to be some more information (e.g. the general approach and key to solving the problem), and also more code/syntax examples to show how different students coded up the solution.  Even with "one general approach", there can still be a variety of strategies.

For example, declaring a module-level [gigasecond timedelta constant](https://exercism.org/tracks/python/exercises/gigasecond/solutions/vsoto1234) for use in the function, rather than [re-creating the object](https://exercism.org/tracks/python/exercises/gigasecond/solutions/t3j11) for every function call.   Or declaring [gigasecond as a constant number](https://exercism.org/tracks/python/exercises/gigasecond/solutions/BNAndras). 

There can be difference in how the parameters to `timedelta` are given ([named](https://exercism.org/tracks/python/exercises/gigasecond/solutions/simisc) vs [positional](https://exercism.org/tracks/python/exercises/gigasecond/solutions/archrisV) ), and diversity in how a `gigasecond` itself is declared, with different students using `pow()` or `**`, or writing the number out, or using scientific notation.

I know that this might feel like we're drawing distinctions that aren't there, or adding detail that's not interesting.  However, this is _dig deeper_, and we should be diving into as many ways (_and tradeoffs for them_) to solve the problem as possible. For someone new to Python, all of these details help them become familiar with details of syntax and conventions used for the language.  There can also be subtle performance tradeoffs between seemingly identical solutions (_like the module-level object/constant vs the in-function object_).

Please take a look at the approaches for [leap](https://github.com/exercism/python/blob/main/exercises/practice/leap/.approaches/introduction.md), [grains](https://github.com/exercism/python/blob/main/exercises/practice/grains/.approaches/introduction.md), [acronym]( https://github.com/exercism/python/blob/main/exercises/practice/acronym/.approaches/introduction.md), and [pythagorean-triplet](https://github.com/exercism/python/blob/main/exercises/practice/pythagorean-triplet/.approaches/introduction.md) for some examples of how we've set up the documents, and gone into detail.

Should you want more/different feedback from other maintainers, I'd encourage you to make a [forum post](forum.exercism.org) post in the Python category.

SNEAXIII:
Hi @BethanyG,
I am new here and, as you see, I don't really know what I have to put in the introduction.

Thanks for the detailed answer. Of course, you are right, and I will make the change. I was afraid of flooding people with explanations.

SNEAXIII

SNEAXIII:
Hi :),
Here is my second iteration. I am sure I miss something, but I don't know what...

SNEAXIII:
Hi @BethanyG,
Did you see my previous commit ?
Have a great day

BethanyG:
Hi @SNEAXIII,

Thanks for the bump.  I've been pretty busy, and haven't had a chance to review your updates yet.  I should be able to get to them by this weekend, and will let you know if I get to them earlier.  Thanks for your patience.  🙂

BethanyG:
Closing this as we are going in a different direction for these approaches.

PR #3779, opened 2024-10-02
Fixed Typo

Tousend1000:
Fixed Typo on line 6 and 7.
ellement -> element


kotp:
I have been inactive enough on this track to no longer have the ability to reopen the issue, but this is valid, and I would approve it if I could.

You can draw attention to this by opening a post on the forum, as directed in the automated response, but also please link to this issue there and also link back to the post here, so that there is cohesion to the communication channels.

PR #3800, opened 2024-10-28
[Acronym]:  Incorrect Reference to Isogram Performance Article

Ryrden:
For some reason, the performance article was pointing to another problem, this got me confused when I was studying the `dig deeper`.


BethanyG (approved):
Many thanks for this! 🎉

PR #3812, opened 2024-11-04
Update instructions.append.md

xanni:
Clarify that the built-in exponentiation operator ** is equivalent to pow() and therefore doesn't demonstrate an implementation of a square root algorithm, and also note that for the domain of this exercise there are available solutions using only positive integers.

### DIFF
--- a/exercises/practice/square-root/.docs/instructions.append.md
+++ b/exercises/practice/square-root/.docs/instructions.append.md
@@ -3,10 +3,10 @@
 ## How this Exercise is Structured in Python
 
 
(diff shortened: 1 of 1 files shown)


BethanyG (approved):
🎉   Thank you for this.  I made one suggestion.  We try to to format our markdown documents so that there is one sentence per line.  They are easier to maintain that way, and the markdown engine concatenates them for us.

Other than that, this looks great, so I approved.  I'll wait on your change/acceptance to merge it.  😄

PR #3815, opened 2024-11-05
Move generic instructions into the correct section.

xanni:


BethanyG:
@xanni  👋🏽 

Thanks for the second improvement.  😄 

Because this change touches `instructions.md` rather than just `instructions.append.md`, it will have to be proposed on the forum as a change to **`problem-specificiations`**.  I'm fine with changing the `instructions.append.md`, but we need agreement of at least 3 maintainers before we can make changes to the cross-track `instructions.md`.

And you will have to do the `instructions.md` part of the change in the [`problem specifications`](https://github.com/exercism/problem-specifications/blob/main/exercises/square-root/description.md) repo.

PR #3819, opened 2024-11-07
[Chaitana's Colossal Coaster] / [List Methods]: Change instructions to not return a copy

leonqadirie:
While returning a copy (not altering the original) makes more sense in the provided setting, this is:
* not tested.
* encourages just using the `sorted()` function instead of the `copy()` and `sort()` methods I suspect we want to see in this exercise - as it introduces the list method concept..

If we want to test for the returned value being a copy, keeping the original intact, this might be the corrected test:

```python
@pytest.mark.task(taskno=7)
    def test_sorted_names(self):
        test_data = ["Steve", "Ultron", "Natasha", "Rocket"]
        test_data_copy = test_data.copy()
        expected = ["Natasha", "Rocket", "Steve", "Ultron"]
        actual_result = sorted_names(test_data)

        error_message = (
            f"Called sorted_names({test_data})."
            f"The function returned {actual_result}, but the tests "
            f"expected {expected}."
        )

        self.assertListEqual(actual_result, expected, msg=error_message)
        self.assertListEqual(
            test_data,
            test_data_copy,
            msg="Function should not modify the original list",
        )
```

### DIFF
--- a/exercises/concept/chaitanas-colossal-coaster/.docs/instructions.md
+++ b/exercises/concept/chaitanas-colossal-coaster/.docs/instructions.md
@@ -134,7 +134,7 @@ You should update the `list` and also `return` the name of the person who was re
 For administrative purposes, you need to get all the names in a given queue in alphabetical order.
 
 
-Define the `sorted_names()` function that takes 1 argument,  `queue`, (the `list` of people standing in the queue), and returns a `sorted` copy of the `list`.
+Define the `sorted_names()` function that takes 1 argument,  `queue`, (the `list` of people standing in the queue), and returns the `sorted` `list`.
 
 
 ```python


leonqadirie:
[Forum discussion here](https://forum.exercism.org/t/chaitanas-colossal-coaster-misleading-expectations/10798)

BethanyG:
Hi @leonqadirie  👋🏽 

Thanks for submitting this.  If students want to use `sorted()`, and understand what it does (_returns a copy of said iterable_), they are free to do so.  It has the exact same effect as manually making a copy and then mutating that copy.  Either is fine.  So I think we leave the instructions as-is.

To your second point -- yes, we probably should be checking to see if the list was mutated.  I will have a think on that one.

leonqadirie:
> Hi @leonqadirie 👋🏽
> 
> Thanks for submitting this. If students want to use `sorted()`, and understand what it does (_returns a copy of said iterable_), they are free to do so. It has the exact same effect as manually making a copy and then mutating that copy. Either is fine. So I think we leave the instructions as-is.
> 
> To your second point -- yes, we probably should be checking to see if the list was mutated. I will have a think on that one.

Yes, I didn't want to suggest prohibiting `sorted()` or anything alike, my suggestion was meant to focus on being explicit w.r.t. mutation (a topic somewhat dear to my heart). So either:
* not explicitly asking for a copy, or
* if explicitly asking for a copy, ensuring a copy is provided via tests

Thanks for consideration :)

PR #3821, opened 2024-11-15
Update instructions.md

chivalry:
Perhaps it was a typo but 4 is not the average of the list, nor the average of the first and last, while 5 is. So either the middle item should be 5 or the result should be `False`.


IsaacG:
[Forum discussion here](https://forum.exercism.org/t/13807).

PR #3825, opened 2024-11-18
[List Methods/Chaitana's Colossal Coaster]: Added Queue Validation Tests & Instructions Note

BethanyG:
Per [forum discussion](https://forum.exercism.org/t/chaitanas-colossal-coaster-misleading-expectations/10798/9).

Not exactly a fan of all the queue checks, but we are going to run a little experiment to see if it increases or decreases complains and or completions.

### DIFF
--- a/exercises/concept/chaitanas-colossal-coaster/.docs/instructions.md
+++ b/exercises/concept/chaitanas-colossal-coaster/.docs/instructions.md
@@ -12,6 +12,8 @@ There are two queues for this ride, each represented as a `list`:
 
 You have been asked to write some code to better manage the guests at the park.
  You need to implement the following functions as soon as possible before the guests (and your boss, Chaitana!) get cranky.
+ Make sure you read carefully.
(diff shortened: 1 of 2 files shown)


BethanyG:
@ErikSchierboom  -- is it OK to merge this, or should I keep it open for you?

ErikSchierboom:
@BethanyG This does change the test file so that would re-queue all existing solutions. Is that necessary, or could this be merged with `[no important files changed]`?

BethanyG:
It does change the test file  to add more test scenarios & validate that the passed-in lists are mutated - so as a test to make sure the [bot fired](https://forum.exercism.org/t/no-important-files-changed-action-not-firing/13673/12), it works.

 Because it adds tests, my first reaction would be to re-test solutions.  But this exercise has a little over 16k completions, and  I believe the majority of solutions would still pass the new tests.  So I would probably apply `[no important files changed]` here.

ErikSchierboom:
Ah, I missed that this was also a test! Sorry. 

>  So I would probably apply [no important files changed] here.

Agreed.

PR #3832, opened 2024-11-29
PEP refactor

Ryrden:
When I faced this code, It was very hard to read, so I've refactored it following Python PEP recommendations

Here is all guidelines I've followed:

- Indentation: "Use 4 spaces per indentation level."  [PEP 8 - Indentation](https://peps.python.org/pep-0008/#indentation)

- Line Length: "Limit all lines to a maximum of 79 characters." [PEP 8 - Line Length](https://peps.python.org/pep-0008/#maximum-line-length)

- Blank Lines: "Surround top-level function and class definitions with two blank lines." [PEP 8 - Blank Lines](https://peps.python.org/pep-0008/#blank-lines)

- Spacing Around Operators: "Put a space on both sides of an assignment (=) and comparison (<, >, ==) operators." [PEP 8 - Spacing Around Operators](https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements)

- Comments: "Inline comments should be separated by at least two spaces from the statement." [PEP 8 - Comments](https://peps.python.org/pep-0008/#comments)

### DIFF
--- a/exercises/practice/pythagorean-triplet/.approaches/introduction.md
+++ b/exercises/practice/pythagorean-triplet/.approaches/introduction.md
@@ -146,23 +146,20 @@ Although it is important to note that this solution could have chosen a better n
 ```python
 def triplets_with_sum(number):
     def calculate_medium(small):
-         
         # We have two numbers, but need the third.
         return (number ** 2 - 2 * number * small) / (2 * (number - small))
 
     two_sides = (
-                  (int(medium), small) for 
-                  small in range(3, number // 3) if
-                  
-                  #Calls calculate_medium and assigns return value to variable medium 
-                  small < (medium := calculate_medium(small)) and 
-                  medium.is_integer()
-                  )
+        (int(medium), small) for small in range(3, number // 3) if
+
+        #Calls calculate_medium and assigns return value to variable medium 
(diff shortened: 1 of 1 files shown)


BethanyG:
Hi @Ryrden 👋🏽 

Thanks for submitting this.  However, this approach was written by someone who was very careful to format the code in a specific fashion.  PEP-8 states (emphasis mine):

> **A Foolish Consistency is the Hobgoblin of Little Minds**
> ...
> _A style guide is about consistency. Consistency with this style guide is important. **Consistency within a project is more important.** Consistency within one module or function is the most important._
>
> _However, know when to be inconsistent – **sometimes style guide recommendations just aren’t applicable.** When in doubt, use your best judgment. Look at other examples and decide what looks best._

And while PEP 8 does say line length of 79, we adhere to a line length of 120 to avoid onerous and excessive file length and line breaks.  We also make extensive use of implicit concatenation within comprehensions and generator expressions.   That is why the code is broken up the way it is. 

I am sorry that this was difficult for you to read, but (_for now - we may revisit in the future_) we want to keep the file as-is.

Ryrden:
Never mind, Thanks for the feedback!

PR #3876, opened 2025-03-02
[Swift Scheduling]: Added Files for New Practice Exercise

BethanyG:


BethanyG:
Decided to leave this as-is, and to see if we can do a follow-on exercise that practices [datetime.strftime()](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes).

PR #3947, opened 2025-07-27
Update instructions.append.md

jmh-git:
Instruction does not match the unit test. The exception message tested in the unit test is more precise. Hence, I've changed the instruction.


BethanyG (approved):
Thank you very much for this! 🚀

PR #3948, opened 2025-07-28
[Pangram]: fix redundant double iteration in dig deeper solution

princemuel:
Issue: Current implementation creates an intermediate list from a set, which is redundant.

**Before:**
```python
return len([ltr for ltr in set(sentence.lower()) if ltr.isalpha()]) == 26
```
**After:**
```python
return len(set(ltr for ltr in sentence.lower() if ltr.isalpha())) == 26
```
Why this is better:
- Eliminates double iteration: Old version iterates once for set(), again for list comprehension
- Removes unnecessary list creation: No need to convert set → list just to count
- Better memory usage: Generator expression feeds directly into set constructor
- Same time complexity but more efficient constant factors

The old approach first deduplicates all the characters, then filters alphabetic ones. The new approach filters while building the set, which is the natural order of operations for this problem.


princemuel:
Hi @BethanyG 

I just saw this. I'm currently swamped with work at the moment but I'd like to work on it over the weekend.

princemuel:
@BethanyG Hi. So I've updated the relevant docs, code and links 🤖 Please review and let me know if there are any changes to be made.

princemuel:
I noticed the [config.json][config-link] file. After the review, would I need to add my username to it? 🤔

[config-link]: https://github.com/exercism/python/blob/main/exercises/practice/pangram/.approaches/config.json

BethanyG (approved):
@princemuel   Heya!

Apologies for the delay in review.  This looks good, and I have approved it. 

Thanks so much for doing this.  😄   I've added you in as a contributor for both the approach and the benchmarking.  That should up your contribution rep.  Please let me know if you see any issues or have any questions.

I will merge to the website shortly.

PR #4039, opened 2025-12-03
test: add case for odd starting value in Collatz sequence

titikshit:
**Overview**
Adds a missing test case for handling odd-number Collatz inputs.

**Motivation**
Odd starts (like 25) help surface an easy mistake: using / produces floats, which breaks the integer sequence. 
Using // keeps the logic consistent.
This test highlights the issue, and I suggest recommending // in the implementation

**Changes**
Added test_odd_number_start to verify steps(25) == 23.


BethanyG:
Hi @titikshit 👋🏽 

This problem is a practice exercise that has been pulled from a [common repository](https://github.com/exercism/problem-specifications/tree/main/exercises/collatz-conjecture).  The content and tests have been auto-generated.  To add, alter, or propose a test, we ask that you open an issue in the [forum](forum.exercism.org) under `problem specifications`, since that is where this change would go so that it benefits more than just the Python track.

Issue #3795, opened 2024-10-21
Insufficient test cases for "Resistor Color Expert"


xanni:
I submitted code that passed all the existing tests but failed to correctly handle some actual resistor values in use in the E48 (2% tolerance) and larger series.  There are currently no tests for 5-band resistors where the second band is black, which my earlier iteration handled wrongly.

BethanyG:
Linking [forum discussion](https://forum.exercism.org/t/insufficient-test-cases-for-resistor-color-expert/13465/2).

Happy to accept suggested test cases and/or PR.  😄

xanni:
PR created: https://github.com/exercism/python/pull/3797

BethanyG:
I am going to keep this open until we figure out if we want 3-band cases.  I am leaning towards "no", but want to discuss it a bit on the forum first.  😄

BethanyG:
Hi @deekshahavaragi  -- while I appreciate the enthusiasm, the forum discussion related to this came down on the side of no additional edge cases, since they were wandering away from the intent of the exercise.  I forgot to close this issue, but will be doing so now.

Additionally, your PR appears to only have made whitespace changes - was that intentional?

PR #4047, opened 2025-12-13
Add additional edge-case tests for resistor-color-expert (#3795)

deekshahavaragi:
Adds additional edge-case tests for resistor-color-expert as discussed in issue #3795.


BethanyG:
Per the cited forum thread, this is not a test case we want added.  Thanks!

PR #4122, opened 2026-04-08
[D&D Character] fix typos in approaches

colinleach:
As you said: 

> So we'll see all the typos shortly....

This is the first tranche. There seems to be something cursed about `setattr`.

Where I removed the bit that says `(_as is the variant below_)` on the last line of that document, I wondered if the whole paragraph is repeated needlessly? Line 78 duplicates line 41.

### DIFF
--- a/exercises/practice/dnd-character/.approaches/introduction.md
+++ b/exercises/practice/dnd-character/.approaches/introduction.md
@@ -229,7 +229,7 @@ class Character:
         self.hitpoints = modifier(self.constitution) + 10
 ```
 
-Listing out each ability vs looping through and using `setatter()` has identical results for the object.
+Listing out each ability vs looping through and using `setattr()` has identical results for the object.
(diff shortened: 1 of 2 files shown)


BethanyG:
> Where I removed the bit that says (_as is the variant below_) on the last line of that document, I wondered if the whole paragraph is repeated needlessly? Line 78 duplicates line 41.

Yes - it is repetitive, but I thought I'd keep the pattern of fully describing each code block.  I know for that last one the "big idea" is where to put the `ability` tuple ... but decided that duplicating the sentence wasn't the end of the world.

Conversely, if you'd like to change the code example to use one of the other non-static method variants, we can then talk about how it follows one of the other approaches.  😄

colinleach:
> if you'd like to change the code example

I really wouldn't like to do this!

No worries about the repetition. The main point stands: we couldn't have "the variant below" as the last words in the document, where there's no "below".

BethanyG (approved):

PR #4126, opened 2026-04-13
New string formatting concept exercise

BNAndras:
WIP but will need introduction, instructions, hints, etc. I'll also need to undo the formatting on the track config since it's pretty noisy.

### DIFF
--- a/config.json
+++ b/config.json
@@ -19,10 +19,18 @@
     "average_run_time": 2
   },
   "files": {
-    "solution": ["%{snake_slug}.py"],
-    "test": ["%{snake_slug}_test.py"],
-    "example": [".meta/example.py"],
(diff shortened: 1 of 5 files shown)


BethanyG:
I really apologize - got distracted, and this has sat too long.  It might be best (between the docstring changes and the config.json changes) to re-submit this instead of trying to fix merge conflicts.

We've changed the docstring styles to Google - so you'll need to look at the other concept exercises to see how that looks.

I also noticed you don't have any  tasks that cover `<str>.format()`.  Weren't we going to add at least one or maybe two?

BethanyG:
@BNAndras  - any interest in resubmitting this, or do you want to move on?  😄

BethanyG:
I am going to close this for now. We can take it up again sometime, if there is interest.

Issue #4165, opened 2026-05-13
[Typo Patrol]: Typos & Similar Errors in Concept Exercise instructions, stub docstrings, & exemplar docstrings.


Yrahcaz7:
As per [this forum post](http://forum.exercism.org/t/cater-waiter-docstrings-have-a-different-style-and-some-typos/50755/5), this issue will be a collection of all typos and similar errors I find from now on, grouped by exercise/concept.

## Concept Exercises

- [x] [Locomotive Engineer](https://exercism.org/tracks/python/exercises/locomotive-engineer) (addressed in [PR 4173](https://github.com/exercism/python/pull/4173)):

    <details>
    <summary>details</summary>
  
    - Typos in the docstrings in [the stub](https://github.com/exercism/python/blob/main/exercises/concept/locomotive-engineer/locomotive_engineer.py):
      - Line 8 has an indentation of 7 spaces instead of 8.
      - Line 21 and 59 are missing the `:` that is normally in between the type and description of parameters.
      - On line 34, the parameter description "`arbitrary number of stops.`" is not capitalized. Line 62 (`list[tuple]: the list of rows of wagons.`) is the same.
      - On line 59, the `wagons_rows` parameter is said to be of type `list[tuple]`, but it is actually of type `list[list[tuple]]`. The return type has the same problem (line 62).
    - [The exemplar solution](https://github.com/exercism/python/blob/main/exercises/concept/locomotive-engineer/.meta/exemplar.py)'s docstrings have most of these issues too, as well as an extra period at the end of line 11.
    - Typos in [`hints.md`](https://github.com/exercism/python/blob/main/exercises/concept/locomotive-engineer/.docs/hints.md):
      - Line 5 has incorrect grammar, it should probably be something along these lines instead: "To extract multiple arguments in the function parameters, you can pack them with the `*args` operator for `list` or `tuples` or `**kwargs` for keyword-based arguments." This replaces "parameters so can you" with "parameters, you can".
      - The comma on line 14 isn't necessary and probably isn't grammatical anyway.
      - Line 31 isn't grammatical, it should be something along these lines instead: "`zip(*iterators)` can be used to transpose a nested `list`." This replaces "can use used" with "can be used".
[...]

PR #4179, opened 2026-05-15
[Basics, Conditionals, and List Methods]: Corrected Typos, Formatting, & Grammar Issues

BethanyG:
Please see [Issue 4165](https://github.com/exercism/python/issues/4165) for more details.

### DIFF
--- a/concepts/basics/about.md
+++ b/concepts/basics/about.md
@@ -55,8 +55,8 @@ For example, `my_first_variable` can be re-assigned many times using `=`, and ca
 
 
 ```python
->>> my_first_variable = 1  # <--my_first_variable bound to an integer object of value one.
->>> my_first_variable = 2  # <--my_first_variable re-assigned to integer value 2.
+>>> my_first_variable = 1  # my_first_variable bound to an integer object of value one.
+>>> my_first_variable = 2  # my_first_variable re-assigned to integer value 2.
 
 >>> print(type(my_first_variable))
 <class 'int'>
--- a/concepts/basics/about.md
+++ b/concepts/basics/about.md
@@ -102,7 +102,7 @@ MY_FIRST_CONSTANT = "Some other value"
 
 ## Functions
 
-In Python, units of functionality are encapsulated in [_functions._][functions], which are themselves [objects][objects] (_it's [turtles all the way down][turtles all the way down]_).
+In Python, units of functionality are encapsulated in [_functions_][functions], which are themselves [objects][objects] (_it's [turtles all the way down][turtles all the way down]_).
 
 Functions can be executed by themselves, passed as arguments to other functions, nested, or bound to a class.
 When functions are bound to a [class][classes] name, they're referred to as [methods][method objects].
@@ -114,7 +114,7 @@ Statements for the _body_ of the function begin on the line following `def` and 
 
 
 ```python
-# The body of a function is indented by 2 spaces, & prints the sum of the numbers.
+# The body of a function is indented by 2 spaces & prints the sum of the numbers.
 def add_two_numbers(number_one, number_two):
   total = number_one + number_two
   print(total)  
@@ -126,7 +126,7 @@ def add_two_numbers(number_one, number_two):
 # Inconsistent indentation in your code blocks will raise an error.
 >>> def add_three_numbers_misformatted(number_one, number_two, number_three):
 ...     result = number_one + number_two + number_three   # This was indented by 4 spaces.
-...    print(result)     #this was only indented by 3 spaces
+...    print(result)     # <--This was only indented by 3 spaces.
 ...
 ...
   File "<stdin>", line 3
@@ -157,7 +157,7 @@ def add_two_numbers(number_one, number_two):
 ```
 
 Functions that do not have an _explicit_ expression following a `return` will _implicitly_ return the [`None`][none] object.
-The details of `None` will be covered in a later exercise.
+The details of `None` will be covered in a later concept.
 For the purposes of this exercise and explanation, `None` is a placeholder that represents nothing, or null:
 
 
@@ -208,7 +208,7 @@ Dot (`.`) notation is used for calling functions defined inside a class or modul
 
 ```python
 >>> def raise_to_power(number, power):
-        return number ** power
+...     return number ** power
 ...
 
 >>> raise_to_power(3,3) # Invoking the function with the arguments 3 and 3.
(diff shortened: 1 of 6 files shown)


concepts/basics/about.md
```diff
@@ -239,9 +239,8 @@ TypeError: raise_to_power() missing 1 required positional argument: 'power'
 
 
 # Importing the math module
-import math
-
->>> math.pow(2,4)  # Calling the pow() function from the math module
+>>> import math
+>>> math.pow(2,4)  # <--Calling the pow() function from the math module.
```
Yrahcaz7, line 243: Since this was changed, the occurrence on line 75 should be as well.

concepts/basics/about.md
```diff
@@ -333,10 +343,11 @@ errors is specified, then the object must expose a data buffer
 that will be decoded using the given encoding and error handler.
 Otherwise, returns the result of object.__str__() (if defined)
 or repr(object).
-encoding defaults to sys.getdefaultencoding().
+encoding defaults to 'utf-8'.
```
Yrahcaz7, line 346: Was this statement just incorrect before? I think `sys.getdefaultencoding()` doesn't always return "utf-8".

concepts/conditionals/about.md
```diff
@@ -57,18 +57,17 @@ else:
...
+...      if number % 3 == 0 and number % 5 == 0:
+...          say = 'FizzBuzz!'
+...      elif number % 5 == 0:
+...          say = 'Buzz!'
+...      elif number % 3 == 0:
+...          say = 'Fizz!'
+...      else:
+...          say = str(number)
+...
+...      return say
```
Yrahcaz7: It seems like this block got indented by an extra space during the reformatting.

concepts/conditionals/about.md
Yrahcaz7, line 122: Since the indentation of the `else` statement was changed, this example no longer works.

concepts/conditionals/about.md
```diff
@@ -152,29 +152,29 @@ Objects that are evaluated in this fashion are considered "truthy" or "falsy", a
...
 ```python
 >>> def truthy_test(thing):
-        if thing:
-            print('This is Truthy.')
-        else:
-            print("Nope. It's Falsey.")
+...     if thing:
+...         print('This is Truthy.')
+...     else:
+...         print('Nope. It's Falsy.')
```
Yrahcaz7: This needs the `'` in the middle of the string to be escaped now.

concepts/conditionals/introduction.md
```diff
@@ -58,16 +58,16 @@ else:
...
+...      if number % 3 == 0 and number % 5 == 0:
+...          say = 'FizzBuzz!'
+...      elif number % 5 == 0:
+...          say = 'Buzz!'
+...      elif number % 3 == 0:
+...          say = 'Fizz!'
+...      else:
+...          say = str(number)
+...
+...      return say
```
Yrahcaz7: This block also got indented by extra space.

Yrahcaz7 (changes requested):

concepts/basics/about.md
BethanyG: Since this example is copied directly from the terminal after running `print(str.__doc__)`, I am assuming that it changed between Python versions.  Not positive, but I think that this bit was last touched on Exercism prior to Python 3.11.5, but I can't find it in the history.  You _might_ be able to find the `.__doc__` change for `str` in the [Python language repo](https://github.com/python) or [docs](https://docs.python.org/3/index.html) (_careful! these default to Python 3.14, so you need to search by Python version_) somewhere, but  you'd have to dig by Python version.  It is the Python docsring for the `str` class/type.

concepts/conditionals/about.md
BethanyG: I am starting to remember why I ripped out many of these.  This is  really brittle, and not exactly something we should be [using a lot of in our docs](https://devguide.python.org/documentation/style-guide/#code-examples).  I am going to keep them here since these exercises are early in the track, and students should be exposed to how the REPL looks/works.  But I think for later exercises, I am going to omit these continuation marks.  As stated in the developer guide, they make it hard to cut and paste the code into the REPL.  And maintaining the examples like this will be a perpetual source of back and forth methinks.

concepts/basics/about.md
Yrahcaz7: Huh... "utf-8" seems to have been the default as far back as Python 3.2, and the docs don't seem to specify the default before then. I wonder where the `sys.getdefaultencoding()` came from.

BethanyG: Dunno.  It might have been me or someone else running (unknowingly) a Python 2.7 version in an env.  It's fixed now, tho. 🤷🏽‍♀️  These are the sorts of things that make me crazy when we upgrade the Python versions.  🙂   It's never "as simple as installing a new Python".  😉

concepts/conditionals/about.md
Yrahcaz7: I see... Maybe it would be a good idea to move the function out of the REPL input and into a separate code block. I think there are a few examples of that in the docs, and I think it works well when used with a function definition.

concepts/basics/about.md
```diff
@@ -55,25 +55,25 @@ For example, `my_first_variable` can be re-assigned many times using `=`, and ca
 
 
 ```python
->>> my_first_variable = 1  # my_first_variable bound to an integer object of value one.
->>> my_first_variable = 2  # my_first_variable re-assigned to integer value 2.
+>>> my_first_variable = 1  # <--my_first_variable bound to an integer object of value one.
+>>> my_first_variable = 2  # <--my_first_variable re-assigned to integer value 2.
```
Yrahcaz7: Some suggestions: I think the comments read better if the "one" is changed to a numeric "1". The "my_first_variable" also looks a little strange when right up against the `<--`, maybe surrounding the "my_first_variable" with backticks would work?

concepts/conditionals/about.md
Yrahcaz7: This line still has the problem with the quotes.

concepts/conditionals/about.md
```diff
@@ -57,17 +57,7 @@ else:
...
-...          say = 'FizzBuzz!'
-...      elif number % 5 == 0:
-...          say = 'Buzz!'
-...      elif number % 3 == 0:
-...          say = 'Fizz!'
-...      else:
-...          say = str(number)
-...
-...      return say
+>>>
```
Yrahcaz7: If this is removed, then `classic_fizzbuzz()` will be called before the reader sees it in the next code block. If one of the function definitions is removed, it should probably be the second one.

Yrahcaz7 (changes requested):

concepts/conditionals/about.md
BethanyG: Let's leave these alone once corrected this round.  I really REALLY don't want to reformat all the code examples in all the concept exercises yet again.  That's not really a high leverage activity.  

We'll omit continuations starting with `list`s, and (maybe) transition to plain code examples eventually.  But _not right now_.  These docs will never be perfect, and trying will pull energy from creating new concepts and exercises and other things that are higher priority.

concepts/basics/about.md
BethanyG: Backticks don't render when they are used in comments and code blocks and I think that adds visual noise.  

I can put a space in between <-- and the rest of the comment or put it up against the `#`.  Although bumping it up against the `#` makes it look like it somehow belongs to the comment mark, which it doesn't.

I can also omit the <-- if it doesn't seem helpful.

BethanyG: I'm omitting them in this case for readability.

Yrahcaz7: Hmm... For other places with more going on, having the `<--` seems useful, but here I don't think it adds very much. Maybe omitting them here would be better.

concepts/conditionals/about.md
BethanyG: This was a mistake.  Corrected.

Yrahcaz7 (approved):

Issue #4055, opened 2025-12-24
[Exercise Approaches]:  Audit/Cleanup of Track Approaches 2025/2026


BethanyG:
<mark>**This is a tracking issue for existing approaches on the Python track**</mark>
This was Part 1.  You can find Part 2 [here under issue 4197](https://github.com/exercism/python/issues/4197)

[Related forum thread for general discussion](https://forum.exercism.org/t/python-track-audit-of-approaches-docs/20229)

During discussion on this [Forum Post](https://forum.exercism.org/t/fix-syntax-in-space-age-mentoring-notes-on-python-track/19954/17), @petrem volunteered to audit the Python track approaches for the following:

-  Spelling, grammar, typos
-  Python syntax
-  Clarity of writing & explanation
- Readability and maintainability
- White space
- Other Exercism markdown specifications as needed
-  Possible missing/additional approaches (_noted as TODOs of future work_)

See the table of existing/in progress/missing approaches below.  Exercises with one ✅ are unaudited, exercises with two ✅ ✅  have been reviewed, changed and merged. <mark>highlighted</mark> exercises are in review.

Exercises without checkmarks or PRs do not yet have approaches.  If you want to work on an exercise that is missing an approach, please check in with the maintainers to avoid conflicting with current track priorities or others who may be working on things. 

<details>
<summary>Table of Existing/In Progress/Missing Approaches</summary>

|Completed? | Exercise Name|
|--- | ---|
|✅ | Acronym|
| | Affine Cipher|
|✅  | All Your Base|
|IN PROGRESS (PR forthcoming)| Allergies|
| | Alphametics|
| IN PROGRESS (PR forthcoming)| Anagram|
| IN PROGRESS (PR forthcoming)| Armstrong Numbers|
|✅ | Atbash Cipher|
| | Bank Account|
| | Binary Search Tree|
| | Binary Search|
|✅ | Bob|
|IN PROGRESS (PR forthcoming)| Book Store|
| | Bottle Song|
| | Bowling|
| | Change|
|✅ | Circular Buffer|
| | Clock|
|✅ | Collatz Conjecture|
| | Complex Numbers|
| | Connect|
| | Crypto Square|
|✅ | Darts|
| | Diamond|
| | Difference Of Squares|
| | Diffie Hellman|
|✅  | Dnd Character|
| | Dominoes|
| | Dot Dsl|
|IN PROGRESS (PR 3378) | Etl|
| | Flatten Array|
| | Food Chain|
| | Forth|
[...]

PR #4191, opened 2026-05-19
[RNA Transcription & Atbash Cipher]: Approaches Cleanup

Yrahcaz7:
As discussed in [this issue](https://github.com/exercism/python/issues/4055).

### DIFF
--- a/exercises/practice/atbash-cipher/.approaches/introduction.md
+++ b/exercises/practice/atbash-cipher/.approaches/introduction.md
@@ -5,10 +5,10 @@ Atbash cipher in Python can be solved in many ways.
 ## General guidance
 
 The first thing is to have a "key" mapping — possibly in a `dict` or `str.maketrans()`, otherwise the value would have to be calculated on the fly.
-Then, you have to "clean" up the string to be encoded by removing punctuation/whitespace.
+Next, you have to "clean" up the string to be encoded by removing punctuation/whitespace.
 Finally, you break it up into chunks of five before returning it.
 
-For decoding, it's similar — clean up (which automatically joins the chunks) and translate using the _same_ key — the realization that the same key can be used is crucial in solving this in an idiomatic manner.
+For decoding, the process is similar — clean up (_which automatically joins the chunks_) and translate using the **_same_** key — realizing that the same key can be used is crucial in solving this in an idiomatic manner.
 
 ## Approach: Separate functions
 
--- a/exercises/practice/atbash-cipher/.approaches/introduction.md
+++ b/exercises/practice/atbash-cipher/.approaches/introduction.md
@@ -1,44 +1,54 @@
 # Introduction
+
 Atbash cipher in Python can be solved in many ways.  
 
 ## General guidance
-The first thing is to have a "key" mapping - possibly in a `dict` or `str.maketrans`, otherwise the value would have to be calculated on the fly.
-Then, you have to "clean" up the string to be encoded by removing numbers/whitespace.
+
+The first thing is to have a "key" mapping — possibly in a `dict` or `str.maketrans()`, otherwise the value would have to be calculated on the fly.
+Then, you have to "clean" up the string to be encoded by removing punctuation/whitespace.
 Finally, you break it up into chunks of five before returning it.
 
-For decoding, it's similar - clean up (which automatically joins the chunks) and translate using the _same_ key - the realization that the same key can be used is crucial in solving this in an idiomatic manner.
+For decoding, it's similar — clean up (which automatically joins the chunks) and translate using the _same_ key — the realization that the same key can be used is crucial in solving this in an idiomatic manner.
 
-## Approach: separate functions
-We use `str.maketrans` to create the encoding. 
-In `encode`, we use a [generator expression][generator-expression] in `str.join`.
+## Approach: Separate functions
+
+We use `str.maketrans()` to create the encoding.
+In `encode()`, we use a [generator expression][generator-expression] in `str.join()`.
+
 ```python
 from string import ascii_lowercase
+
 ENCODING = str.maketrans(ascii_lowercase, ascii_lowercase[::-1])
 
-def encode(text: str):
+def encode(text):
     res = "".join(chr for chr in text.lower() if chr.isalnum()).translate(ENCODING)
     return " ".join(res[index:index+5] for index in range(0, len(res), 5))
 
-def decode(text: str):
-    return "".join(chr.lower() for chr in text if chr.isalnum()).translate(ENCODING)
+def decode(text):
+    return "".join(chr.lower() for chr in text if not chr.isspace()).translate(ENCODING)
 ```
+
 Read more on this [approach here][approach-separate-functions].
 
-## Approach: mono-function
-Notice that there the majority of the code is repetitive? 
-A fun way to solve this would be to keep it all inside the `encode` function, and merely chunk it if `decode` is False:
-For variation, this approach shows a different way to translate the text.
+## Approach: Mono-function
+
+Notice that the majority of the code is repetitive?
+A fun way to solve this would be to keep it all inside the `encode()` function, and merely chunk it if `decode` is `False`:
+For variation, this approach also shows a different way to translate the text.
+
 ```python
 from string import ascii_lowercase as asc_low
+
 ENCODING = {chr: asc_low[id] for id, chr in enumerate(asc_low[::-1])}
 
-def encode(text: str, decode: bool = False):
-    res = "".join(ENCODING.get(chr, chr) for chr in text.lower() if chr.isalnum())
-    return res if decode else " ".join(res[index:index+5] for index in range(0, len(res), 5))
+def encode(text, decode = False):
+    line = "".join(ENCODING.get(chr, chr) for chr in text.lower() if chr.isalnum())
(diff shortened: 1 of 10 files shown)


exercises/practice/atbash-cipher/.approaches/mono-function/content.md
```diff
@@ -1,46 +1,53 @@
...
-def decode(text: str):
+def decode(text):
     return encode(text, True)
 ```
-To explain the translation: we use a `dict` comprehension in which we reverse the ASCII lowercase digits, and enumerate through them - that is, `z` is 0, `y` is 1, and so on. 
-We access the character at that index and set it to the value of `c` - so `z` translates to `a`.
 
-In the calculation of the result, we try to obtain the value of the character using `dict.get`, which accepts a default parameter. 
-In this case, the character itself is the default - that is, numbers won't be found in the translation key, and thus should remain as numbers.
+Here, we use a dictionary comprehension in which we reverse the ASCII lowercase digits, and enumerate through them — that is, `z` has index 0, `y` has index 1, and so on.
```
BethanyG:
```suggestion
Here, we use a dictionary comprehension in which we reverse the order of the ASCII lowercase digits and enumerate through them — that is, `z` is at index 0, `y` is at index 1, and so on.
```

exercises/practice/atbash-cipher/.approaches/mono-function/content.md
```diff
@@ -1,46 +1,53 @@
...
+For each character, we set the value of `chr` in the resulting dictionary to the character at the respective index — so `z` would translate to `a`.
```
BethanyG:
```suggestion
For each code point, we set the value of `chr` in the resulting dictionary to the code point at the respective index — so `z` translates to `a`.
```

exercises/practice/atbash-cipher/.approaches/mono-function/content.md
```diff
@@ -1,46 +1,53 @@
...
 
-We use a [ternary operator][ternary-operator] to check if we actually mean to decode the function, in which case we return the result as is. 
+In the calculation of the result, we try to obtain the value of the character using `dict.get()`, which accepts a default parameter.
```
BethanyG:
```suggestion
In the calculation of the result, we try to obtain the value of the code point using `dict.get()`, which accepts a default parameter.
```

exercises/practice/atbash-cipher/.approaches/mono-function/content.md
```diff
@@ -1,46 +1,53 @@
...
+In this case, the character itself is the default — that is, numbers won't be found in the translation key, and thus should remain as numbers.
```
BethanyG:
```suggestion
In this case, the code point itself is the default — that is, numbers won't be found in the translation key, and thus should remain as numbers.
```

exercises/practice/atbash-cipher/.approaches/mono-function/content.md
```diff
@@ -1,46 +1,53 @@
...
+
+We use a [conditional expression (also known as a ternary operator)][conditional-expression] to check if we actually mean to decode the function, in which case we return the result as is.
 If not, we chunk the result by joining every five characters with a space.
```
BethanyG:
```suggestion
If not, we "chunk" the result by joining every five code points with a space.
```

exercises/practice/atbash-cipher/.approaches/mono-function/content.md
```diff
@@ -1,46 +1,53 @@
...
 
-Another possible way to solve this would be to use a function that returns a function that encodes or decodes based on the parameters:
+Another possible way to solve this would be to use a function that returns another function that encodes or decodes based on the outer function's parameter:
```
BethanyG:
```suggestion
Another possible way to solve this would be to use a function that returns another function (_a higher-order function or [closure][closure]_) that encodes or decodes based on the outer function's parameter:
```

exercises/practice/atbash-cipher/.approaches/mono-function/content.md
```diff
@@ -1,46 +1,53 @@
...
-[decorator]: https://realpython.com/primer-on-python-decorators/
\ No newline at end of file
+The logic is the same — the only change is that now we use use one function that generates two _other_ functions based on the boolean value of its parameter.
+
+Here, we first call `code()` with no argument and set `encode` to the function that's returned, which performs encoding.
+Then we call `code(True)` to get the decoding version of the function and set `decode` to that function.
+
+After that, we can call `encode()` and `decode()` as normal, and both functions successfully perform their indended task.
+
+[conditional-expression]: https://docs.python.org/3/reference/expressions.html#conditional-expressions
```
BethanyG, line 54:
```suggestion
[closure]: https://realpython.com/python-closure/
[conditional-expression]: https://docs.python.org/3/reference/expressions.html#conditional-expressions
```

exercises/practice/atbash-cipher/.approaches/introduction.md
```diff
@@ -1,44 +1,54 @@
 # Introduction
+
 Atbash cipher in Python can be solved in many ways.  
 
 ## General guidance
-The first thing is to have a "key" mapping - possibly in a `dict` or `str.maketrans`, otherwise the value would have to be calculated on the fly.
-Then, you have to "clean" up the string to be encoded by removing numbers/whitespace.
+
+The first thing is to have a "key" mapping — possibly in a `dict` or `str.maketrans()`, otherwise the value would have to be calculated on the fly.
+Then, you have to "clean" up the string to be encoded by removing punctuation/whitespace.
```
BethanyG:
```suggestion
Next, you have to "clean" up the string to be encoded by removing punctuation/whitespace.
```

exercises/practice/atbash-cipher/.approaches/introduction.md
```diff
@@ -1,44 +1,54 @@
...
 Finally, you break it up into chunks of five before returning it.
 
-For decoding, it's similar - clean up (which automatically joins the chunks) and translate using the _same_ key - the realization that the same key can be used is crucial in solving this in an idiomatic manner.
+For decoding, it's similar — clean up (which automatically joins the chunks) and translate using the _same_ key — the realization that the same key can be used is crucial in solving this in an idiomatic manner.
```
BethanyG:
```suggestion
For decoding, the process is similar — clean up (_which automatically joins the chunks_) and translate using the **_same_** key — realizing that the same key can be used is crucial in solving this in an idiomatic manner.
```

exercises/practice/rna-transcription/.approaches/dictionary-join/content.md
```diff
@@ -18,7 +17,7 @@ It indicates that the value is not intended to be changed.
 In the `to_rna()` function, the [`join()`][join] method is called on an empty string,
 and is passed the list created from a [generator expression][generator-expression].
 
-The generator expression iterates each character in the input,
+The generator expression iterates over each character in the input,
```
BethanyG:
```suggestion
The generator expression iterates over each code point in the input,
```

BethanyG (changes requested):
I am being a pendant here, but I think it is really important to poke students about "character" vs "code point".  We have to be careful, even with basic ASCII to not confuse the two - it becomes a deadly mental model the minute you get to anything with a diacritical mark or accent.  Its even worse when you work with any multi-codepoint glyphs (like emoji or Kanji).  It's a ...whole thing.

Otherwise, great work here.  Thanks so much for cleaning this mess up.

exercises/practice/atbash-cipher/.approaches/separate-functions/content.md
```diff
@@ -1,45 +1,57 @@
...
 
-def decode(text: str):
-    return "".join(chr.lower() for chr in text if chr.isalnum()).translate(ENCODING)
+def decode(text):
+    return "".join(chr.lower() for chr in text if not chr.isspace()).translate(ENCODING)
 ```
-In `encode`, we first join together every character if the character is alphanumeric - as we use `text.lower()`, the characters are all lowercase as needed.
+
+In `encode()`, we first join together every character that is alphanumeric — as we use `text.lower()`, the characters are all lowercase as needed.
 Then, we translate it and return a version joining every five characters with a space in between.
```
Yrahcaz7: [Line 18/22] Should this be something like "every code point that is an alphanumeric character" instead of "every character that is alphanumeric"?

[Line 19/23] Should this instance of "characters" be "code points"?

BethanyG: Yup.  _You got me!_  😆   I missed those.  Sorry about that.

Yrahcaz7: Totally fine! You could've even just left the note about character points and I would've changed them myself, really.

BethanyG (approved):
Still a tad triggered by those last "character" refs but agree they are better as-is for clarity.  (Or at least I hope that brings clarity).

BethanyG:
Shall we let this one sit and re-read, or are you read for merge?  😄

Yrahcaz7:
You bet I am ready to merge! I am not letting that <s>slop</s> err... subpar approach remain there any longer!

BethanyG:
BTW, if another approach for any of these comes to you, please feel free to write one (or not!).

PR #4196, opened 2026-05-21
Remove unfinished `functools` concept

Yrahcaz7:
As per [this issue](https://github.com/exercism/python/issues/4165#issuecomment-4503602427), the concept will be removed. It should be reworked and re-added at a later date.


BethanyG (approved):
Thank you.  💙

PR #4198, opened 2026-05-21
[Wordy] Clean up Approaches

Yrahcaz7:
As discussed in #4055 and #4197.

PR #4202, opened 2026-05-22
[Wordy] Add `tuple-and-index` approach

Yrahcaz7:
Adds another Wordy approach [as discussed in this PR](https://github.com/exercism/python/pull/4198#issuecomment-4504509150).

Also fixes a few errors in the Wordy approach docs that were missed in that PR.

### DIFF
--- a/exercises/practice/wordy/.approaches/tuple-and-index/content.md
+++ b/exercises/practice/wordy/.approaches/tuple-and-index/content.md
@@ -67,7 +67,7 @@ However, some operator phrases are multiple words long, so we need to check if `
 If there *is* an extra operator word, then we need to check if it is present as the next word in `words`.
 If it is not present, we raise a `ValueError`, else we increment `operand_index` by `1` to get the correct index.
 
-Here we call the helper function again, setting `operand` to the number at index `operand_index`, and raising a `ValueError` if it is not a number.
+Here we use the helper function `str_to_int()` again, setting `operand` to the number at index `operand_index` (and raising a `ValueError` if it is not a number).
 After this, the approach uses [structural pattern matching][structural-pattern-matching] to modify `result` using `+=`, `-=`, `*=`, or `//=` depending on the `operator_index`.
 However, this section could easily be modified to use a similar method to any of the other approaches, such as an `if-elif-else` block or a tuple of `lambda`s.
 
--- a/exercises/practice/wordy/.approaches/tuple-and-index/content.md
+++ b/exercises/practice/wordy/.approaches/tuple-and-index/content.md
@@ -64,8 +64,8 @@ If `words[index]` is not an operator word, the raised `ValueError` will be caugh
 
 Next, we need to get the second operand for the operator (we already have `result` for the first one), so we set `operand_index = index + 1`.
 However, some operator phrases are multiple words long, so we need to check if `EXTRA_OPERATOR_WORDS[operator_index] is not None`.
-If there is an extra operator word, we need to check if it is present as the next word in `words`.
-If it is not, we raise a `ValueError`, else we increment `operand_index` by `1` to get the correct index.
+If there *is* an extra operator word, then we need to check if it is present as the next word in `words`.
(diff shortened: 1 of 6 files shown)


exercises/practice/wordy/.approaches/tuple-and-index/content.md
```diff
@@ -0,0 +1,81 @@
...
+
+Inside the loop, we use [`OPERATOR_WORDS.index()`][sequence-index-method] to get the index of the operator at index `index` of `words`.
+If `words[index]` is not an operator word, the raised `ValueError` will be caught by the [`try-except`][handling-exceptions] statement, and the `except` block will determine the correct error to raise instead.
+
+Next, we need to get the second operand for the operator (we already have `result` for the first one), so we set `operand_index = index + 1`.
+However, some operator phrases are multiple words long, so we need to check if `EXTRA_OPERATOR_WORDS[operator_index] is not None`.
+If there *is* an extra operator word, then we need to check if it is present as the next word in `words`.
+If it is not present, we raise a `ValueError`, else we increment `operand_index` by `1` to get the correct index.
+
+Here we call the helper function again, setting `operand` to the number at index `operand_index`, and raising a `ValueError` if it is not a number.
```
BethanyG: You might want to name the helper function since this explanation is long enough to have to scroll back up to see the code.

exercises/practice/wordy/.approaches/tuple-and-index/content.md
```diff
@@ -0,0 +1,81 @@
...
+After this, the approach uses [structural pattern matching][structural-pattern-matching] to modify `result` using `+=`, `-=`, `*=`, or `//=` depending on the `operator_index`.
+However, this section could easily be modified to use a similar method to any of the other approaches, such as an `if-elif-else` block or a tuple of `lambda`s.
+
+At the end of the loop, we set `index` to `operand_index + 1`, as we have finished processing everything up to and including `operand_index`.
+Then the loop continues until `index` reaches or exceeds `len(words)` or something raises an uncaught error.
```
BethanyG:
```suggestion
Then the loop continues until `index` reaches or exceeds `len(words)` or an uncaught error is raised.
```

exercises/practice/wordy/.approaches/tuple-and-index/content.md
```diff
@@ -0,0 +1,81 @@
...
+When the loop ends, we know that we have finished processing the whole string, and thus we return `result`.
```
BethanyG:
```suggestion
When the loop ends, we know that we have finished processing the entire string, and thus we return `result`.
```

BethanyG (approved):
Approving this - good work!!

The three things I found are more suggestion than "OMG change this".  So you can decide if you agree or disagree.  😄 

LMK if you'd like to let this sit, or have it merged.  Thanks for doing this!!

exercises/practice/wordy/.approaches/tuple-and-index/content.md
Yrahcaz7: Good idea, but I'm not sure how to change the sentence to include that. Any change I try seems to make it read awkwardly...

BethanyG: What about 

> Here we call the helper function `str_to_int()` again, 

Or is that just too much?

BethanyG: Its fine if it just doesn't work.  It was just a thought - no need to cram it in if its not working.

Yrahcaz7: I changed it to: "Here we use the helper function `str_to_int()` again, setting `operand` to the number at index `operand_index` (and raising a `ValueError` if it is not a number)."

I think it works if that last part is in parentheses.

BethanyG (approved):
Nice work.  😄   Ready to merge?

Yrahcaz7:
I think so!

Issue #4197, opened 2026-05-21
[Exercise Approaches]: Audit/Cleanup of Track Approaches 2026 (part 2)


Yrahcaz7:
<mark>**This is a tracking issue for existing approaches on the Python track.**</mark>
(Part 1 of the audit/cleanup was issue #4055.)

In general, we are auditing/reviewing the following:

- Spelling, grammar, and typos
- Python syntax
- Clarity of writing and explanation
- Readability and maintainability
- White space
- Other [Exercism markdown specifications](https://exercism.org/docs/building/markdown/markdown) as needed
- Possible missing/additional approaches (_noted as TODOs of future work_)

See the lists for each category below. Exercises in "Merged Audits" have been reviewed, changed, and merged. Exercises in "Unaudited" have approaches that have not been audited yet.

Exercises in "Approaches In Progress" do not have finished/published approaches yet, while those in "Missing Approaches" do not have any known work on their approaches yet.

If you want to work on an exercise that is missing an approach, please check in with the maintainers to avoid conflicting with current track priorities or others who may be working on things.

<details>
<summary><strong>Merged Audits</strong></summary>

- Acronym: #4258
- Atbash Cipher: #4191
- Collatz Conjecture: #4250
- Darts: #4187
- Leap: #4252
- Reverse String: #4226
- RNA Transcription: #4052 and #4191
- Robot Name: #4051
- Rotational Cipher: #4184
- Sublist: #4190
- Wordy: #4198

</details>

<details>
<summary><strong>Unaudited</strong></summary>

- All Your Base
- Bob: **PR forthcoming**
- Circular Buffer
- DnD Character
- Flower Field (formerly Minesweeper)
- Grains
- Hamming
- Isogram
- Luhn
- Matching Brackets
- Nth Prime
- Palindrome Products
- Pangram
- Pig Latin
- Raindrops
- Roman Numerals
- Scrabble Score
- Sieve
- Yacht

</details>

<details>
<summary><strong>Approaches In Progress</strong></summary>

- Allergies
- Anagram
- Armstrong Numbers
- Book Store
- ETL: #3378
- Gigasecond: #3772
- ISBN Verifier
- Markdown
- Matrix
- Perfect Numbers
- Protein Translation
- Robot Simulator
- Saddle Points
- Say
- Space Age
- Sum Of Multiples: #3375
- Transpose
- Triangle
- Word Count: #3371

</details>

<details>
<summary><strong>Missing Approaches</strong></summary>

[...]

PR #4226, opened 2026-05-30
[Reverse String] Clean up approach docs

Yrahcaz7:
As discussed [here](https://github.com/exercism/python/issues/4197#issuecomment-4579447820).

Unfortunately, one line of `performance/snippet.md` had to be cut off to make the table render properly while keeping it within the 8-line limit.

Side note: Maybe there should be a doc somewhere that says how to link to headers in pages on the Exercism website? There was one in these docs, but it didn't work. I had to convert invalid characters to hyphens add an "h-" prefix to fix it: [`#h-using-map-and-lambbda-with-join-instead-of-a-loop`](https://exercism.org/tracks/python/exercises/reverse-string/approaches/additional-approaches#h-using-map-and-lambbda-with-join-instead-of-a-loop).


Yrahcaz7:
@BethanyG, I added another commit, but it isn't showing up. Will it get auto-added when the PR is reopened, or do I need to do some `git` weirdness?

BethanyG:
> @BethanyG, I added another commit, but it isn't showing up. Will it get auto-added when the PR is reopened, or do I need to do some git weirdness?

Apologies.  Just getting to this (busy morning).  That is very strange.  Did you push your change to your fork?  That should have been picked up automatically.  Also make sure you were on your local branch.  I cannot tell you how many times I've made an addition to something and then realized I'd switched branches (😱).  But it also might be the PR state.  I will reopen and we can check that now.

BethanyG:
> _Side note: Maybe there should be a doc somewhere that says how to link to headers in pages on the Exercism website? There was one in these docs, but it didn't work. I had to convert invalid characters to hyphens add an "h-" prefix to fix it: [#h-using-map-and-lambbda-with-join-instead-of-a-loop](https://exercism.org/tracks/python/exercises/reverse-string/approaches/additional-approaches#h-using-map-and-lambbda-with-join-instead-of-a-loop)._

🤔   Wonder if there is a way we can test this by (maybe) doing a test approach with a bunch of said links?  And then we can update the docs, or propose a change to them.  Lemme go dig and see if I can find the Exercism docs in question.  The likely path would be to  propose the changes on the forum.  So:

1.  Open a test PR (approach, hint file, instruction append file...) with header links both ways to test.
2.  Once we figure it out, propose a change to the Exercism docs to alert to the correct format on the forum
3.  PR the change (if you get the go-ahead).

BethanyG:
Try the push again to see if it shows up  now,  Otherwise, we have some git-wierdness to deal with.  Sorry.

Yrahcaz7:
It seems like the commit appeared after the PR was re-opened.

<br>

> Open a test PR (approach, hint file, instruction append file...) with header links both ways to test.

Is that necessary? I think you can see the correct format by inspecting the HTML of any page on the Exercism website. (That's how I figured it out, at least.)

Example:

Screenshot 2026-05-30 at 3 43 26 PM

<hr>

Edit: Just realized that it still won't work after the PR because "lambbda" will be changed to "a lambda". Let me fix that real quick...

BethanyG:
So the reason I want to test this is because there is one scenario where you are linking to something in the _same_ doc, and one where you are linking to something in a different doc.  IIRC, there was some sort of fix needed for Anchor Links - and I haven't found it yet.  So there might be two different forms.  Let me dig a bit.

BethanyG:
So [this is the PR in question](https://github.com/exercism/website/pull/6665), and this is now _my_ question:  _Since this says "automatically", is this part of processing, and so we **shouldn't** put the `h-` in because the code will do it twice, or do we **need** to put it in?_

BethanyG:
Looking at it closer, it does appear that we need to do that for the Anchor Links (prepend `h-`).  And if so, we certainly need to change/amend [this doc](https://exercism.org/docs/building/markdown/markdown).  But someone more fluent with Ruby and/or linking to Anchor Links needs to weigh in.

Yrahcaz7:
Should I start a thread on the forum about this then?

BethanyG:
Yeah - would you?  _Thank you_ 💙. Make sure to link to that [PR](https://github.com/exercism/website/pull/6665/changes), as well as the approach and what you saw in the JS terminal.  And maybe ask other maintainers what they've done to link to an Anchor on a page.

And I do think the Markdown Spec needs editing, as well and perhaps the [Style Guide](https://exercism.org/docs/building/tooling)? 

I don't wanna bug Erik too much, but this is probably an area where he has more expertise than the rest of us.

BethanyG:
_HOLD UP!_  Should have searched the forum better.  😅 

[This](https://forum.exercism.org/t/links-to-headings-in-docs/7328) discusses Anchor Links.  And here is an [additional PR](https://github.com/exercism/website/pull/6299) as a fix.....

Still doesn't have any explicit documentation, tho.  And I think we need some — bot "automatic" (_use the copy buttons_) and "manual" (_prepend `h-` to the section header in the URL_). 😄

Yrahcaz7:
It seems like the "automatic" way only works on the contributing docs (and maybe some other parts of the website), but not on approach docs.

exercises/practice/reverse-string/.approaches/additional-approaches/content.md
```diff
@@ -4,152 +4,142 @@ Below are some interesting strategies that are distinct from the canonical appro
...
                   (given[index] & LENGTH_MASK).bit_count())
         
-        #Calculate the index start.
-        location = index + seq_len +1
-
-        #Prepend the byte segment to the output bytearray
+        # Calculate the index start.
+        location = index + seq_len + 1
+        
+        # Prepend the byte segment to the output bytearray
```
BethanyG:
```suggestion
        # Prepend the byte segment to the output bytearray.
```

Yrahcaz7:
Somewhat of a tangent, but it seems the style guide says:

> Use the "[Oxford Comma](https://en.wikipedia.org/wiki/Serial_comma)" (also known as the Serial Comma) in lists. For example, rather than "I love my parents, Lady Gaga and Humpty Dumpty", write "I love my parents, Lady Gaga, and Humpty Dumpty".

I _have_ been using it when writing new approaches, but now I realize I missed some of those in the typo patrol...

BethanyG:
> I have been using it when writing new approaches, but now I realize I missed some of those in the typo patrol...

Let's start a new issue for Oxford Commas.  😄   And we can loop back and take care of them afterwards (_in all the docs where we missed it_).  That way, we can save the approaches audit from stalling in a recursive loop.  😂 

We could also potentially audit both the track docs and the repo docs.  But we should do that in its own issue as well.

exercises/practice/reverse-string/.approaches/built-in-reversed/content.md
```diff
@@ -1,51 +1,48 @@
...
     return output
 ```
 
-This version uses `reversed()` to reverse a `range()` object rather than feed a start/stop/step to `range()` itself.
-It then uses the reverse range to iterate over the input string and concatenate each code point to a new 'output' string.
+This version uses `reversed()` to reverse a `range()` object rather than feed a `start`/`stop`/`step` to `range()` itself.
+It then uses the reverse `range` to iterate over the input string and concatenate each code point to a new `output` string.
 This has over-complicated `reversed()`, as it can be called directly on the input string with almost no overhead.
-This has also incurs the performance hit of repeated concatenation to the 'output' string.
+This has also incurs the performance hit of repeated concatenation to the `output` string.
```
BethanyG:
```suggestion
This has also incured the performance hit of repeated concatenation to the `output` string.
```

exercises/practice/reverse-string/.approaches/list-and-join/content.md
```diff
@@ -96,53 +91,50 @@ def reverse(text):
     return output.decode("utf-8")
 ```
 
-
-This variation is operationally the same as variations #2 & #3 above, except that it encodes the string to a `utf-8` [bytearray](https://docs.python.org/3/library/stdtypes.html#bytearray).
- It then iterates over the bytearray to perform the swaps.
-Finally, the bytearray is decoded into a `utf-8` string to return the reversed word.
+This variation is operationally the same as variations #2 and #3 above, except that it encodes the string to a `utf-8` [`bytearray`](https://docs.python.org/3/library/stdtypes.html#bytearray).
```
BethanyG:
```suggestion
This variation is operationally the same as variations #2 and #3 above, except that it encodes the string to a `utf-8` [`bytearray`][bytearray-docs].
```

BethanyG: As much as I hate to do this to you, this should be a reflink.  I will try to suggest it here in review, but I might not be able to.  Here is the link that needs to be added to this at the end: `https://docs.python.org/3/library/stdtypes.html#bytearray`

exercises/practice/reverse-string/.approaches/list-and-join/content.md
```diff
@@ -96,53 +91,50 @@ def reverse(text):
...
-Measurements were taken on a  3.1 GHz Quad-Core Intel Core i7 Mac running MacOS Ventura.
+Measurements were taken on a 3.1 GHz Quad-Core Intel Core i7 Mac running MacOS Ventura.
 Tests used `timeit.Timer.autorange()`, repeated 3 times.
 Time is reported in seconds taken per string after calculating the 'best of' time.
 The [`timeit`][timeit] module docs have more details, and [note.nkmk.me][note_nkmk_me] has a nice summary of methods.
 
 [note_nkmk_me]: https://note.nkmk.me/en/python-timeit-measure/
 [timeit]: https://docs.python.org/3/library/timeit.html#python-interface
-[approach-additional-approaches]: https://exercism.org/tracks/python/exercises/reverse-string/.approaches/additional-approaches
+[approach-additional-approaches]: https://exercism.org/tracks/python/exercises/reverse-string/approaches/additional-approaches
```
BethanyG, line 140:
```suggestion
[approach-additional-approaches]: https://exercism.org/tracks/python/exercises/reverse-string/approaches/additional-approaches
[bytearray-docs]: https://docs.python.org/3/library/stdtypes.html#bytearray
```

exercises/practice/reverse-string/.articles/performance/code/Benchmark.py
```diff
@@ -38,75 +38,76 @@ def reverse_range(text):
...
 
-# #Set up columns and rows for Pandas Data Frame
+# Set up columns and rows for Pandas Data Frame
 col_headers = [f'Str Len: {len(string)}' for string in words]
 row_headers = ['reverse slice', 'iterate & prepend', 'iterate with range', 'list swap', 'list reverse',
                'reversed builtin', 'map and join']
 labels = row_headers
 
-# # empty dataframe will be filled in one cell at a time later
+# Empty dataframe will be filled in one cell at a time later
```
BethanyG:
```suggestion
# Empty dataframe will be filled in one cell at a time later.
```

exercises/practice/reverse-string/.articles/performance/code/Benchmark.py
```diff
@@ -38,75 +38,76 @@ def reverse_range(text):
...
 df = pd.DataFrame(np.nan, index=row_headers, columns=col_headers)
 
-# #Function List to Call When Timing
+# Function List to Call When Timing
```
BethanyG:
```suggestion
# Function List to Call When Timing.
```

BethanyG (changes requested):
Wow.  That was a lot of work!  Bumping the Rep on this.  😄 

I had to work pretty hard to find anything worth requesting change on.  And it was sorta a stretch.  The only thing I couldn't via suggest was to go back and do a sweep of in-line links, and make sure we use reflinks instead.  Buuut.  The more I think about it, the more I think we should do a new issue for reflinks, and not weigh this particular fix down with that.  😄

exercises/practice/reverse-string/.approaches/list-and-join/content.md
Yrahcaz7: It's fine. I think if I batch the suggestions together, it should work just fine 😄

Yrahcaz7:
@BethanyG, I'm a little confused as to what the earlier 👎 means... Does the "automatic" way work for you on the approach docs? I tried on both Safari and Firefox, and both times I didn't see any link button when hovering over the area where it should be.

BethanyG:
> @BethanyG, I'm a little confused as to what the earlier 👎 means... Does the "automatic" way work for you on the approach docs? I tried on both Safari and Firefox, and both times I didn't see any link button when hovering over the area where it should be.

Sorry!  The 👎🏽  was a "BOOO" reaction to the situation — not disagreement.  It doesn't work for me either, and should.  But we should also have docs about it.  _THANK YOU_ btw for opening a forum issue.  😄

BethanyG (approved):

BethanyG:
Are we ready to merge?

Yrahcaz7:
🤔 I'm conflicted between waiting to get responses on the forum post vs just pushing it and seeing if it works. Thoughts?

BethanyG:
> 🤔 I'm conflicted between waiting to get responses on the forum post vs just pushing it and seeing if it works. Thoughts?

On one hand, it was broken before — don't see how it could be more broken.  On the other?  A discussion might be good, although I _suspect_ most folx involved would not have encountered the situation, and would have absolutely no idea how to solve it (_apart from Aaron and Erik who made the changes in the first place_).

Yrahcaz7:
I guess it would be better to just merge and test if it works, then.

Yrahcaz7:
It seems like it works! (At first I thought that it didn't, but of course it was just a caching issue)

PR #4269, opened 2026-07-15
Fix test order in Raindrops exercise

deepshekhardas:
## Summary

Added numeric prefixes to test method names in the Raindrops exercise to ensure tests run in the defined order rather than alphabetical order.

### Changes:
1. Updated 
aindrops_test.py - renamed test methods with


BethanyG:
@deepshekhardas,

This repo does not accept community contributions.  This is also not a change we want made, as this is one exercise out of 144.  This is a change that is likely needed systemically and elsewhere.  Thanks.

PR #4273, opened 2026-07-28
[Ellen's Alien Game] Test that an instance variable is used for the health, not a class variable.

IsaacG:
In response to [this forum post](https://forum.exercism.org/t/removing-unnecesary-line-in-test-in-ellens-alien-game/68114/3).


IsaacG:
[This solution](https://gist.github.com/IsaacG/a5c9c5a9bcc11b76c1c01b6177230d7f) uses a class var for the health. It passes the existing tests and fails this new test.

exercises/concept/ellens-alien-game/classes_test.py
```diff
@@ -7,7 +7,7 @@
 except ImportError as import_fail:
     # pylint: disable=raise-missing-from
     raise ImportError("\n\nMISSING CLASS --> We tried to import the 'Alien' class from "
-                      "your classes.py file, but could not find it." 
+                      "your classes.py file, but could not find it."
```
BethanyG: The space here needs to remain (but _inside_ the quotes, or a \n could be used as well.)  Otherwise, with Python's string concatenation rules, it will be smoosed together with the following sentence with no space at all between. And indeed: its broken in production right now. Another alternative is preceding the follow-on sentence.

BethanyG:
```suggestion
                      "your classes.py file, but could not find it. "
```

BethanyG (changes requested):
Just one nit (and the error was in the test file already).  Either a space or a \n  will do.

Per our Discord discussion, I'd love it if we added some comments and also cleaned up that one test.  I'll put a comment below, once I get out of review mode.

PS - comments don't have to be long or exhaustive, but maybe some ... breadcrumbs so we don't forget what happened......

BethanyG:
The offending test, and my proposed edit (delete everything that's commented out)

```python
    @pytest.mark.task(taskno=6)
    def test_alien_class_variable(self):
        """Test class attribute/variables are identical across instances."""

        alien_one, alien_two = Alien(0, 2), Alien(-6, -1)
        # Alien.health = 6

        created_error_message = ('Created two new Aliens and requested the '
                                 'total_aliens_created attribute for each one. '
                                 f'Received {alien_one.total_aliens_created, alien_two.total_aliens_created} '
                                 f'for total_aliens_created, but the tests expect '
                                 f'the class attributes for each newly created Alien to be identical. ')

        # health_error_message = ('Created two new Aliens and requested the '
        #                         f'health attribute for each one. Received {alien_one.health, alien_two.health} '
        #                         'for health, but the tests expect the class '
        #                         'attributes for each newly created Alien to be identical. ')

        self.assertEqual(alien_two.total_aliens_created,
                         alien_one.total_aliens_created,
                         msg=created_error_message)
        #
        # self.assertEqual(alien_two.__class__.health,
        #                  alien_one.__class__.health,
        #                  msg=health_error_message)
```

exercises/concept/ellens-alien-game/classes_test.py
IsaacG: Done!

BethanyG (approved):
Looks good!  Thank you so much for doing this! 🚀

